PR #800, opened 2024-05-01
Migrate to threshold signing

JesseAbram:
This branch is meant to change the Subgroup system and will be merged to master when ready

- [x] Remove Subgroups
- [x]  Remove Private Mode
- [x] Remove SigPartySize
- [x] Add Threshold DKG and Signing


ameba23:
@JesseAbram maybe a good next step for this branch would be to have our test setup use 3 TSS nodes instead of 2, ahead of adding support for t-of-n signing.  Because with 2 nodes, t is always going to need to 2.  What do you think?

I'm not sure how easy or difficult this will be.  If it turns out to be complicated and lots of stuff needs re-writing, maybe we should take the opportunity to switch to using docker-compose for some of our tests.  Although of course that means we also need to change to the docker-compose setup to have 3 nodes.

pallets/registry/src/lib.rs
```diff
@@ -461,28 +446,24 @@ pub mod pallet {
...
             // if no one has sent in a verifying key yet, use current
             if registering_info.verifying_key.is_none() {
                 registering_info.verifying_key = Some(verifying_key.clone());
             }
 
             let registering_info_verifying_key =
                 registering_info.verifying_key.clone().ok_or(Error::<T>::NoVerifyingKey)?;
-
-            if registering_info.confirmations.len() == T::SigningPartySize::get() - 1 {
+            // TODO
```
JesseAbram, line 465: remove

HCastano (approved):
Good stuff guys 💪 🙌 

I'm not entirely sure if everything here is quite working as expected, but this branch is at a point where we should get it into `master` and be a bit more diligent about getting fixes in through PRs.

PR #832, opened 2024-05-14
Downgrade `check_in_registration_group` error log to warning

JesseAbram:
The error is not really an error and should just log a warning


HCastano (approved):
I pushed up the event change - I want to get this into today's release 😄

PR #836, opened 2024-05-15
Release `v0.1.0-rc.1`

HCastano:
Since this is the first release meant for public consumption, I figured it would be a
good idea to move from a `PATCH` level version to a `MINOR` level version 🎉

This is a small release mostly focused on internal changes. There are also no runtime
changes which is nice since it means we don't need to update benchmarks.

On the runtime side, I've change the `spec_version` to follow our Cargo version number
convention (`MAJOR_MINOR_PATCH`). Developers shouldn't necessarily rely on
the `spec_version` to carry any real information about runtime compatiblity, but I
thought it would be easier for us to match runtime versions with releases. I'm open to
suggestions on this.


JesseAbram:
we should probably hold off a day or two for the team to get us their endowed addresses?

HCastano:
> we should probably hold off a day or two for the team to get us their endowed addresses?

We can include this in a follow up `rc` or the full release. But at least with this tag we can hand something off to the SDK team

JesseAbram (approved):

runtime/src/lib.rs
```diff
@@ -208,7 +208,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // We update this if the runtime behaviour has changed. When this happens we set the
     // `impl_version` to `0`.
     #[allow(clippy::zero_prefixed_literal)]
-    spec_version: 012,
+    spec_version: 00_01_00,
```
ameba23, line 211: Maybe we need an explanation in the release checklist as to what constitutes a major/minor/patch spec version change

ameba23 (approved):
:1st_place_medal:

runtime/src/lib.rs
HCastano: Yep I plan to hijack your `RELEASE_CHECKLIST` PR and add some info about that there 😉

HCastano:
I wanted to do an Ansible testnet deployment of this before merging, but it doesn't looks like I've got all the right credentials to do that - and I'm not sure when the DevOps team is gonna be around to help me out.

Instead I've done a sanity check with the local Docker compose setup, which uses the [test-hc-release-d638aec1-v0.1.0-rc.1](https://hub.docker.com/layers/entropyxyz/entropy/test-hc-release-d638aec1-v0.1.0-rc.1/images/sha256-33d84bcc71b3e6330d49d793599d37c7147f7ef66b01e8e31f60e27b6fbef0fa?context=explore) images.

I've basically spun up the network and ran:

```
 cargo run -p entropy-test-cli -- register One public ./crates/testing-utils/template_barebones.wasm
 cargo run -p entropy-test-cli -- sign $VERIFYING_KEY "Hello, Docker Compose"
 ```

With that I'm gonna merge this and move forward with the release

PR #838, opened 2024-05-15
Change to mnemonic entry for cli

JesseAbram:
While using the test cli it expects a name like alice one etc, however this makes it hard to input mnemonics which we will need for testnet as those named accounts no longer and should not have funds

### DIFF
--- a/crates/test-cli/src/main.rs
+++ b/crates/test-cli/src/main.rs
@@ -73,7 +73,7 @@ enum CliCommand {
         /// This is used to send the register extrinsic and so it must be funded
         ///
         /// Optionally may be preceeded with "//" eg: "//Bob"
-        program_account_name: String,
+        mnemonic: String,
         /// The access mode of the Entropy account
         #[arg(value_enum, default_value_t = Default::default())]
         key_visibility: Visibility,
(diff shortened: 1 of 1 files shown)


ameba23 (changes requested):
I would want to not replace one with the other, but make it still possible to use a name, as i use this all the time for testing with docker-compose, where we do have the pre-funded accounts, and i don't want to be copying and pasting mnemonics all the time.  You can also make a name that is hard to guess with `subkey inspect //<a very long name>`.

My proposal is to attempt to pass the given string as a mnemonic and if that fails interpret it as a name just as before.  So both are possible.

If we are really worried about people stealing testnet tokens then maybe its time to add private key storage to this. It would be easy to do, but seems silly to do when we have others working on a full featured CLI is JS.

ameba23 (approved):
Ok i think i over-reacted - [Pair::from_string](https://docs.rs/sp-core/22.0.0/sp_core/crypto/trait.Pair.html#method.from_string_with_seed) will actually handle both mnemonic and names for us, so i expect this will work fine as it is.  So the only difference is that adding the `//` prefix is needed. I am fine with this.

JesseAbram:
> Ok i think i over-reacted - [Pair::from_string](https://docs.rs/sp-core/22.0.0/sp_core/crypto/trait.Pair.html#method.from_string_with_seed) will actually handle both mnemonic and names for us, so i expect this will work fine as it is. So the only difference is that adding the `//` prefix is needed. I am fine with this.

oh ya maybe I should have mentioned you can still pass through //Alice, I checked to make sure

HCastano:
For the record, we could already pass mnemonics into the test CLI account field prior to this change

PR #843, opened 2024-05-17
Add a way to change program modification account

JesseAbram:
Realized while tweeting on twitter that we didn't have a method to do this, was easy to do, allows people to sell accounts which idk ppl on CT like

mainly that is a joke....although all true, but this wasn't here and we should have a way to let users do it


ameba23 (approved):
I'm tentatively approving.  I feel like there could maybe be something dangerous around giving someone an account without their consent but i can't come up with something thats not already possible with `register`.

At worst, you can make it impossible for someone to use their program modification account with another entropy account by filling up `ModifiableKeys` for their program modification account to `MAX_MODIFIABLE_KEYS`.  But since this is already possible with `register` this PR doesn't make that any worse.  It also not really that much of a problem.

HCastano:
@JesseAbram pls don't merge this until the `v0.1.0` release is published 🙏

(I also want to take a look, but the above is more important lol)

JesseAbram:
> I'm tentatively approving. I feel like there could maybe be something dangerous around giving someone an account without their consent but i can't come up with something thats not already possible with `register`.
> 
> At worst, you can make it impossible for someone to use their program modification account with another entropy account by filling up `ModifiableKeys` for their program modification account to `MAX_MODIFIABLE_KEYS`. But since this is already possible with `register` this PR doesn't make that any worse. It also not really that much of a problem.

ya we can write an issue to track, not impossible to fix you can batch a tx to remove a modifiable key and then set the one you want in the same tx......we can also fix by having it require a sig from a program_mod_account, but that would remove the ability to keep that account cold which I don't like opened #848

pallets/registry/src/lib.rs
```diff
@@ -374,12 +376,66 @@ pub mod pallet {
             .into())
         }
 
+        /// Allows a user's program modification account to change itself
+        #[pallet::call_index(3)]
+        //TODO fix
```
HCastano, line 381: What needs to be addressed here?

pallets/registry/src/lib.rs
```diff
@@ -374,12 +376,66 @@ pub mod pallet {
             .into())
         }
 
+        /// Allows a user's program modification account to change itself
```
HCastano, line 379:
```suggestion
        /// Allows a user's program modification account to change itself.
```

pallets/registry/src/lib.rs
```diff
@@ -374,12 +376,66 @@ pub mod pallet {
...
+        #[pallet::call_index(3)]
+        //TODO fix
+        #[pallet::weight({
+                 <T as Config>::WeightInfo::change_program_mod_key(MAX_MODIFIABLE_KEYS)
+             })]
+        pub fn change_program_mod_key(
```
HCastano, line 385:
```suggestion
        pub fn change_program_modification_account(
```

pallets/registry/src/tests.rs
```diff
@@ -371,6 +371,86 @@ fn it_changes_a_program_pointer() {
...
+        assert_eq!(Registry::registered(expected_verifying_key.clone()).unwrap(), registered_info);
+
+        assert_noop!(
+            Registry::change_program_mod_key(
+                RuntimeOrigin::signed(2),
+                expected_verifying_key.clone(),
+                3
+            ),
+            Error::<Test>::NotAuthorized
+        );
```
HCastano, line 416: I found this pretty confusing since I thought `ModifyableKeys` would be populated earlier. Maybe it's worth having a comment here that you're specifically testing the case of a program mod account not having control over a certain verifying key, and thus the manual setup?

HCastano (approved):
Some small comments, but this looks fine.

Since the `call_index` changed, can you add the `Bump spec_version` and `Bump transaction_version` labels to this PR?

pallets/registry/src/lib.rs
JesseAbram: nothing will remove it is for me to remember to fix the weights which I did

pallets/registry/src/tests.rs
JesseAbram: ya will add a comment, it should never happen, but just cuz I can't think of a reason that it can't happen doesn't mean it won't so a noop test to me seems prudent

PR #850, opened 2024-05-22
Improve `--no-sync` docs and tiny sync check

HCastano:
There's been some confusion around the usage of `--no-sync`. This PR adds some
documentation around when it should and shouldn't be used.

Admittedly, the crux of the issue stems from having TSS IP addresses on-chain without any
way to determine if they're properly synced or not. If we knew this status we could have
the TSS itself determine whether or not to accept a peers key-shares.

I've also done one little tidy of the logic check around syncing 🧹

cc @vitropy @mixmix

### DIFF
--- a/crates/threshold-signature-server/src/validator/api.rs
+++ b/crates/threshold-signature-server/src/validator/api.rs
@@ -67,62 +67,57 @@ pub const BATHC_SIZE_FOR_KEY_VALUE_GET: usize = 10;
 /// - finding a server in their subgroup that is synced
 /// - getting all shards from said validator
 #[tracing::instrument(skip(kv_store))]
-pub async fn sync_validator(no_sync: bool, dev: bool, endpoint: &str, kv_store: &KvManager) {
-    if !no_sync {
-        let api = get_api(endpoint).await.expect("Issue acquiring chain API");
-        let rpc = get_rpc(endpoint).await.expect("Issue acquiring chain RPC");
-        let mut is_syncing = true;
-        let sleep_time = Duration::from_secs(20);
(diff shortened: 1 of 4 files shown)


crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -67,62 +67,57 @@ pub const BATHC_SIZE_FOR_KEY_VALUE_GET: usize = 10;
...
-        }
-        // if not in subgroup retry until you are
-        let mut my_subgroup = get_subgroup(&api, &rpc, signer.account_id()).await;
-        while my_subgroup.is_err() {
-            tracing::warn!("The signing account is not in the validator set, retrying sync");
+pub async fn sync_validator(dev: bool, endpoint: &str, kv_store: &KvManager) {
+    let api = get_api(endpoint).await.expect("Issue acquiring chain API");
+    let rpc = get_rpc(endpoint).await.expect("Issue acquiring chain RPC");
+    let mut is_syncing = true;
+    let sleep_time = Duration::from_secs(20);
```
ameba23, line 74: Not related to this PR but ideally this should be a constant

JesseAbram (approved):
looks good I wouldn't touch this too much going forward as it will be removed when #800 gets merged (as no longer syncing keyshares)

HCastano:
@ameba23 I agree with your suggestions, but ya if we're just gonna get rid of this code with Jesse's PR I'm gonna leave the code as is. If it turns out that the panics in the sync code cause problems in the meantime we address them

mixmix:
thanks @HCastano , my main desire was to surface the unclarity which was effecting devops
@vitropy did this add clarity for you?

PR #851, opened 2024-05-23
Bump synedrion to unreleased git version ahead of adding t-of-n in Migrate-To-Threshold branch

ameba23:
This updates synedrion with the latest changes - moving towards adding threshold signing but not actually functionally changing things for now.

Stored keyshares in the kvdb are now no longer `KeyShare<KeyParams>` but `(ThresholdKeyShare<KeyParams, PartyId>, AuxInfo<KeyParams, PartyId>)`. `entropy-protcol` has a type alias for this, `entropy_protocol::KeyShareWithAuxData`.  But we could also make a struct with some nice helpers.

The proactive refresh protocol is still called proactive refresh in our code, but under the hood it is now a key reshare with the same participants as before. This is to distinguish it from a key reshare which changes participants, which we may want to add at some point.

### DIFF
--- a/crates/protocol/src/protocol_transport/mod.rs
+++ b/crates/protocol/src/protocol_transport/mod.rs
@@ -27,12 +27,11 @@ use futures::{SinkExt, StreamExt};
 use noise::EncryptedWsConnection;
 pub use subscribe_message::SubscribeMessage;
 #[cfg(feature = "server")]
-use tokio::net::TcpStream;
 use tokio::sync::{broadcast, mpsc};
 #[cfg(feature = "server")]
-use tokio_tungstenite::{connect_async, tungstenite, MaybeTlsStream, WebSocketStream};
+use tokio_tungstenite::{tungstenite, MaybeTlsStream, WebSocketStream};
 
-use crate::{errors::UserRunningProtocolErr, PartyId, ProtocolMessage};
+use crate::{PartyId, ProtocolMessage};
 
 /// Channels between a remote party and the signing or DKG protocol
 pub struct WsChannels {
@@ -176,27 +175,6 @@ pub async fn ws_to_channels<T: WsConnection>(
             }
         }
     }
-}
-
-/// Open a ws connection in a native environment
-#[cfg(feature = "server")]
-pub async fn open_ws_connection(
-    address: String,
-) -> Result<WebSocketStream<MaybeTlsStream<TcpStream>>, UserRunningProtocolErr> {
-    let (ws_stream, _response) = connect_async(address)
-        .await
-        .map_err(|e| UserRunningProtocolErr::Connection(e.to_string()))?;
-    Ok(ws_stream)
-}
-
-/// Open a ws connection on wasm with the JS websocket API
-#[cfg(feature = "wasm")]
-pub async fn open_ws_connection(
-    address: String,
-) -> Result<gloo_net::websocket::futures::WebSocket, UserRunningProtocolErr> {
-    let ws_stream = gloo_net::websocket::futures::WebSocket::open(&address)
-        .map_err(|e| UserRunningProtocolErr::Connection(e.to_string()))?;
-    Ok(ws_stream)
 }
 
 // This dummy trait is only needed because we cant add #[cfg] to where clauses
--- a/crates/protocol/src/protocol_transport/mod.rs
+++ b/crates/protocol/src/protocol_transport/mod.rs
@@ -26,7 +26,6 @@ use errors::WsError;
 use futures::{SinkExt, StreamExt};
 use noise::EncryptedWsConnection;
 pub use subscribe_message::SubscribeMessage;
-#[cfg(feature = "server")]
 use tokio::sync::{broadcast, mpsc};
 #[cfg(feature = "server")]
 use tokio_tungstenite::{tungstenite, MaybeTlsStream, WebSocketStream};
--- a/crates/protocol/tests/helpers/mod.rs
+++ b/crates/protocol/tests/helpers/mod.rs
@@ -30,11 +30,12 @@ use entropy_shared::X25519PublicKey;
 use futures::future;
 use sp_core::{sr25519, Pair};
 use std::{
+    fmt,
     sync::{Arc, Mutex},
     time::Duration,
 };
 use subxt::utils::AccountId32;
-use synedrion::KeyShare;
+use synedrion::{AuxInfo, KeyShare};
 use tokio::{
     net::{TcpListener, TcpStream},
     time::timeout,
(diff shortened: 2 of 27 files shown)


ameba23:
@fjarri I wanted to make this update to have our current setup working with the latest changes to synedrion, before adding support for t-of-n signing in a separate PR.

But i am not sure if it is possible to do a key refresh with non-threshold keyshares anymore, as it looks like there is no longer a way for to update a `KeyShare` with a `KeyShareChange` from the refresh protocol.  

I see that `ThresholdKeyShare` allows you to combine a `ThresholdKeyShareSeed` with a `KeyShareChange` (with `ThresholdKeyShare::new`).  Does this mean we will need to store the keyshare seed as well as the keyshare, in order to create updated keyshares following a refresh?  Or is there some way of getting the keyshare seed from an existing keyshare?

fjarri:
> But i am not sure if it is possible to do a key refresh with non-threshold keyshares anymore, as it looks like there is no longer a way for to update a KeyShare with a KeyShareChange from the refresh protocol.

Yes, the CGGMP's KeyRefresh won't work with threshold shares. I think we will have to use KeyResharing as KeyRefresh, by resharing to the same set of nodes.

> I see that ThresholdKeyShare allows you to combine a ThresholdKeyShareSeed with a KeyShareChange (with ThresholdKeyShare::new).

This has been removed in https://github.com/entropyxyz/synedrion/pull/119 because it turned out it doesn't work as expected :) Also, that PR made the following changes:
- Key shares and auxiliary data are now separate. So no more KeyShareSeed/KeyShare - it's just (Threshold)KeyShare now. 
- There is an AuxGen protocol that just generates the auxiliary data.
- `make_interactive_signing_session()` takes a `KeyShare` and `AuxInfo` objects. 
- The threshold sequence test was updated accordingly.

AuxGen may not be necessary depending on the resolution on https://github.com/entropyxyz/synedrion/issues/98, but for now the philosophy is the following: nodes participating in AuxGen collectively agree that all the public auxiliary data is valid, and safe, and the corresponding nodes own the secret parts of it, and so on. After you run KeyResharing you need to update auxiliary data for the new set of nodes (or just update it on the regular basis for the same set of nodes, as appropriate).

ameba23:
All tests are now passing at the `entropy-protocol` level.  But in `entropy-tss` the proactive refresh test fails.

crates/protocol/src/protocol_transport/mod.rs
```diff
@@ -178,27 +176,6 @@ pub async fn ws_to_channels<T: WsConnection>(
     }
 }
 
-/// Open a ws connection in a native environment
-#[cfg(feature = "server")]
-pub async fn open_ws_connection(
```
ameba23, line 183: Unrelated to this PR but these are no longer needed as they relate to private mode which has been removed

crates/protocol/tests/helpers/mod.rs
```diff
@@ -50,11 +52,16 @@ struct ServerState {
...
     Sign(RecoverableSignature),
-    ProactiveRefresh(KeyShare<KeyParams>),
-    Dkg(KeyShare<KeyParams>),
+    ProactiveRefresh(ThresholdKeyShare<KeyParams, PartyId>),
+    Dkg(KeyShareWithAuxInfo),
+}
+
+impl fmt::Debug for ProtocolOutput {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Success")
```
ameba23, line 63: This is hacky but wont be needed anymore once `KeyShare` / `ThresholdKeyShare` implement `Debug` which is coming soon. Its also only used in these tests - it never gets logged out but its for some reason needed to be able to unwrap a `Result` for this type.

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -119,11 +120,11 @@ pub async fn create_clients(
 
 pub async fn spawn_testing_validators(
     passed_verifying_key: Option<Vec<u8>>,
-    // If this is true a keyshare for the user will be generated and returned
-    extra_private_keys: bool,
+    // If this is given a keyshare for the user will be generated and returned
+    extra_private_key: Option<sr25519::Pair>,
```
ameba23, line 124: Needed because keyshares are now intrinsicly tied to an identity, so if you want and extra keyshare you need to say who it is for.

That said im not sure if we actually use this anywhere now that private mode is gone and the refresh test no longer uses it. But thats for another PR.

crates/threshold-signature-server/src/signing_client/protocol_execution/context.rs
```diff
@@ -13,26 +13,27 @@
...
+    pub key_share: ThresholdKeyShare<KeyParams, PartyId>,
+    pub aux_info: AuxInfo<KeyParams, PartyId>,
 }
 
 impl SignContext {
-    pub fn new(sign_init: SignInit, key_share: KeyShare<KeyParams>) -> Self {
-        Self { sign_init, key_share }
-    }
-
-    pub fn msg_to_sign(&self) -> &PrehashedMessage {
```
ameba23, line 35: I removed this because `synedrion::PrehashedMessage` doesnt exist anymore, and we never actually call this method.

ameba23:
To get the `entropy-tss` `test_proactive_refresh` test to work with both eve and dave, i had to add a 'deterministic key share' constant for dave.  For the record, the code which generated the associated verifying key for that is here:

```rust
use entropy_protocol::PartyId;
use hex_literal::hex;
use lazy_static::lazy_static;
use rand_core::OsRng;
use subxt::utils::AccountId32;
use synedrion::{k256::ecdsa::SigningKey, KeyShare};

lazy_static! {
    pub static ref DETERMINISTIC_KEY_SHARE_DAVE: [u8; 32] =
        hex!("06b07fd12cdfb94fbde3ff2098e9f19bb11b00959680cfbd15c914b025f298d7");
}

fn main() {
    let signing_key =
        Some(SigningKey::from_bytes((&*DETERMINISTIC_KEY_SHARE_DAVE).into()).unwrap());

    let ids = vec![
        PartyId::new(AccountId32([1u8; 32])),
        PartyId::new(AccountId32([2u8; 32])),
    ];
    let shares = KeyShare::<synedrion::TestParams, PartyId>::new_centralized(
        &mut OsRng,
        &ids,
        signing_key.as_ref(),
    );
    let verifying_key = shares[0]
        .verifying_key()
        .to_encoded_point(true)
        .as_bytes()
        .to_vec();
    println!("{:?}", verifying_key);
}
```

JesseAbram (approved):

PR #853, opened 2024-05-24
Move TSS mnemonic out of keystore

HCastano:
Right now the mnemonic used to derive the TSS SR25519 account ID and X25519 public keys
is generated and stored within the encrypted keystore.

This means that it is impossible for operators to predict, prefund, or back up these
accounts. This can make certain operations, such as moving the TSS to a different server,
challenging.

This PR adds a `--mnemonic` flag which allows operators to pass a BIP-39 mnemonic when
starting up the process.

This flag is not required if:
- An existing mnemonic has been found in the KVDB
- A development account is being used (e.g `--alice`)
- The TSS is being run is compiled in a `test` configuration

If the flag _is_ passed in however, it will **overwrite any existing mnemonic in the
database**. While potentially dangerous if operators aren't aware of this behaviour, I
figured that having it this way would at least keep things consistent and predictable
from an operators point of view.

It would be nice to have different options for passing the mnemonic, such as a file or
environment variable - so I'll probably address this in a follow up.

For now, operators can use something like the 1Password CLI to avoid leaving mnemonics in
plaintext.

`entropy-tss --mnemonic '$(op read "op://Operator/tss-mnenmonic")'`

cc @vitropy @cooldracula


JesseAbram (approved):
ya Id say in a follow up making a mnemonic and an overwrite_mnemonic option seems nice as if someone really knows what they are doing can call it

HCastano:
> ya Id say in a follow up making a mnemonic and an overwrite_mnemonic option seems nice as if someone really knows what they are doing can call it

I thought about having a two flag system, but imo it makes the user experience more
convoluted. A single flag with a well-known behaviour ("it writes a mnemonic") makes more
sense to me.

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -160,84 +164,119 @@ pub struct StartupArgs {
...
+    /// The BIP-39 mnemonic (i.e seed phrase) to use for deriving the Threshold Signature Server
+    /// SR25519 account ID and the X25519 public key.
+    ///
+    /// The SR25519 account is responsible for signing and submitting extrinsics to the Entropy
+    /// network.
+    ///
+    /// The X25519 public key is used for encrypting/decrypting messages to other threshold
+    /// servers.
+    ///
+    /// **Warning**: Passing this flag will overwrite any existing mnemonic! If you would like to
```
ameba23, line 183: Maybe we can also put a warning here that the mnemonic will be in your shell command history if you use this option. Most shells will omit the command from history if it is preceded with a space character.

ameba23 (approved):
Great work. This fixes a pretty serious usability problem that its currently hard to back up the mnemonic, and if you forget the kvdb password it is gone forever.

I would definitely like to see the option to pass it an an environment variable due to it otherwise ending up in shell command history.

As for the clobbering existing mnemonic thing i don't really know. I can see its useful to have deterministic behavior, but i guess its not totally idiot proof.

HCastano:
> As for the clobbering existing mnemonic thing i don't really know. I can see its useful to have deterministic behavior, but i guess its not totally idiot proof.

Yeah, it's a little bit of a risk, but I suspect from an automation point of view it'll make things more clear. We can change this if Vi or Zack complain haha

crates/threshold-signature-server/src/helpers/launch.rs
HCastano: Sure, I can include something - but I'll do it in one of the follow-ups. I don't really want to re-run the CI for this 😄

Issue #855, opened 2024-05-27
Add prerequisites for the Entropy Test CLI.


johnnymatthews:
In `/crates/test-cli`, the user is prompted to install the Entropy Test CLI with `cargo install entropy-test-cli`. If the user runs that command without OpenSSL installed, they'll hit an error halfway through the compilation.

# Tasks

- [x] Add a prerequisites section to the `/crates/test-cli/README.md` file listing all the dependencies that this CLI requires.

PR #860, opened 2024-05-29
Test CLI store program command no longer defaults to the barebones program

ameba23:
The store program command in the test-cli now takes a program file as a mandatory positional argument. Before it was optional and fell back to the barebones program.

Removing this means we can remove the dependency `entropy-testing-utils` which in turn has `entropy-tss` as a dependency. This means we have quite a few less dependencies.

We do still need openssl so this doesn't fix https://github.com/entropyxyz/entropy-core/issues/855 - but it might reduce the need for some other packages.

See related discussion: https://github.com/entropyxyz/entropy-core/pull/856#issuecomment-2135631441

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ At the moment this project **does not** adhere to
 ### Added
 - Add a way to change program modification account  ([#843](https://github.com/entropyxyz/entropy-core/pull/843))
 
+### Changed
+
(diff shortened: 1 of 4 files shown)


JesseAbram (approved):
oh, I already removed this in #856 but did not remove it from the cargo.toml, I can do that now and we can avoid this PR?

PR #864, opened 2024-05-30
Add support for `--mnemonic-file` and `THRESHOLD_SERVER_MNEMONIC`

### LINKED ISSUE #862: Allow TSS `mnemonic` to be passed in via environment variables and files
Right now there's only one way to pass a mnemonic into the TSS KVDB, the `--mnemonic`
flag introduced in #853.

We should give operators different ways to pass in this mnemonic though. For example,
environment variables or files:

```
TSS_MNEMONIC="alarm ... where" entropy-tss
entropy-tss --mnenmonic-file "/path/to/file.txt"
```

HCastano:
This is a follow up to #853, in which the `--mnemonic` flag was introduced.

Now we add support for passing a mnemonic via a file with the `--mnemonic-file` flag, or
through an environment variable, `THRESHOLD_SERVER_MNEMONIC`.

The behaviour around overwriting existing mnemonics if any of these are supplied remains
the same.

Closes #862.

### DIFF
--- a/crates/threshold-signature-server/src/main.rs
+++ b/crates/threshold-signature-server/src/main.rs
@@ -57,7 +57,29 @@ async fn main() {
 
     let app_state = AppState::new(configuration.clone(), kv_store.clone());
 
-    if let Some(mnemonic) = args.mnemonic {
+    // We consider the inputs in order of most to least explicit: CLI flag, supplied file,
+    // environment variable.
+    let user_mnemonic = args
+        .mnemonic
+        .or_else(|| {
+            args.mnemonic_file.map(|path| {
+                let file = std::fs::read(path).expect("Unable to read mnemonic file.");
+                let mnemonic = std::str::from_utf8(&file)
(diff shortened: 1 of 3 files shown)


crates/threshold-signature-server/src/main.rs
```diff
@@ -57,7 +57,29 @@ async fn main() {
...
+                let mnemonic = std::str::from_utf8(&file)
+                    .expect("Unable to convert provided mnemonic to UTF-8 string.")
+                    .trim();
+
+                bip39::Mnemonic::parse_normalized(mnemonic)
+                    .expect("Unable to parse given mnemonic.")
+            })
+        })
+        .or_else(|| {
+            std::env::var("THRESHOLD_SERVER_MNEMONIC").ok().map(|mnemonic| {
```
JesseAbram, line 76: adding a the dotenv crate would be an nice extra option

JesseAbram (approved):

ameba23 (approved):
:100:

PR #866, opened 2024-05-30
Replace timestamp with block number

### LINKED ISSUE #857: Timestamps coming from js are too unreliable, switch to block number instead
Why is this issue relevant?
Timestamps are added to the signature requests from the sdk, but node timestamps are notoriously unreliable when converting from milliseconds -> nanoseconds, seconds causing Invalid Type errors through serde, as well as `Message too old` errors
What steps are required to resolve this?

Does this change the spec? HTTP, extrinsic, or storage? Is it breaking? Clearly describe the new interface.

JesseAbram:
Closes #857

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -171,12 +171,12 @@ pub async fn sign(
     let message_hash_hex = hex::encode(message_hash);
     let validators_info = get_current_subgroup_signers(api, rpc, &message_hash_hex).await?;
     tracing::debug!("Validators info {:?}", validators_info);
-
+    let block_number = rpc.chain_get_header(None).await?.ok_or(ClientError::BlockNumber)?.number;
     let signature_request = UserSignatureRequest {
         message: hex::encode(message),
         auxilary_data: Some(vec![auxilary_data.map(hex::encode)]),
         validators_info: validators_info.clone(),
-        timestamp: get_current_time(),
+        block_number,
         hash: HashingAlgorithm::Keccak,
         signature_verifying_key: signature_verifying_key.to_vec(),
     };
@@ -434,14 +434,4 @@ async fn select_validator_from_subgroup(
         }
     };
     Ok(address.clone())
-}
-
-#[cfg(feature = "full-client-wasm")]
-fn get_current_time() -> SystemTime {
-    use std::time::{Duration, UNIX_EPOCH};
-    UNIX_EPOCH + Duration::from_secs(js_sys::Date::now() as u64)
-}
-#[cfg(not(feature = "full-client-wasm"))]
-fn get_current_time() -> SystemTime {
-    SystemTime::now()
 }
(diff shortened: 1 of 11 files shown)


frankiebee:
💛

crates/client/src/client.rs
```diff
@@ -435,13 +434,3 @@ async fn select_validator_from_subgroup(
     };
     Ok(address.clone())
 }
-
-#[cfg(feature = "full-client-wasm")]
-fn get_current_time() -> SystemTime {
-    use std::time::{Duration, UNIX_EPOCH};
-    UNIX_EPOCH + Duration::from_secs(js_sys::Date::now() as u64)
-}
-#[cfg(not(feature = "full-client-wasm"))]
```
ameba23, line 444: Ah good that we no longer need to worry about the system clock on different platforms here

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -48,7 +48,7 @@ use crate::{
 #[derive(Debug, Deserialize, Serialize, Clone)]
 pub struct Keys {
     pub keys: Vec<String>,
-    pub timestamp: SystemTime,
+    pub block_number: BlockNumber,
```
ameba23, line 51: Ah so we also change this in the `sync_kvdb` payload

ameba23 (approved):
:1st_place_medal:

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -206,7 +208,7 @@ async fn test_sync_kvdb() {
     assert_eq!(result_4.status(), 500);
     assert_eq!(result_4.text().await.unwrap(), "Forbidden Key");
 
-    keys.timestamp = keys.timestamp.checked_sub(TIME_BUFFER).unwrap();
+    keys.block_number = keys.block_number + BLOCK_BUFFER + 10;
```
HCastano, line 211: Why do we add an extra `10` here if we already have the `BLOCK_BUFFER`?

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -235,6 +242,7 @@ pub async fn get_and_store_values(
     recip_server_info: ServerInfo<subxt::utils::AccountId32>,
     signer: &PairSigner<EntropyConfig, sr25519::Pair>,
     x25519_secret: &StaticSecret,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
```
HCastano, line 245: Would be nice to not need to pass this in, but 🤷‍♂️

HCastano (approved):
👍

crates/threshold-signature-server/src/validator/api.rs
JesseAbram: ya cuz it shares the check_stale function but I wouldn't worry to much, gonna need a whole re write when we go to threshold

crates/threshold-signature-server/src/validator/api.rs
JesseAbram: ya but wouldn't worry about it, all this is gonna need a re write when we go to threshold

PR #868, opened 2024-05-31
t of n key generation and signing at the `entropy-protocol` level 

ameba23:
This updates the DKG protocol to internally use the 3 protocols needed for setting up threshold keyshares (key init, reshare, and aux gen), and adds an integration test for `entropy-protocol` demonstrating t of n signing.

It is not yet possible to test this with `entropy-tss` because of not having enough nodes in our test setup, so currently `entropy-tss` always uses a threshold of n.

During the DKG protocols, we need to send the verifying key to the parties who were not present in the Key Init session. To be able to do this, the payload of a `ProtocolMessage` is now an enum which may either be a synedrion protocol message, or a verifying key.  This is a bit ugly - it might be nicer if this passing of the verifying key could be represented as a synedrion `CombinedMessage`.

### DIFF
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -182,7 +182,7 @@ pub async fn execute_signing_protocol(
 /// Execute dkg.
 #[tracing::instrument(
     skip_all,
-    fields(threshold_accounts, my_idx),
+    fields(threshold_accounts, my_idx, threshold),
     level = tracing::Level::DEBUG
 )]
 pub async fn execute_dkg(
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -202,9 +202,6 @@ pub async fn execute_dkg(
     // Setup channels for the next session
     let chans = Channels(broadcaster.clone(), rx);
 
-    let old_threshold = init_keyshare.to_threshold_key_share().threshold();
-    println!("Old threshold {}", old_threshold);
-
     // If were a member of t, send verifying_key to others
     // Otherwise receive verifying_key from another
     let verifying_key = init_keyshare.verifying_key();
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -19,7 +19,8 @@ use rand_core::{CryptoRngCore, OsRng};
 use sp_core::{sr25519, Pair};
 use subxt::utils::AccountId32;
 use synedrion::{
-    make_interactive_signing_session, make_key_gen_session, make_key_resharing_session,
+    make_aux_gen_session, make_interactive_signing_session, make_key_init_session,
+    make_key_resharing_session,
     sessions::{FinalizeOutcome, Session},
     signature::{self, hazmat::RandomizedPrehashSigner},
     AuxInfo, KeyResharingInputs, NewHolder, OldHolder, PrehashedMessage, RecoverableSignature,
@@ -40,6 +41,7 @@ pub type ChannelOut = Broadcaster;
 /// Thin wrapper broadcasting channel out and messages from other nodes in
 pub struct Channels(pub ChannelOut, pub ChannelIn);
 
+#[derive(Clone)]
 struct PairWrapper(sr25519::Pair);
 
 impl signature::Keypair for PairWrapper {
@@ -64,7 +66,7 @@ impl RandomizedPrehashSigner<sr25519::Signature> for PairWrapper {
 async fn execute_protocol_generic<Res: synedrion::MappedResult<PartyId>>(
     mut chans: Channels,
     session: Session<Res, sr25519::Signature, PairWrapper, PartyId>,
-) -> Result<Res::MappedSuccess, GenericProtocolError<Res>> {
+) -> Result<(Res::MappedSuccess, mpsc::Receiver<ProtocolMessage>), GenericProtocolError<Res>> {
     let tx = &chans.0;
     let rx = &mut chans.1;
 
@@ -114,7 +116,7 @@ async fn execute_protocol_generic<Res: synedrion::MappedResult<PartyId>>(
         }
 
         match session.finalize_round(&mut OsRng, accum)? {
-            FinalizeOutcome::Success(res) => break Ok(res),
+            FinalizeOutcome::Success(res) => break Ok((res, chans.1)),
             FinalizeOutcome::AnotherRound {
                 session: new_session,
                 cached_messages: new_cached_messages,
@@ -161,7 +163,7 @@ pub async fn execute_signing_protocol(
     )
     .map_err(ProtocolExecutionErr::SessionCreation)?;
 
-    Ok(execute_protocol_generic(chans, session).await?)
+    Ok(execute_protocol_generic(chans, session).await?.0)
 }
 
 /// Execute dkg.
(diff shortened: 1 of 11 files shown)


crates/protocol/src/execute_protocol.rs
```diff
@@ -64,7 +66,7 @@ impl RandomizedPrehashSigner<sr25519::Signature> for PairWrapper {
 async fn execute_protocol_generic<Res: synedrion::MappedResult<PartyId>>(
     mut chans: Channels,
     session: Session<Res, sr25519::Signature, PairWrapper, PartyId>,
-) -> Result<Res::MappedSuccess, GenericProtocolError<Res>> {
+) -> Result<(Res::MappedSuccess, mpsc::Receiver<ProtocolMessage>), GenericProtocolError<Res>> {
```
ameba23, line 69: In order to keep the channels for the next of the 3 protos, we return the receiver for incoming messages.

crates/protocol/src/execute_protocol.rs
```diff
@@ -175,20 +192,111 @@ pub async fn execute_dkg(
...
+                .map_err(ProtocolExecutionErr::SessionCreation)?;
+
+        let (init_keyshare, rx) = execute_protocol_generic(chans, session).await?;
+
+        // Setup channels for the next session
+        let chans = Channels(broadcaster.clone(), rx);
+
+        // Send verifying key
+        let verifying_key = init_keyshare.verifying_key();
+        for party_id in party_ids.iter() {
```
ameba23, line 223: Here everyone who joined the key_init protocol sends the verifying key to everyone who wasn't in the key_init protocol. This means they will receive it multiple times.  Maybe better would be to deterministically appoint one person to be the verifying key sender for this session, so it only gets sent once per party. I can make an issue for this.

crates/protocol/src/protocol_message.rs
```diff
@@ -28,8 +28,17 @@ pub struct ProtocolMessage {
...
     /// Identifier of the destination of this message
     pub to: PartyId,
+    /// Either a synedrion protocol message or a verifying key
+    /// We need to send verifying keys during DKG to parties who were not present for the key init
+    /// session
+    pub message_or_verifying_key: MessageOrVerifyingKey,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum MessageOrVerifyingKey {
```
HCastano, line 38: If we ever want to add a variant here this'll become confusing pretty quick. I suggest we change it to something like: `MessageType`, `MessageVariant`, `BroadcastResult`

crates/protocol/src/execute_protocol.rs
```diff
@@ -175,20 +192,111 @@ pub async fn execute_dkg(
...
+            Some(OldHolder { key_share: init_keyshare.to_threshold_key_share() }),
+            chans,
+        )
+    } else {
+        // Wait to receive verifying_key
+        let mut rx = chans.1;
+        let message = rx.recv().await.ok_or_else(|| {
+            ProtocolExecutionErr::IncomingStream("Waiting for validating key".to_string())
+        })?;
+        if let MessageOrVerifyingKey::VerifyingKey(verifing_key_encoded) =
```
HCastano, line 246:
```suggestion
        if let MessageOrVerifyingKey::VerifyingKey(verifying_key_encoded) =
```

crates/protocol/src/execute_protocol.rs
```diff
@@ -175,20 +192,111 @@ pub async fn execute_dkg(
...
+            // Setup channels for the next session
+            let chans = Channels(broadcaster.clone(), rx);
+            (verifying_key, None, chans)
+        } else {
+            return Err(ProtocolExecutionErr::UnexpectedMessage);
+        }
+    };
+
+    // Now reshare to all n parties
+    let inputs = KeyResharingInputs {
```
HCastano, line 269: Why do we need to do this reshare step immediately after the DKG?

crates/protocol/src/execute_protocol.rs
```diff
@@ -175,20 +192,111 @@ pub async fn execute_dkg(
     chans: Channels,
     threshold_pair: &sr25519::Pair,
     threshold_accounts: Vec<AccountId32>,
+    threshold: usize,
```
HCastano, line 195: We may want to update the `fields` of the tracing macro above to match the new signature

crates/protocol/src/protocol_message.rs
```diff
@@ -28,8 +28,17 @@ pub struct ProtocolMessage {
     pub from: PartyId,
     /// Identifier of the destination of this message
     pub to: PartyId,
+    /// Either a synedrion protocol message or a verifying key
+    /// We need to send verifying keys during DKG to parties who were not present for the key init
+    /// session
```
HCastano, line 33:
```suggestion
    /// Either a Synedrion protocol message or a verifying key.
    ///
    /// We need to send verifying keys during DKG to parties who were not present for the key init
    /// session.
```

HCastano (commented):
Haven't had a chance to look at everything yet, will continue tomorrow.

crates/protocol/src/execute_protocol.rs
JesseAbram: to my understanding in synedrion to do threshold keys you need to do a dkg with no threshold then reshare it and make it a t of n with the resharing

JesseAbram (approved):

crates/protocol/src/execute_protocol.rs
```diff
@@ -96,13 +101,25 @@ async fn execute_protocol_generic<Res: synedrion::MappedResult<PartyId>>(
...
+                        session.current_round()
+                    ))
+                })?;
+
+                if let MessageOrVerifyingKey::CombinedMessage(payload) =
+                    message.message_or_verifying_key
+                {
+                    break (message.from, *payload);
+                } else {
+                    tracing::warn!("Got verifying key during protocol - ignoring");
```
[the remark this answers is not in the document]
ameba23: Yes but we have to handle this case.  Currently everyone in the key init sends it to everyone who wasn't, and the ones who were not proceed as soon as they get the first one, meaning the others pop up here.  I'm planning to handle this better in a follow up.

crates/protocol/src/protocol_message.rs
ameba23: Agree, i will change it for now. Long term i am hoping that @fjarri can come up with a way we can put the verifying key in a `CombinedMessage` and we wont need this type.

crates/protocol/tests/protocol.rs
```diff
@@ -31,6 +32,7 @@ mod helpers;
 use helpers::{server, ProtocolOutput};
 
 #[test]
+#[serial]
```
HCastano, line 35: How come you had to add this to the existing tests?

crates/protocol/src/execute_protocol.rs
HCastano: We would want some redundancy or fallback mechanism here though - what if the designated party drops offline during this phase?

crates/protocol/src/execute_protocol.rs
```diff
@@ -226,7 +330,32 @@ pub async fn execute_proactive_refresh(
         make_key_resharing_session(&mut OsRng, &shared_randomness, pair, &party_ids, &inputs)
             .map_err(ProtocolExecutionErr::SessionCreation)?;
 
-    let new_key_share = execute_protocol_generic(chans, session).await?;
+    let new_key_share = execute_protocol_generic(chans, session).await?.0;
 
     new_key_share.ok_or(ProtocolExecutionErr::NoOutputFromReshareProtocol)
 }
+
+/// Psuedo-randomly select a subset of the parties of size t
```
HCastano, line 338:
```suggestion
/// Psuedo-randomly select a subset of the parties of size `threshold`
```

HCastano (approved):
I can't claim to understand all the nitty gritty details here, but from what I do understand it looks good 👍

crates/protocol/tests/protocol.rs
ameba23: I've seen a bit of wierdness with this sometimes not working in CI when it seems to work locally when i run them individually. TBH im not sure whether this was the problem or not. 

but i feel like since these are supposed used for benchmarks, its silly to run them all at the same time using the max amount of CPU cores available.

crates/protocol/src/execute_protocol.rs
ameba23: We need everybody for the next session so if anybody drops off its game over and we have to start fresh.

We can take this conversation to the issue:
https://github.com/entropyxyz/entropy-core/issues/878

PR #870, opened 2024-05-31
Add validator helpers to cli

### LINKED ISSUE #865: Add validator helper functions to the Cli
The rust CLI internally has been talked about helping validators do certain actions, however, those actions are not in the CLI yet, so adding functions like change_threshold address or change threshold server info is needed

JesseAbram:
Closes #865


crates/test-cli/src/lib.rs
```diff
@@ -134,9 +130,26 @@ enum CliCommand {
...
+        #[arg(short, long)]
+        mnemonic_option: Option<String>,
+    },
+    /// Allows a validator to change their endloint
+    ChangeEndpoint {
+        /// New endpoint to change to
+        new_endpoint: String,
+        // The mnemonic to use for the call, should be stash address
+        #[arg(short, long)]
+        mnemonic_option: Option<String>,
```
ameba23, line 143: As I already said with the other ones, i think `--mnemonic-option` is not a good name for this argument. In the context of this CLI this a mandatory argument. If they don't give it the command will not work.

ameba23 (approved):
:100: Great to have these available in the CLI. I'd be curious to try if changing an endpoint works.

As mentioned in the comments I would prefer if these were functions in `entropy-client` which are called here.

They should maybe also be documented in the readme.

crates/test-cli/src/lib.rs
```diff
@@ -134,9 +130,26 @@ enum CliCommand {
...
-        /// A name from which to generate a keypair, eg: "Alice"
-        ///
-        /// Optionally may be preceeded with "//", eg: "//Alice"
+        /// The mnemonic to use for the call
+        #[arg(short, long)]
+        mnemonic_option: Option<String>,
+    },
+    /// Allows a validator to change their endloint
+    ChangeEndpoint {
+        /// New endpoint to change to
```
HCastano, line 139: I would give an example of the expected format. Although I guess that can take all sorts of different forms, from bare IP addresses to FQDNs

crates/test-cli/src/lib.rs
```diff
@@ -134,9 +130,26 @@ enum CliCommand {
...
         /// The path to a file containing the program aux interface (defaults to empty)
         aux_data_interface_file: Option<PathBuf>,
-        /// A name from which to generate a keypair, eg: "Alice"
-        ///
-        /// Optionally may be preceeded with "//", eg: "//Alice"
+        /// The mnemonic to use for the call
+        #[arg(short, long)]
+        mnemonic_option: Option<String>,
+    },
+    /// Allows a validator to change their endloint
```
HCastano, line 137:
```suggestion
    /// Allows a validator to change their endpoint
```

crates/test-cli/src/lib.rs
```diff
@@ -134,9 +130,26 @@ enum CliCommand {
...
+    ChangeEndpoint {
+        /// New endpoint to change to
+        new_endpoint: String,
+        // The mnemonic to use for the call, should be stash address
```
HCastano, line 141: And I guess to be more specific, this is the _validator_ stash account, correct?

crates/test-cli/src/lib.rs
```diff
@@ -134,9 +130,26 @@ enum CliCommand {
...
+        mnemonic_option: Option<String>,
+    },
+    /// Allows a validator to change their threhsold accounts
+    ChangeThresholdAccounts {
+        /// New threshold account
+        new_tss_account: String,
+        /// New x25519 public key
+        new_x25519_public_key: String,
+
+        // The mnemonic to use for the call, should be stash address
```
HCastano, line 152:
```suggestion
        /// The mnemonic to use for the call, should be stash address
```

crates/test-cli/src/lib.rs
HCastano: So we shouldn't even have this as an `Option` then, eh?

crates/test-cli/src/lib.rs
```diff
@@ -404,6 +417,64 @@ pub async fn run_command(
...
+
+            let change_endpoint_tx =
+                entropy::tx().staking_extension().change_endpoint(new_endpoint.into());
+            let in_block =
+                submit_transaction_with_pair(&api, &rpc, &user_keypair, &change_endpoint_tx, None)
+                    .await?;
+            let result_event = in_block
+                .find_first::<entropy::staking_extension::events::EndpointChanged>()?
+                .ok_or(anyhow!("Error with transaction"))?;
+            println!("Event result: {:?}", result_event);
```
HCastano, line 439: I'd be a bit more specific with what we're printing here

crates/test-cli/src/lib.rs
```diff
@@ -404,6 +417,64 @@ pub async fn run_command(
...
+        CliCommand::ChangeEndpoint { new_endpoint, mnemonic_option } => {
+            let mnemonic = if let Some(mnemonic_option) = mnemonic_option {
+                mnemonic_option
+            } else {
+                passed_mnemonic.expect("No Mnemonic set")
+            };
+
+            let user_keypair = <sr25519::Pair as Pair>::from_string(&mnemonic, None)?;
+
+            println!("User account: {}", user_keypair.public());
```
HCastano, line 429: I'd also be more specific here (and same for the other command)

HCastano (approved):
Good call on adding these here. Main thing I'd try and address is moving some of the code to the client and potentially adding tests for them

crates/test-cli/src/lib.rs
JesseAbram: no it will work if the mnemonic is in the environment, a mnemonic is needed but it is not needed to be passed in

ameba23: Ah ok yeah you are right.

crates/test-cli/src/lib.rs
JesseAbram: well this one returns an event that is pretty descriptive and has the extra data needed

PR #871, opened 2024-06-02
Adds deps to Test CLI install instructions.

johnnymatthews:
Tell the reader to install OpenSSL and PKG Config, along with Debian/Ubuntu and Arch pasteables.

Closes https://github.com/entropyxyz/entropy-core/issues/855

### DIFF
--- a/crates/test-cli/README.md
+++ b/crates/test-cli/README.md
@@ -28,6 +28,34 @@ When using the local docker compose setup, be aware you need to set the TSS host
 127.0.0.1 bob-tss-server
 ```
 
+You'll also need the following packages:
+
+1. OpenSSL:
+
+    ```shell
+    # Debian/Ubuntu
+    sudo apt install libssl-dev
+    ```
+
+    ```shell
+    # Arch
+    # OpenSSL comes pre-installed on most Arch releases.
+    # However, to install a specific version run:
+    sudo pacman -S openss3-3.0
+    ```
+    
+2. `pkg-config`:
+
+    ```shell
+    # Debian/Ubuntu
+    sudo apt install pkg-config
+    ```
+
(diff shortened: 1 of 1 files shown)


ameba23 (approved):
Great.  If needing these is an issue we could probably remove the dependency on openssl quite easily as we don't actually use TLS for anything.  But i expect most people have it installed anyway.

crates/test-cli/README.md
```diff
@@ -28,6 +28,34 @@ When using the local docker compose setup, be aware you need to set the TSS host
...
+    # Debian/Ubuntu
+    sudo apt install libssl-dev
+    ```
+
+    ```shell
+    # Arch
+    # OpenSSL comes pre-installed on most Arch releases.
+    # However, to install a specific version run:
+    sudo pacman -S openss3-3.0
+    ```
```
HCastano, line 45: So tbh I wouldn't bother with Arch installation steps. Instead I would replace this with Brew ones (assuming it's not installed by default)

crates/test-cli/README.md
```diff
@@ -28,6 +28,34 @@ When using the local docker compose setup, be aware you need to set the TSS host
...
+
+    ```shell
+    # Debian/Ubuntu
+    sudo apt install pkg-config
+    ```
+
+    ```shell
+    # Arch
+    sudo pacman -S pkgconf
+    ```
```
HCastano, line 57: Same thing here

crates/test-cli/README.md
ameba23: ooh its getting political. don't bother with the brew ones, go for nixos.

HCastano (approved):
Thanks!

PR #879, opened 2024-06-04
Use t of n signing in `entropy-tss`

ameba23:
This adds t of n signing to `entropy-tss`.

Currently the setup is as follows:
- t is set to be 75% of the number of tss nodes, truncating to integer (meaning 2 of 3 in our 3 node test setup).
- Although t is stored as part of the keyshare, we don't (yet) use it when selecting validators for signing, we assume it is constant. This means if the number of validators changes, everything breaks. I don't want to address that in this PR, as that would involve some design decision we should make together. Im trying to just add t of n signing in an un-opinionated way.
- The signing committee is selected by taking it the list of validators from a `entropy::storage().session().validators()` query, sorting it, and taking the first t validators. This means we always use the same nodes for signing, which is bad, but i think we should fix this in a follow up - eg: by moving back to using the message hash to pick a subset.

## Changes to `spawn_testing_validators`

Another thing this PR does is de-duplicate the `spawn_testing_validators` and `create_clients` functions from `testing-utils` and `entropy-tss`.  I needed to make quite a few changes to these functions and having to do that in two places was not nice.
 
We needed to have 2 copies of these because for integration tests we needed them to be accessible when the 'test' config attribute was false. But only having them in `testing-utils` made issues with cfg(test) not being true there.  So i have added a `test_helpers` feature flag to `entropy-tss` which exposes `spawn_testing_validators` even if `cfg(test)` is false.

## Changes to `Broadcaster`

This a bit hacky. There were issues with protocol messages popping up in the wrong sub-session.  When doing DKG, the first participant to finish the reshare and start the aux gen would send the first aux gen message whilst the other 2 parties were still exchanging reshare messages. Since the same websocket connection and channels are used for both protocols, the message would appear in the wrong session.

My way of handling this is to put any message from a session other than the current one back into the incoming message channel.  To do this i cloned the sender and kept it together with the outgoing message sender, so that in the protocol loop i could use it to put messages we dont want yet back into the incoming message queue.

So `Broadcaster` now has an extra field with this sender.

## Changes to `ProtocolMessage`

`ProtocolMessage` now includes a session ID hash so we can know which sub-session the message is intended for.  This shouldn't really be needed as synedrion messages already contain a similar hash, but don't expose it publicly.  I have made an issue in synedrion to expose this to us so we don't need to duplicate data that goes over the wire.
https://github.com/entropyxyz/synedrion/issues/123


ameba23:
Stepping up to 3 nodes in tests, i have had a few issues with timeouts - things take quite a bit longer and sometimes don't happen before a timeout gets reached.

This has been an issue for TSS nodes submitting register confirmations.  To save extra api calls to the chain, we compute the nonce before starting all DKGs, and then just increment it for each register confirmation that gets sent out.  In some cases i get a bad signature error, and i think the issue is that DKG takes so long that this nonce becomes stale as it is based on a block hash from before we start the protocol.

ameba23:
@fjarri i am getting a 'no entry found for key' error from synedrion when running a test for t of n signing, which uses pre-generated test keyshares made with a helper fn: https://github.com/entropyxyz/entropy-core/blob/445e350d19c0652370e37adf66000ef4a413765a/crates/testing-utils/src/create_test_keyshares.rs#L30

This helper fn runs uses `KeyShare::new_centralized` with a given signing key, so that we get keyshares with a known verifying key. It then runs the re-share and aux gen protocols using the test setup copied from the synedrion integration tests.

I think the error must be caused a mismatch of the wrong keyshare and / or aux data being given to the wrong party.  But i have tried almost everything i can think of to try to ensure that that is not the case.  It would be nice if `ThresholdKeyShare` and `AuxData` had `owner()` methods, like `KeyShare` does, so we can check this.

```
thread 'integration_test_sign_public' panicked at /home/turnip/.cargo/git/checkouts/synedrion-a33f7e087cfadddd/2537311/synedrion/src/entities.rs:377:36:
no entry found for key
```

The failing test is here: https://github.com/entropyxyz/entropy-core/blob/445e350d19c0652370e37adf66000ef4a413765a/crates/threshold-signature-server/tests/sign.rs#L38

ameba23:
> @fjarri i am getting a 'no entry found for key' error from synedrion when running a test for t of n signing, which uses pre-generated test keyshares made with a helper fn:

@fjarri - actually you can ignore this - the problem is not with synderion - i was giving the wrong party ids.

node/cli/src/service.rs
```diff
@@ -254,6 +254,7 @@ pub fn new_partial(
 }
 
 /// Result of [`new_full_base`].
+#[allow(dead_code)]
```
ameba23, line 257: Unrelated to this PR - i had to add it because of some new clippy rule.  We should probably investigate why this is dead code.

runtime/src/lib.rs
```diff
@@ -1136,9 +1136,9 @@ impl pallet_sudo::Config for Runtime {
 }
 
 parameter_types! {
-  pub const ImOnlineUnsignedPriority: TransactionPriority = TransactionPriority::max_value();
+  pub const ImOnlineUnsignedPriority: TransactionPriority = TransactionPriority::MAX;
```
ameba23, line 1139: Also for clippy

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -138,17 +142,17 @@ async fn test_get_signer_does_not_throw_err() {
...
 }
+
 #[tokio::test]
 #[serial]
 async fn test_sign_tx_no_chain() {
     initialize_test_logger().await;
     clean_tests();
 
-    let one = AccountKeyring::Dave;
+    let one = AccountKeyring::Eve;
```
ameba23, line 152: Ive changed Dave to Eve in a lot of the signing tests, because originally i only had one set of pre-generated shares - for Eve.  Later i realised the proactive refresh test needs Dave too so i have added Dave's keyshares as well. So if we wanted we could change these back to dave.

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -592,19 +626,10 @@ async fn test_store_share() {
...
-        UnsafeQuery::new(hex::encode(new_verifying_key.to_vec()), [].to_vec()).to_json();
-    // check get key before registration to see if key gets replaced
-    let response_key = client
-        .post("http://127.0.0.1:3001/unsafe/get")
-        .header("Content-Type", "application/json")
-        .body(get_query.clone())
-        .send()
-        .await
-        .unwrap();
+    let response_key = unsafe_get(&client, hex::encode(&new_verifying_key), 3001).await;
```
ameba23, line 629: This hasnt changed functionally i just added a helper to do the get request

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -1083,6 +1099,133 @@ async fn test_increment_or_wipe_request_limit() {
     clean_tests();
 }
 
+#[tokio::test]
+#[serial]
+async fn test_threshold_dkg_and_sign_with_3_nodes() {
```
ameba23, line 1104: This test is just the `store_share` test plus sign a message.  I wrote it before i had the pre-generated shares as this was the only way to test signing. So maybe we don't need this test anymore, as everything it tests is repeated elsewhere.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ At the moment this project **does not** adhere to
 
 ## [[Unreleased]](https://github.com/entropyxyz/entropy-core/compare/release/v0.1.0...master)
 
+### Breaking Changes
+
+- In [#879](https://github.com/entropyxyz/entropy-core/pull/879) threshold signing is introduced,
+  using a fix threshold based on the number of validators.
```
ameba23, line 15: I am gonna think up a better explanation as this is a big breaking change

ameba23:
@JesseAbram @HCastano i have marked this as ready for review, even though the `store_share` test has one time failed with an issue i still can't figure out. I can't recreate it locally. Its is an invalid signature error when submitting the register transaction.  So feels like its not really related to this PR.

See: https://app.circleci.com/pipelines/github/entropyxyz/entropy-core/4432/workflows/fd5ddac6-f9a4-455a-9b3a-18ea7fbd7119/jobs/16098?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary

If either of you can think what could be causing this that would be great.

crates/client/src/user.rs
```diff
@@ -46,13 +46,19 @@ pub async fn get_signers_from_chain(
...
+    let mut validators = query_chain(api, rpc, all_validators_query, None)
         .await?
         .ok_or_else(|| SubgroupGetError::ChainFetch("Get all validators error"))?;
     let block_hash = rpc.chain_get_block_hash(None).await?;
     let mut handles = Vec::new();
 
-    for validator in all_validators {
+    // For now we use a fix proportion of the number of validators as the threshold
+    let threshold = (validators.len() as f32 * 0.75) as usize;
+    // For now we just take the first t validators as the ones to perform signing
```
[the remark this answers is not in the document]
JesseAbram: and in the issue state that this should be handled by chain (probably in the params pallet)

crates/protocol/src/execute_protocol.rs
```diff
@@ -195,24 +206,28 @@ pub async fn execute_dkg(
...
+    party_ids.sort();
 
     let pair = PairWrapper(threshold_pair.clone());
 
     let my_party_id = PartyId::new(AccountId32(threshold_pair.public().0));
 
-    let shared_randomness = session_id.blake2()?;
-    let (key_init_parties, includes_me) =
-        get_key_init_parties(&my_party_id, threshold, &party_ids, &shared_randomness)?;
+    let session_id_hash = session_id.blake2(Some("key_init"))?;
```
JesseAbram, line 217: Can this be an enum instead of a string?

JesseAbram (approved):

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -201,10 +155,36 @@ pub async fn spawn_testing_validators(
...
+/// Add the pre-generated test keyshares to a kvdb
+async fn put_keyshares_in_db(holder_name: &str, kvdb: KvManager) {
+    let user_names_and_verifying_keys = [("eve", EVE_VERIFYING_KEY), ("dave", DAVE_VERIFYING_KEY)];
+
+    for (user_name, user_verifying_key) in user_names_and_verifying_keys {
+        let keyshare_bytes = {
+            let project_root =
+                project_root::get_project_root().expect("Error obtaining project root.");
+            let file_path = project_root.join(format!(
+                "crates/testing-utils/keyshares/production/{}-keyshare-held-by-{}.keyshare",
```
HCastano, line 180: Why are the keyshares with production parameters used over the ones with the test parameters?

crates/threshold-signature-server/src/user/tests.rs
HCastano: If it's not too much work I would revert this in order to keep the diff more manageable

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -259,3 +239,18 @@ pub async fn run_to_block(rpc: &LegacyRpcMethods<EntropyConfig>, block_run: u32)
         current_block = rpc.chain_get_header(None).await.unwrap().unwrap().number;
     }
 }
+
+/// Get a value from a kvdb using unsafe get
+#[cfg(test)]
+pub async fn unsafe_get(client: &reqwest::Client, query_key: String, port: u32) -> Vec<u8> {
```
HCastano, line 245: 👍

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -118,15 +111,10 @@ pub async fn create_clients(
...
 
-pub async fn spawn_testing_validators(
-    passed_verifying_key: Option<Vec<u8>>,
-    // If this is given a keyshare for the user will be generated and returned
-    extra_private_key: Option<sr25519::Pair>,
-    // If true keyshare and verifying key is deterministic
-    deterministic_key_share: bool,
-) -> (Vec<String>, Vec<PartyId>, Option<KeyShareWithAuxInfo>) {
+/// Spawn 3 TSS nodes with pre-stored keyshares
+pub async fn spawn_testing_validators() -> (Vec<String>, Vec<PartyId>) {
```
HCastano, line 115: Not quite related to this line, but this reminded me of something: I wonder if we have to change the `integration-tests` chainspec now to reflect this change (and I guess more broadly the changes going into the main feature branch)

crates/testing-utils/src/tss_server_process.rs
HCastano, line 1: 👍

crates/protocol/src/lib.rs
```diff
@@ -196,9 +200,23 @@ impl Hash for SessionId {
 
 impl SessionId {
     /// Take the hash of the session ID - used as uniqueness in the protocol
-    pub fn blake2(&self) -> Result<Vec<u8>, ProtocolExecutionErr> {
+    /// Optionally with some extra data used to identify a sub-session
+    pub fn blake2(&self, sub_session: Option<&str>) -> Result<[u8; 32], ProtocolExecutionErr> {
```
HCastano, line 204: Why not encode the `sub_session` info as part of the `SessionId` type itself?

crates/testing-utils/src/constants.rs
```diff
@@ -25,7 +25,8 @@ lazy_static! {
         hex!["fe65717dad0447d715f660a0a58411de509b42e6efb8375f562f58a554d5860e"].into(); // subkey inspect //Bob//stash
     pub static ref TSS_ACCOUNTS: Vec<AccountId32> = vec![
         hex!["306bdb49cbbe7104e3621abab3c9d31698b159f48dafe567abb7ea5d872ed329"].into(),
-        hex!["2cbc68e8bf0fbc1c28c282d1263fc9d29267dc12a1044fb730e8b65abc37524c"].into()
+        hex!["2cbc68e8bf0fbc1c28c282d1263fc9d29267dc12a1044fb730e8b65abc37524c"].into(),
+        hex!["946140d3d5ddb980c74ffa1bb64353b5523d2d77cdf3dc617fd63de9d3b66338"].into(),
```
HCastano, line 29: Would be nice to add some comments around what private keys/mnemonics were used for this

crates/testing-utils/src/constants.rs
```diff
@@ -40,6 +41,12 @@ lazy_static! {
         ]
         .try_into()
         .unwrap(),
+        vec![
+            131, 8, 162, 77, 237, 245, 226, 179, 250, 79, 121, 250, 174, 181, 227, 122, 205, 181, 188,
```
HCastano, line 45: Same here

crates/testing-utils/src/create_test_keyshares.rs
```diff
@@ -41,12 +44,9 @@ pub async fn create_test_keyshares(
...
 
-    let keyshares = KeyShare::<TestParams, PartyId>::new_centralized(
-        &mut OsRng,
-        &old_holders,
-        Some(&signing_key),
-    );
-    let aux_infos = AuxInfo::<TestParams, PartyId>::new_centralized(&mut OsRng, &all_parties);
+    let keyshares =
+        KeyShare::<Params, PartyId>::new_centralized(&mut OsRng, &old_holders, Some(&signing_key));
+    let aux_infos = AuxInfo::<Params, PartyId>::new_centralized(&mut OsRng, &all_parties);
```
HCastano, line 49: We touched on this today, but can you remind me what exactly is contained as part of this `AuxInfo`?

crates/threshold-signature-server/src/user/tests.rs
HCastano: If we're already testing DKS and signing elsewhere then ya I'd get rid of this

scripts/create-test-keyshares/Cargo.toml
```diff
@@ -10,7 +10,10 @@ edition    ='2021'
...
 [dependencies]
 entropy-testing-utils={ version="0.1.0", path="../../crates/testing-utils" }
-tokio                ={ version="1.36", features=["macros", "fs", "rt-multi-thread", "io-util", "process"] }
-sp-keyring           ="34.0.0"
-entropy-shared       ={ version="0.1.0", path="../../crates/shared" }
-entropy-kvdb         ={ version="0.1.0", path="../../crates/kvdb", default-features=false }
+tokio={ version="1.36", features=["macros", "fs", "rt-multi-thread", "io-util", "process"] }
+entropy-shared={ version="0.1.0", path="../../crates/shared" }
+entropy-kvdb={ version="0.1.0", path="../../crates/kvdb", default-features=false }
+synedrion={ git="https://github.com/entropyxyz/synedrion", rev="25373111cbb01e1a25d8a5c5bb8f4652c725b3f1" }
```
HCastano, line 16: What branch/tag/whatever is this pointing to? Might be worth adding as comment

HCastano (commented):
I've only taken a first pass at this, I'm gonna need another pass tomorrow before I fully wrap my head around the changes 🙇

crates/threshold-signature-server/src/helpers/tests.rs
ameba23: See https://github.com/entropyxyz/entropy-core/issues/891

crates/protocol/src/lib.rs
ameba23: We maybe could, but we use it in 2 places where we don't want that information because we treat these 3 sub-sessions as part of a single 'DKG' session: In the initial handshake when making a protocol connection (`SubscribeMessage`) and in our mapping of `Listener`s in `entropy-tss` which we use to see if we are expecting a session based on the incoming `SubscribeMessage`

scripts/create-test-keyshares/Cargo.toml
ameba23: It was head of master at the point of starting https://github.com/entropyxyz/entropy-core/pull/868 - can add a comment.  Or maybe @fjarri is gonna make a release for us...

crates/threshold-signature-server/src/helpers/tests.rs
ameba23: Not sure what you mean - the chainspec has changed here: https://github.com/entropyxyz/entropy-core/pull/879/files#diff-e16766add5f3a0eb4dd6bf816b7b64064a417a3e311c8938889cda49a4924137 - maybe you didnt see that bit yet

JesseAbram:
> Stepping up to 3 nodes in tests, i have had a few issues with timeouts - things take quite a bit longer and sometimes don't happen before a timeout gets reached.
> 
> This has been an issue for TSS nodes submitting register confirmations. To save extra api calls to the chain, we compute the nonce before starting all DKGs, and then just increment it for each register confirmation that gets sent out:
> 
> https://github.com/entropyxyz/entropy-core/blob/8bd42e580d7fac737a90c936b70edd3d693bc056/crates/threshold-signature-server/src/user/api.rs#L374
> 
> In some cases i get a bad signature error, and i think the issue is that DKG takes so long that this nonce becomes stale as it is based on a block hash from before we start the protocol.
> 
> In this PR i have made it make a new nonce at the point of making the register confirmation - which involves an extra call per DKG. I guess once everything seems to be working we can try putting it back to how it was.

this is fine, although nonce are not based on blockhash so this is confusing to me

crates/protocol/src/execute_protocol.rs
```diff
@@ -109,13 +112,21 @@ async fn execute_protocol_generic<Res: synedrion::MappedResult<PartyId>>(
...
+                    }
                 } else {
                     tracing::warn!("Got verifying key during protocol - ignoring");
                 }
             };
-
+            // Put messages which were not for this session back onto the incoming message channel
+            for message in messages_for_later.into_iter() {
+                tx.incoming_sender.send(message).await?;
+            }
```
HCastano, line 129: I don't totally follow what's going on here. Is this:
1. Sending the protocol messages back to itself, or
2. Using a cloned version of the original sender and sending a message from that original sender again?

crates/protocol/src/execute_protocol.rs
```diff
@@ -163,11 +174,11 @@ pub async fn execute_signing_protocol(
 
     let pair = PairWrapper(threshold_pair.clone());
 
-    let shared_randomness = session_id.blake2()?;
+    let session_id_hash = session_id.blake2(None)?;
```
HCastano, line 177: Question for a few lines up: why don't we sort the `party_ids` here, but we do in the DKG protocol?

crates/protocol/src/lib.rs
```diff
@@ -196,9 +200,34 @@ impl Hash for SessionId {
 
 impl SessionId {
     /// Take the hash of the session ID - used as uniqueness in the protocol
-    pub fn blake2(&self) -> Result<Vec<u8>, ProtocolExecutionErr> {
+    /// Optionally with some extra data used to identify a sub-session
+    pub fn blake2(
+        &self,
+        sub_session: Option<DkgSubsession>,
```
HCastano, line 206: Okay so these subsessions are only for DKG? No other protocol (signing or refresh) has any?

crates/protocol/src/lib.rs
```diff
@@ -196,9 +200,34 @@ impl Hash for SessionId {
...
+        Ok(hasher.finalize().into())
     }
 }
+
+/// A sub-protocol of the DKG protocol
+#[derive(Debug)]
+pub enum DkgSubsession {
+    KeyInit,
+    Reshare,
+    AuxGen,
```
HCastano, line 222: Some docs here would be nice

crates/protocol/src/protocol_transport/broadcaster.rs
```diff
@@ -14,16 +14,27 @@
...
 /// A wrapper around [broadcast::Sender] for broadcasting protocol messages
 #[derive(Debug, Clone)]
-pub struct Broadcaster(pub broadcast::Sender<ProtocolMessage>);
+pub struct Broadcaster {
+    /// Channel for outgoing protocol messages to all parties
+    pub broadcast: broadcast::Sender<ProtocolMessage>,
+    /// Channel for incoming protocol messages from all parties
+    /// A clone of the sender is kept here so that we can use it in the session loop to put messages
+    /// destined for a different sub-session back into the incoming queue
+    pub incoming_sender: mpsc::Sender<ProtocolMessage>,
```
HCastano, line 32: Okay so the `sender` here is the original sender of the message - but because it was for a different sub-session we use this to re-queue the message?

HCastano (commented):
I need one more pass sorry lol

crates/protocol/src/execute_protocol.rs
ameba23: Its 1.  It is putting the message back into the channel which takes incoming messages from the task which process websocket messages to here. So it is just like putting the message back into the incoming message queue, to be processed in the next session.

crates/protocol/src/execute_protocol.rs
ameba23: The vec of account ids are already sorted further up the stack - well actually i think they are not sorted, but instead of taking them in the order the user gives them we use the order we get from a query to the chain.  which is (hopefully) always the same.  We could sort them here too to be sure though.

crates/protocol/src/lib.rs
ameba23: Thats right.

crates/protocol/src/protocol_transport/broadcaster.rs
ameba23: Yeah. 'Sender' meaning the sending part of the channel.  because the 'Receiver' does not give us a method to put messages back into the channel.  But because with mpsc you can clone the sender as many times as you want, we can keep a sender both in the task which handles ws messages and in the task which runs the protocol.

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -311,16 +304,17 @@ async fn setup_dkg(
...
-        confirm_registered(
-            &api,
-            rpc,
-            sig_request_address,
-            &signer,
-            verifying_key,
-            nonce + i as u32,
-        )
-        .await?;
+        confirm_registered(&api, rpc, sig_request_address, &signer, verifying_key, nonce).await?;
```
HCastano, line 317: Why did this change from `nonce + 1` to just `nonce` (I don't remember the original reason for the `+1`)?

node/cli/src/chain_spec/dev.rs
```diff
@@ -50,10 +50,11 @@ pub fn development_config() -> ChainSpec {
             vec![
                 crate::chain_spec::authority_keys_from_seed("Alice"),
                 crate::chain_spec::authority_keys_from_seed("Bob"),
+                crate::chain_spec::authority_keys_from_seed("Charlie"),
             ],
             vec![],
             get_account_id_from_seed::<sr25519::Public>("Alice"),
-            vec!["127.0.0.1:3001", "127.0.0.1:3002"],
+            vec!["127.0.0.1:3001", "127.0.0.1:3002", "127.0.0.1:3003"],
```
HCastano, line 57: As a follow-up we'll also need to  change the `devenet_local_config()`. This would go hand in hand with changes to the Docker Compose setup

node/cli/src/chain_spec/integration_tests.rs
HCastano, line 1: Do we not need to add `Charlie` to the `integration_tests_genesis_config()` call?

runtime/src/lib.rs
HCastano, line 1: Not entirely sure, but because of these changes we might need to bump the `impl_version`. Can you add that tag to the PR?

HCastano (approved):
Left a few more questions/comments, but otherwise this looks good.

Great work on this 🔥

crates/protocol/src/execute_protocol.rs
ameba23: Right now - bad. Synderion refers to parties by their index in the list, so if different parties order the list differently they get into a muddle. But until now we have consistently got the same ordering back from our chain queries when making them on the server side (with subxt). Also @fjarri is planning to change things so that ordering wont be important anymore, to get around this issue.

crates/threshold-signature-server/src/user/api.rs
ameba23: `nonce + i`. Before we were enumerating the accounts we need to create and creating a nonce once and incrementing it by `i` each time.  

I think we can probably revert this change.  I did it when i couldn't work out why things weren't working, and i hope that that now we can go back to how it was.  But i am not gonna get to it today, so it will have to be for a followup.

node/cli/src/chain_spec/integration_tests.rs
ameba23: I'll do it.  But confusingly we are not actually using it for most of these tests.  For all the tss tests except the proactive refresh one we use `test_context_stationary` which uses the 'dev' chainspec.

ameba23:
> this is fine, although nonce are not based on blockhash so this is confusing to me
@JesseAbram - yeah i think probably we can revert this change. But i am not going to get to it today.

ameba23:
> > this is fine, although nonce are not based on blockhash so this is confusing to me

@JesseAbram - yeah i think probably we can revert this change. But i am not going to get to it today.

crates/threshold-signature-server/src/helpers/tests.rs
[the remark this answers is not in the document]
ameba23: I think theres other places in the tests where we could use this where we make an unsafe get to bob rather than alice, so im gonna leave it like this.

ameba23:
@JesseAbram @HCastano unfortunately i have found a problem with this, which i don't know how to fix.

In experimenting with trying to do a reshare with less then n parties, i noticed that there was a mistake in our the `execute_proactive_refresh` fn. 

The threshold was being set to n rather than t.  That is, it was always being set to 3 when it should be 2.

In this commit just now i changed it to what it should be: https://github.com/entropyxyz/entropy-core/pull/879/commits/f37bea595cfc3fe4f285725f25fd93d881a0501a 

And this causes proactive refresh to fail with:

```
thread 'signing_client::tests::test_proactive_refresh' panicked at /home/turnip/.cargo/git/checkouts/synedrion-a33f7e087cfadddd/2537311/synedrion/src/www02/key_resharing.rs:345:29:
no entry found for key
```
I think to get to the bottom of this issue we need a test which does proactive refresh and then signs a message.

ameba23:
I have found a workaround which is setting the `old_threshold` to `n` and the new one to `t`. So we should still end up with `t` of `n` shares.  So i am happy with this and gonna merge, but i would like to still make an issue about this, either here or in synedrion, and ideally i would like the refresh test to sign a message after refreshing, to be sure that it actually works.

PR #881, opened 2024-06-07
Add blake2 as built in hash function and make `HashingAlgorithm` non-exhaustive

ameba23:
Blake2 is used to hash ECDSA signer payloads on substrate / polkadot. So i think it makes sense to have it as one of our built in hash functions. This PR does that.

Ive also marked the `HashingAlgorithm` enum as `non_exhaustive`.  This means we handle the case that it was some unknown variant and return a user error if it was.  The idea is to future proof - meaning we can add extra variants without breaking validators who didn't yet update their software.  There are probably arguments against doing this, so im willing to take it out if people think its a bad idea.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 At the moment this project **does not** adhere to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [[Unreleased]](https://github.com/entropyxyz/entropy-core/compare/release/v0.1.0...master)
+## [Unreleased](https://github.com/entropyxyz/entropy-core/compare/release/v0.1.0...master)
 
 ### Breaking Changes
 - In [#866](https://github.com/entropyxyz/entropy-core/pull/866) timestamp was removed from `UserSignatureRequest` and replaced with block_number. Thus check_stale now uses block_number for stale checks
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ At the moment this project **does not** adhere to
 - Add a way to change program modification account  ([#843](https://github.com/entropyxyz/entropy-core/pull/843))
 - Add support for `--mnemonic-file` and `THRESHOLD_SERVER_MNEMONIC` ([#864](https://github.com/entropyxyz/entropy-core/pull/864))
 - Add validator helpers to cli ([#870](https://github.com/entropyxyz/entropy-core/pull/870))
+- Add blake2 as built in hash function and make HashingAlgorithm non-exhaustive ([#881](https://github.com/entropyxyz/entropy-core/pull/881)
```
ameba23, line 19: Im not sure if this counts as a breaking change. I think old JSON encoded signature requests will still work. But the shared type `HashingAlgorithm` has changed so strictly speaking this probably is breaking.

HCastano: Yeah this would be breaking for the `entropy-shared` crate since: a) we added a new field to a public enum and b) with the addition of `non_exhaustive` existing callers now need to handle the wildcard case.

We should also add this PR to the `Added` section.

HCastano (approved):
Cool cool.

Do you have any other docs supporting the use of `Blake2` for `ECDSA`? I've found a couple places in the Substrate codebase ([1](https://github.com/paritytech/substrate/blob/8bc45ed4be61b83dcfac77bfc8002858ff8042ec/primitives/runtime/src/lib.rs#L409C1-L409C51), [2](https://github.com/paritytech/polkadot-sdk/blob/9dfe0fee74ce1e4b7f99c1a5122b635aa43a1e5f/substrate/primitives/core/src/ecdsa.rs#L105)), but I was hoping to find a non-source code reference to this as well

ameba23:
> Do you have any other docs supporting the use of `Blake2` for `ECDSA`? I've found a couple places in the Substrate codebase ([1](https://github.com/paritytech/substrate/blob/8bc45ed4be61b83dcfac77bfc8002858ff8042ec/primitives/runtime/src/lib.rs#L409C1-L409C51), [2](https://github.com/paritytech/polkadot-sdk/blob/9dfe0fee74ce1e4b7f99c1a5122b635aa43a1e5f/substrate/primitives/core/src/ecdsa.rs#L105)), but I was hoping to find a non-source code reference to this as well

No.  It looks like the `k256` crate which synedrion is based on uses sha256 by default, so i wouldn't say that blake2 is the standard for ecdsa/secp256k1 by any means:  https://docs.rs/k256/0.13.1/src/k256/schnorr/signing.rs.html#216

JesseAbram (approved):

PR #883, opened 2024-06-07
Add faucet test

JesseAbram:


node/cli/src/chain_spec/integration_tests.rs
```diff
@@ -247,14 +247,15 @@ pub fn integration_tests_genesis_config(
                 ),
                 (
                     get_account_id_from_seed::<sr25519::Public>("Eve"),
-                    1,
-                    Some(crate::chain_spec::tss_x25519_public_key::EVE),
+                    0,
```
JesseAbram, line 250: Not needed to be private and allows me to use the already deterministic generated eve verifying key and shards

node/cli/src/chain_spec/integration_tests.rs
```diff
@@ -247,14 +247,15 @@ pub fn integration_tests_genesis_config(
...
+                    None,
                     BoundedVec::try_from(EVE_VERIFYING_KEY.to_vec()).unwrap(),
                 ),
                 (
                     get_account_id_from_seed::<sr25519::Public>("Ferdie"),
-                    2,
-                    None,
+                    1,
+                    // Can use eve's key, just need a key to talk to and encrypt for proactive refresh
+                    Some(crate::chain_spec::tss_x25519_public_key::EVE),
```
JesseAbram, line 258: A little cheeky but we are removing private soon anyways so idk just ignore

ameba23 (approved):
Great work getting this done and fully tested, i was pretty much ready to give up on the whole idea, at one point it seemed almost impossible.

Not sure if you actually wanted a review yet, so forgive all the nitpicks if this actually isnt yet finished.

As already mentioned in the faucet program repo i would think it would make sense to have the chain specific stuff (genesis hash, spec version, etc) in the program config rather than aux data. But thats not for this PR anyway.

Re: making eve not private mode anymore, i would have thought that would break some of the other tests. But if CI is happy then so i am i, since private mode is anyway on its way out.

JesseAbram:
> Great work getting this done and fully tested, i was pretty much ready to give up on the whole idea, at one point it seemed almost impossible.
> 
> Not sure if you actually wanted a review yet, so forgive all the nitpicks if this actually isnt yet finished.
> 
> As already mentioned in the faucet program repo i would think it would make sense to have the chain specific stuff (genesis hash, spec version, etc) in the program config rather than aux data. But thats not for this PR anyway.
> 
oh I missed that comment but it is because it will change (on our chain and between other substrate chains) so this would allow it to not be redeployed and that info is easily grabbed from chain (although other chains would need to swap metadata). But this way the program is generalizable to substrate chains and update proof 
 
> Re: making eve not private mode anymore, i would have thought that would break some of the other tests. But if CI is happy then so i am i, since private mode is anyway on its way out.
yaaaa I could have added another key but we are removing private mode soon, and I had to do some hacky things but better then adding code we need to remove

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -1681,6 +1687,203 @@ async fn test_device_key_proxy() {
...
+
+    let (validator_ips, _validator_ids, keyshare_option) =
+        spawn_testing_validators(Some(EVE_VERIFYING_KEY.to_vec()), false, true).await;
+    let substrate_context = test_node_process_testing_state(true).await;
+    let entropy_api = get_api(&substrate_context.ws_url).await.unwrap();
+    let rpc = get_rpc(&substrate_context.ws_url).await.unwrap();
+
+    let verifying_key =
+        keyshare_option.clone().unwrap().verifying_key().to_encoded_point(true).as_bytes().to_vec();
+    let verfiying_key_account_string = blake2_256(&verifying_key);
```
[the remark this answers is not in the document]
JesseAbram: maybe can call it hashed verifying key

PR #884, opened 2024-06-10
Pregenerated threshold keyshare set for `entropy-tss` tests

ameba23:
This makes it possible to generate a set of threshold keyshares for `entropy-tss` tests, and adds pregenerated keyshares to the repo.

Our signing tests use these in order to test signing without needing to run the reshare protocol for each test.


crates/protocol/src/execute_protocol.rs
```diff
@@ -45,7 +45,7 @@ pub type ChannelOut = Broadcaster;
 pub struct Channels(pub ChannelOut, pub ChannelIn);
 
 #[derive(Clone)]
-struct PairWrapper(sr25519::Pair);
+pub struct PairWrapper(pub sr25519::Pair);
```
ameba23, line 48: I needed to make this public even though it is only used for making test keyshares

crates/testing-utils/src/create_test_keyshares.rs
```diff
@@ -0,0 +1,273 @@
...
+
+    let mut output = Vec::new();
+
+    for i in 0..3 {
+        output.push((new_t_key_shares[i].clone().unwrap(), aux_infos[i].clone()));
+    }
+    output
+}
+
+type MessageOut = (PartyId, PartyId, CombinedMessage<sr25519::Signature>);
```
ameba23, line 108: From here on, everything is copied from the synedrion integration tests.

Ideally i would not want to duplicate that here.  But its not exposed publicly in synedrion, and i needed to modify a few things.  Another option would be to use our entropy-protocol test setup - but then would either need to create these keyshares in a test, or move that code to `testing-utils` which i don't really want to do.

crates/testing-utils/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+//! The base path for where to store keyshares is given as a single command line argument
+//! If it is not given, the current working directory is used
+use entropy_kvdb::kv_manager::helpers::serialize;
+use entropy_shared::DETERMINISTIC_KEY_SHARE_EVE;
+use entropy_testing_utils::create_test_keyshares::create_test_keyshares;
+use sp_keyring::AccountKeyring;
+use std::{env::args, path::PathBuf};
+
+#[tokio::main]
+async fn main() {
```
ameba23, line 26: Not sure if the main fn of testing-utils is the best place for this, but couldn't think of a better home for it.

scripts/create-test-keyshares.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+# Creates a set of keyshares for use in entropy-tss tests
+
+cargo run -p entropy-testing-utils --release -- ./crates/testing-utils/keyshares
```
ameba23, line 5: This assumes we run the script from the project root. Some of our other scripts also make this assumption so i thought its probably ok.

crates/testing-utils/src/main.rs
HCastano: Yeah I don't really like it here tbh.

Could we make a new crate within the `script/` folder and use that instead?

scripts/create-test-keyshares.sh
HCastano, line 5: Why pick to compile this in `release` mode?

scripts/create-test-keyshares.sh
HCastano, line 1: Under what conditions would we need to re-run this script to get new keyshares? Might worth mentioning as part of the script comments

HCastano (approved):
Main thing I'd like to see changed in the location of the script (ideally into `script/`), but otherwise LGTM

scripts/create-test-keyshares.sh
ameba23: To make it run faster.  although of course it compiles slower so doesnt really help.  i can take it out if you like.

scripts/create-test-keyshares.sh
ameba23: Yeah i'll add something. On updating synedrion a minor/major version, and possibly also on updating `entropy-protocol` if there are breaking changes to eg: `PartyId`.

JesseAbram (approved):

PR #886, opened 2024-06-11
Replace link to private repo with succinct explanation.

### LINKED ISSUE #778: Fragmented docs for running a local network
The README has this line:

```markdown
Additional documentation on running a local network can be found [here](https://github.com/entropyxyz/meta/wiki/Local-devnet)
```
This has the effect of fragmenting the docs. I personally love having all the docs in one place, especially if I'm offline and think "I've cloned everything I need down, should be good to go".

It would be nice if all the docs relevant to the repo were in one place.
I would recommend making `docs/LOCAL_DEV.md` in this repo

vitropy:
Issue #778 describes "fragmented" docs, but the documentation that was linked to does not actually describe in much detail how to run a development network. Rather, it describes the technical architecture of a development network's start up sequence, and at some length. For example, it included message sequence diagrams that are a bit outdated today.

A more immediately useful reference for a user who attempts to spin up an Entropy Network using the local Docker Compose configuration provided here may be to point them directly at the new Entropy CLI tool, and to reference the public-facing documentation that is current. Moreover, the private repository linked is intended for internal Entropy uses only. While that might have been sensible to link to in the past, it no longer is.

While moving some of technical details currently housed within the private repository's wiki might make sense at some point, as issue #778 suggests, I don't think that's of a high enough priority right now to justify the amount of time it would take to present that same material somewhere sensible here, in this repository, else find a new home for it.

For now, it seems more sensible to simply remove the link and provide a reference to the CLI. So, that's what I've done in this commit.

I've also removed the `version` key from the Docker Compose file because modern Docker Engines no longer consider that key when doing schema validation; the `version` key is deprecated and produces a warning to such effect when present. There's no need for it, so it's been removed.

Closes #778.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,10 +38,9 @@ This repository provides a [Docker Compose](https://docs.docker.com/compose/) co
         ```sh
         docker compose logs --follow # Following is also optional.
         ```
-    * [use the Entropy CLI](https://docs.entropy.xyz/reference/cli#install) to interact with the locally running network:
+    * [use the Entropy Test CLI](https://docs.entropy.xyz/reference/rust-testing-interface) to interact with the locally running network:
         ```sh
-        cd /path/to/entropy/cli.source.code
-        yarn start --endpoint ws://127.0.0.1:9944
+        cargo run -p entropy-test-cli -- --chain-endpoint="ws://127.0.0.1:9944" status 
         ```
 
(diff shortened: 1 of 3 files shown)


HCastano:
Sorry for being slow to review this, will take a look tomorrow!

PR #893, opened 2024-06-14
Use test params for keyshares when cfg(test) is true in entropy-tss

ameba23:
Just to see if this fixes the timeout issues i am getting testing DKG - use synedrion test params for `entropy-tss` unit tests.


ameba23:
Closing this for now as it hasn't solved the problem

PR #897, opened 2024-06-18
JS bindings for entropy client

ameba23:
This adds JS bindings to entropy-client.

The idea is to make it easier to migrate the SDK to a newer version of `entropy-core` - as this binds to client code which is all tested in CI in `entropy-core`.  And also to avoid issues with inconsistencies between subxt and polkadot JS - by relying less on polkadot JS. 

There is a very basic [nodejs test CLI for trying this out](https://github.com/entropyxyz/entropy-core/blob/peg/bindings-entropy-client/crates/client/nodejs-test/index.js)

For some reason when using docker-compose i have to specify the chain endpoint as `ws://127.0.0.1:9944` - it does not like `localhost` (gives a connection refused error).

Register is a bit complicated because to poll for registration we need to a way to 'sleep' - rusts `std::thread::sleep` doesn't work.  The `web-sys` crate does gives us access to `window.setTimeout()` but i think i prefer to keep this out of our bindings.

So i split registration into 2 functions.  One which submits the transaction and one which checks for a confirmation with a given account id and returns the verifying key if successful. Running the loop and adding a delay is done externally in JS.

I am occasionally seeing this issue: https://github.com/entropyxyz/entropy-core/issues/542 - which does not effect things working, but means an annoying error message pops up.

Tested manually with nodejs 18.15.0. 

I am not planning to add tests for this because of the issues we had with the JS tests for entropy-protocol. If it this gets used in the SDK it should be tested there, and if it does not get used in the SDK then we probably dont need these bindings.


crates/client/src/errors.rs
```diff
@@ -103,3 +103,10 @@ pub enum ClientError {
     #[error("Verifying key has incorrect length")]
     BadVerifyingKeyLength,
 }
+
+#[cfg(feature = "full-client")]
```
ameba23, line 107: For some reason when compiling for wasm it doesn't like me deriving this trait with thiserror for this particular error type.

ameba23:
I have read the CLA Document and I hereby sign the CLA

johnnymatthews:
I have read the CLA Document and I hereby sign the CLA

ameba23:
Closing this as it seems no-one wants it. Can re-open if that changes.

PR #900, opened 2024-06-20
Add sort to subgroup signer selection

### LINKED ISSUE #890: Inconsistent behaviour between `entropy-core` and `sdk` when selecting validators from message hash
This follows a conversation with @frankiebee about issues with using the SDK with testnet.

I think there might already be an issue for this somewhere that i have not found.

It seems the process for selecting a signing committee from a message hash is working differently using the SDK to how it is done in `entropy-client`, meaning the selected validators are often not 'correct' as the TSS node checks if it is one of the signing commitee using `get_current_subgroup_signers`, and so signing fails.

https://github.com/entropyxyz/entropy-core/blob/90da916499e8c3558aa9d71f71149d3810eaa3f6/crates/client/src/user.rs#L49

Possibly this is due to polkadot js giving the response to the staking extension pallet's `signing_groups` query in a different order to that of the same query made with subxt. Or possibly it is due to to an inconsistency in the sdk's `pickValidators` function with `get_current_subgroup_signers`.

https://github.com/entropyxyz/sdk/blob/6ca59f7289a07a3a4b9ecd056657175a5fdefa17/src/signing/index.ts#L340

If ive understood right there is already a fix for this issue being discussed which would involve sorting the output of the signing groups query in both `get_current_subgroup_signers` and `pickValidators`.

JesseAbram:
Closes #890 

Adds a sort function to help with a more deterministic subgroup get across different libraries and languages, as well adds a helper endpoint to get the subgroup with hash

### DIFF
--- a/crates/threshold-signature-server/src/node_info/api.rs
+++ b/crates/threshold-signature-server/src/node_info/api.rs
@@ -42,6 +42,7 @@ pub async fn get_subgroup_signers(
     State(app_state): State<AppState>,
     message_hash: String,
 ) -> Result<String, UserErr> {
+    tracing::debug!("Message hash {:?}", message_hash);
     let api = get_api(&app_state.configuration.endpoint).await?;
     let rpc = get_rpc(&app_state.configuration.endpoint).await?;
     let subgroup_signers = get_current_subgroup_signers(&api, &rpc, &message_hash).await?;
--- a/crates/threshold-signature-server/src/node_info/api.rs
+++ b/crates/threshold-signature-server/src/node_info/api.rs
@@ -34,11 +34,13 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
     Json(hashing_algos)
 }
 
-/// message is stripped 0x keccak hex encoded hash of message
+/// Returns the list of subgroup signers given a message hash.
+///
+/// Note: the message hash should not include a preceding `0x`.
 #[tracing::instrument(skip_all)]
 pub async fn get_subgroup_signers(
     State(app_state): State<AppState>,
-    message_hash_keccak_hex: String,
+    message_hash: String,
 ) -> Result<String, UserErr> {
     let api = get_api(&app_state.configuration.endpoint).await?;
     let rpc = get_rpc(&app_state.configuration.endpoint).await?;
--- a/crates/threshold-signature-server/src/node_info/api.rs
+++ b/crates/threshold-signature-server/src/node_info/api.rs
@@ -44,7 +44,6 @@ pub async fn get_subgroup_signers(
 ) -> Result<String, UserErr> {
     let api = get_api(&app_state.configuration.endpoint).await?;
     let rpc = get_rpc(&app_state.configuration.endpoint).await?;
-    let subgroup_signers =
-        get_current_subgroup_signers(&api, &rpc, &message_hash_keccak_hex).await?;
+    let subgroup_signers = get_current_subgroup_signers(&api, &rpc, &message_hash).await?;
     Ok(serde_json::to_string(&subgroup_signers)?)
 }
--- a/crates/threshold-signature-server/src/node_info/api.rs
+++ b/crates/threshold-signature-server/src/node_info/api.rs
@@ -34,7 +34,7 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
     Json(hashing_algos)
 }
 
-/// message is keccak hex encoded hash of message
+/// message is stripped 0x keccak hex encoded hash of message
 #[tracing::instrument(skip_all)]
 pub async fn get_subgroup_signers(
     State(app_state): State<AppState>,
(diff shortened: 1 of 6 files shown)


crates/threshold-signature-server/src/node_info/api.rs
```diff
@@ -26,3 +33,16 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
     let hashing_algos = HashingAlgorithm::iter().collect::<Vec<_>>();
     Json(hashing_algos)
 }
+
+/// message is stripped 0x keccak hex encoded hash of message
+#[tracing::instrument(skip_all)]
+pub async fn get_subgroup_signers(
```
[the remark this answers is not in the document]
JesseAbram: ya I mean as a long term solution I don't love this, but if this squashes the problem easy until our next release when we remove all the subgroup stuff then im cool with that compromise

ameba23 (approved):
:100: (Assuming CI gets fixed).

I think ideally we should not need to use the `/subgroup_signers` endpoint - we should be able to sort the array of validator IDs on the client side. That said i don't know how to compare byte arrays lexicographically in pure JS. NodeJS's buffer has a `compare` method which does this though.

crates/threshold-signature-server/src/node_info/api.rs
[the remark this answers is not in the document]
frankiebee: just leaving this here as notes: https://doc.rust-lang.org/std/primitive.slice.html#method.sort

crates/threshold-signature-server/src/node_info/api.rs
```diff
@@ -26,3 +33,16 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
     let hashing_algos = HashingAlgorithm::iter().collect::<Vec<_>>();
     Json(hashing_algos)
 }
+
+/// message is stripped 0x keccak hex encoded hash of message
+#[tracing::instrument(skip_all)]
```
HCastano, line 38: We should include the `message_hash` here, and emit a `debug` event at function entry

crates/threshold-signature-server/src/node_info/api.rs
```diff
@@ -26,3 +33,16 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
     let hashing_algos = HashingAlgorithm::iter().collect::<Vec<_>>();
     Json(hashing_algos)
 }
+
+/// message is stripped 0x keccak hex encoded hash of message
```
HCastano, line 37: This isn't really describing what the API does. I'd recommend something like this:

```suggestion
/// Returns the list of subgroup signers given a message hash.
///
/// Note: the message hash should not include a preceding `0x`.
```

I guess on a sidenote, why not allow the client to pass in the message with the `0x` and just strip it within the function? I think that would be more user friendly

crates/threshold-signature-server/src/node_info/api.rs
```diff
@@ -26,3 +33,16 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
...
+#[tracing::instrument(skip_all)]
+pub async fn get_subgroup_signers(
+    State(app_state): State<AppState>,
+    message_hash_keccak_hex: String,
```
HCastano, line 41:
```suggestion
    message_hash: String,
```

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -1760,7 +1760,7 @@ async fn test_faucet() {
     .await
     .unwrap();
 
-    let amount_to_send = 200000011;
+    let amount_to_send = 200000012;
```
HCastano, line 1763: Why did this change?

crates/threshold-signature-server/src/node_info/tests.rs
```diff
@@ -50,3 +57,31 @@ async fn hashes_test() {
         ]
     );
 }
+
+#[tokio::test]
+#[serial]
+async fn test_get_subgroup() {
```
HCastano, line 63: Not really sure if this is testing the sorting algorithm, it feels like it's just testing that the `subgroup_signers` endpoint calls `get_current_subgroup_signers` internally.

One suggestion here, maybe let's call `get_current_subgroup_signers` directly and test it out with different message hashes and validator sets to see if it matches what we expect.

crates/threshold-signature-server/src/node_info/api.rs
JesseAbram: tend to like to push complexity like that off to the client, plus we are removing this code in a release anyways so rather not push it

crates/threshold-signature-server/src/node_info/tests.rs
JesseAbram: we do in our integration tests, this test, is to test the endpoint

crates/threshold-signature-server/src/user/tests.rs
JesseAbram: because the sorting algorithm changed this is a hacky thing I did to get the proper validator selection by changing the message, this is because I needed the proper verifying keys from one chain spec so instead of creating a whole new chainspec to maintain for one test, I just manipulate the message to select the spun up validators I need........specifically since we are removing it anyways I prefer this method to writing a whole new chainspec

PR #901, opened 2024-06-20
Change currency units

### LINKED ISSUE #888: Review Existential Deposit
It would seem as the ED is 10_000 * 10 ^10 which is probably too high, should lower it to 1 or 0.1

JesseAbram:
Changes currency units to match Polkadot's (which we decided 10 decimals so this should have been changed with that) of note turns existential deposit down to 0.01

Closes #888


ameba23 (approved):
:1st_place_medal:

HCastano (approved):
Code looks fine, but on my end PJS still shows that we have 12 decimals

image

Not entirely sure why though, maybe outdated metadata somewhere 🤷 ?

Also, I wonder if this requires a `spec_version` bump or something 🤔

JesseAbram:
> Code looks fine, but on my end PJS still shows that we have 12 decimals
> 
> image
> Not entirely sure why though, maybe outdated metadata somewhere 🤷 ?
> 
> Also, I wonder if this requires a `spec_version` bump or something 🤔
maybe but we def have a spec bump in the next release so let's not overthink it 

image

fixed it, we were not adding the properties to the dev chainspec

PR #902, opened 2024-06-21
Create four node Docker Compose chainspec

JesseAbram:
This creates a local docker large testnet with 4 validators as discussed here

https://github.com/entropyxyz/sdk/pull/383


JesseAbram:
> Thanks so much for putting this together. It's also helpful to have an example of what needs to change in order to produce a new chain spec like this. I have one small question, commented inline.

ya no worries lol but will only be helpful for like mmmmm a few weeks

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -133,6 +138,12 @@ pub struct StartupArgs {
     /// Use the developer key Alice.
     #[arg(short = 'a', long = "alice")]
     pub alice: bool,
+    /// Use the developer key dave.
+    #[arg(long = "dave")]
+    pub dave: bool,
+    /// Use the developer key Eve.
+    #[arg(short = 'e', long = "eve")]
+    pub eve: bool,
```
HCastano, line 146: We might as well add `--charlie` here too

node/cli/src/chain_spec/mod.rs
```diff
@@ -113,6 +118,20 @@ pub mod tss_x25519_public_key {
         142, 113, 91, 59, 177, 104, 208, 23, 219, 170, 47, 145, 200, 139, 188, 28, 14, 199, 116,
         86, 193, 144, 10, 18, 74, 157, 138, 202, 115, 99, 229, 55,
     ];
+
+    // The `DEFAULT_DAVE_MNEMONIC` is used to derive the public key.
+    /// Mnemonic: "beef dutch panic monkey black glad audit twice humor gossip wealth drive"
+    pub const DAVE: [u8; 32] = [
```
HCastano, line 124: Now that this is available, can you also update this line?

https://github.com/entropyxyz/entropy-core/blob/3cbc794f97cbcc0f92cc27363f744a6529b82294/node/cli/src/chain_spec/integration_tests.rs#L178

node/cli/src/chain_spec/testnet.rs
```diff
@@ -54,6 +54,16 @@ pub fn testnet_local_initial_authorities(
     ]
 }
 
+pub fn testnet_local_large_initial_authorities(
+) -> Vec<(AccountId, AccountId, GrandpaId, BabeId, ImOnlineId, AuthorityDiscoveryId)> {
+    vec![
+        crate::chain_spec::authority_keys_from_seed("Alice"),
+        crate::chain_spec::authority_keys_from_seed("Bob"),
+        crate::chain_spec::authority_keys_from_seed("Dave"),
+        crate::chain_spec::authority_keys_from_seed("Eve"),
```
HCastano, line 63: Why is `Charlie` being skipped?

node/cli/src/chain_spec/testnet.rs
HCastano, line 1: This config should actually be added to the `dev` chain spec, not the testnet one. So we should end up with `devnet-local-large` instead.

node/cli/src/command.rs
```diff
@@ -82,6 +83,7 @@ impl SubstrateCli for Cli {
                 Box::new(chain_spec::integration_tests::integration_tests_config())
             },
             "testnet-local" => Box::new(chain_spec::testnet::testnet_local_config()),
+            "testnet-local-large" => Box::new(chain_spec::testnet::testnet_local_config_large()),
```
HCastano, line 86: I would be more specific here: `devnet-local-four-nodes`

HCastano (changes requested):
Main change needed here is that this should be part of the `dev` chainspecs, not the `testnet` ones.

node/cli/src/chain_spec/testnet.rs
JesseAbram: I am following this docker script https://github.com/entropyxyz/sdk/pull/383/files which selects alice bob dave eve

node/cli/src/chain_spec/testnet.rs
JesseAbram: probably the same for testnet-local then (the two node docker script, I kinda just copied off that)

HCastano: The `testnet-local` config was added intentionally to allow for a lightweight way to check the testnet configuration. We should keep that where it is.

node/cli/src/chain_spec/testnet.rs
HCastano: There's gotta be a deeper reason though, no?

JesseAbram: you can compare notes with my therapist? but no, I just was working off the docker script @vitropy created

node/cli/src/chain_spec/testnet.rs
JesseAbram: actually looking at it the dev stuff has a lot of things in there we need for testing, this is meant to function as a testnet with items we do not need, much like the testnet-local, I don't think this makes sense to move it

HCastano: @JesseAbram The reason I'm saying that this should be a `dev` spec is that the current SDK testing setup uses a 
[two node](https://github.com/entropyxyz/sdk/blob/main/dev/docker-scripts/two-nodes.yaml) `dev` spec.

By changing this to a testnet config it could introduce unintended differences into their
testing environment.

So unless the SDK team (@entropyxyz/js-devs) explicit asked for this and/or is okay with this, then I would
suggest moving it.

node/cli/src/chain_spec/integration_tests.rs
```diff
@@ -175,7 +175,7 @@ pub fn integration_tests_genesis_config(
                         crate::chain_spec::tss_account_id::DAVE.clone(),
                         // This should be a `Dave` account, but we use `Bob` since Dave's mnemonic
                         // is unavailable right now.
```
HCastano, line 177: Thanks for updating the key, do you mind removing this comment now?

node/cli/src/chain_spec/testnet.rs
```diff
@@ -176,6 +186,34 @@ pub fn testnet_local_config() -> crate::chain_spec::ChainSpec {
...
 
+/// The configuration used for a local testnet network spun up using the `docker-compose` setup
+/// provided in this repository.
+///
+/// This configuration matches the same setup as the `testnet`, with the exception that is uses
+/// four well-known accounts (Alice Bob Dave and Eve) as the authorities.
+pub fn testnet_local_config_large() -> crate::chain_spec::ChainSpec {
+    ChainSpec::builder(wasm_binary_unwrap(), Default::default())
+        .with_name("Entropy Testnet Local Large")
+        .with_id("entropy_testnet_local_large")
```
HCastano, line 197: This should be changed to "*-four-nodes"

node/cli/src/chain_spec/testnet.rs
JesseAbram: ahhh I see ok will have to update a lot in devnet config but should be fine

HCastano (approved):
FYI Jesse I pushed up some changes.

I tested that the chainspecs are still building by running:

```
 cargo run -p entropy -- build-spec --raw --chain devnet-local > chainspec-dev-raw.json
cargo run -p entropy -- build-spec --raw --chain devnet-local-four-nodes > chainspec-dev-four-nodes-raw.json
```

I haven't actually ran the four node Compose network though.

PR #907, opened 2024-06-24
Ignore faucet test due to race condition

JesseAbram:
Updates the program to the newest version that removes the header which also removes the race condition in subgroup signer selection


JesseAbram:
well this is embarrassing, turns out the aux data wouldn't effect the message since the partial would create a transaction with the blockheader in it anyways, so ignoring until subgroups are removed

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -1833,6 +1829,7 @@ async fn test_faucet() {
         submit_transaction_requests(validator_ips_and_keys.clone(), generic_msg.clone(), one).await;
     let mut decoded_sig: Vec<u8> = vec![];
     for res in test_user_res {
+        // dbg!(res.unwrap().text().await.unwrap());
```
HCastano, line 1832:
```suggestion
```

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -1687,6 +1687,7 @@ async fn test_device_key_proxy() {
     verify_signature(test_user_res, message_hash, keyshare_option.clone()).await;
 }
 
+#[ignore]
```
HCastano, line 1690: Can you open an issue and link it here pls?

HCastano (approved):

crates/threshold-signature-server/src/user/tests.rs
JesseAbram: https://github.com/entropyxyz/entropy-core/issues/909

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -1686,7 +1686,8 @@ async fn test_device_key_proxy() {
         submit_transaction_requests(validator_ips_and_keys.clone(), generic_msg.clone(), one).await;
     verify_signature(test_user_res, message_hash, keyshare_option.clone()).await;
 }
-
+// ignored for now see https://github.com/entropyxyz/entropy-core/issues/909
```
HCastano, line 1689:
```suggestion

/// FIXME (#909): ... explain why it's being ignored ...
```

Would prefer something like this

PR #911, opened 2024-06-25
Fix failing kvdb tests

JesseAbram:
Turns out there were clean_test() in functions that did not spin up a kvdb causing sometimes (if the kvdb was not cleared in the previous test) a panic and failing test issue. The clean_test() function now checks before trying to delete the file and has been removed when not needed

### DIFF
--- a/crates/kvdb/src/lib.rs
+++ b/crates/kvdb/src/lib.rs
@@ -34,7 +34,7 @@ pub fn get_db_path(testing: bool) -> String {
 }
 
 pub fn clean_tests() {
-    if  fs::metadata(get_db_path(true)).is_ok() {
+    if fs::metadata(get_db_path(true)).is_ok() {
         let _result = std::fs::remove_dir_all(get_db_path(true));
     }
 }
--- a/crates/kvdb/src/lib.rs
+++ b/crates/kvdb/src/lib.rs
@@ -35,7 +35,7 @@ pub fn get_db_path(testing: bool) -> String {
 
 pub fn clean_tests() {
     let db_path = get_db_path(true);
-    if fs::metadata(db_path).is_ok() {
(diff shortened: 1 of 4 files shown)


crates/kvdb/src/lib.rs
```diff
@@ -34,5 +34,7 @@ pub fn get_db_path(testing: bool) -> String {
 }
 
 pub fn clean_tests() {
-    let _result = std::fs::remove_dir_all(get_db_path(true));
+    if fs::metadata(get_db_path(true)).is_ok() {
+        let _result = std::fs::remove_dir_all(get_db_path(true));
+    }
```
HCastano, line 39:
```suggestion
    let db_path = get_db_path(true);
    if fs::metadata(db_path).is_ok() {
        let _result = std::fs::remove_dir_all(db_path);
    }
```

crates/threshold-signature-server/src/health/tests.rs
```diff
@@ -14,17 +14,20 @@
...
 use axum::http::StatusCode;
+use entropy_kvdb::clean_tests;
 use serial_test::serial;
 
 use crate::helpers::tests::{initialize_test_logger, setup_client};
 
 #[tokio::test]
 #[serial]
 async fn health() {
+    clean_tests();
```
HCastano, line 25: Have these tests not been running file without this? iirc it's only the hashes test that's been failing, but I could be wrong

HCastano (approved):

crates/threshold-signature-server/src/health/tests.rs
JesseAbram: No they do run but without a clean db state, which is fine persay but should start from a clean state

PR #912, opened 2024-06-26
Fix pallet treasury benchmarks

JesseAbram:
These were broken and ignored for the last release. This PR fixes that.


runtime/src/lib.rs
```diff
@@ -1009,7 +1010,8 @@ impl pallet_membership::Config<pallet_membership::Instance1> for Runtime {
 
 parameter_types! {
   pub const ProposalBond: Permill = Permill::from_percent(5);
-  pub const ProposalBondMinimum: Balance = DOLLARS;
+  pub const ProposalBondMinimum: Balance = 100 * DOLLARS;
+pub const ProposalBondMaximum: Balance = 500 * DOLLARS;
```
HCastano, line 1014:
```suggestion
  pub const ProposalBondMaximum: Balance = 500 * DOLLARS;
```

runtime/src/lib.rs
```diff
@@ -1041,30 +1044,87 @@ impl pallet_treasury::Config for Runtime {
...
+    }
+
+    #[cfg(feature = "runtime-benchmarks")]
+    fn pay(
+        who: &Self::Beneficiary,
+        _asset_kind: Self::AssetKind,
+        amount: Self::Balance,
+    ) -> Result<Self::Id, Self::Error> {
+        // In case of benchmarks, we adjust the value by multiplying it by 1_000_000_000_000, otherwise it fails with BelowMinimum limit error, because
+        // treasury benchmarks uses only 100 as the amount.
```
HCastano, line 1102:
```suggestion
        // In case of benchmarks, we adjust the value by multiplying it by 1_000_000_000_000,
        // otherwise it fails with BelowMinimum limit error, because treasury benchmarks uses only
        // 100 as the amount.
```

runtime/src/lib.rs
```diff
@@ -1041,30 +1044,87 @@ impl pallet_treasury::Config for Runtime {
...
+        let _ = <Balances as fungible::Mutate<_>>::transfer(
+            &TreasuryAccount::get(),
+            who,
+            amount * 1_000_000_000_000,
```
HCastano, line 1106: So I'm not gonna dig into the logic here, just gonna assume this is fine. But I'd at least change this to something like:

```suggestion
            amount * 100 * DOLLARS
```

runtime/src/lib.rs
```diff
@@ -1041,30 +1044,87 @@ impl pallet_treasury::Config for Runtime {
...
+
+    fn check_payment(_id: Self::Id) -> PaymentStatus {
+        PaymentStatus::Success
+    }
+
+    #[cfg(feature = "runtime-benchmarks")]
+    fn ensure_successful(_: &Self::Beneficiary, _: Self::AssetKind, amount: Self::Balance) {
+        <Balances as fungible::Mutate<_>>::mint_into(
+            &TreasuryAccount::get(),
+            amount * 1_000_000_000_000,
```
HCastano, line 1120: Same here

HCastano (approved):
Thanks for fixing this

runtime/src/lib.rs
HCastano, line 1120:
```suggestion
           amount * 100 * DOLLARS,
```

PR #916, opened 2024-06-26
Run benchmarks as part of CI instead of just checking build

HCastano:
It would be nice to see if the benchmarks actually run all the way through as opposed to
just checking if they simply build.

I'm curious to see how long this'll take to do. If it's too long to do on every PR then I
may change this to just run on `master` builds.


HCastano:
Closing as stale since I haven't had the time to follow up with this.

PR #918, opened 2024-06-28
Jumpstart network

JesseAbram:
Jumpstarts the network with a global key

* Uses same pathways but hardsets global key to 0x000 address


crates/threshold-signature-server/src/user/api.rs
```diff
@@ -325,21 +330,31 @@ async fn setup_dkg(
...
         let sig_request_address = SubxtAccountId32(*address_slice);
-        let user_details =
-            get_registering_user_details(&api, &sig_request_address.clone(), rpc).await?;
+        let mut key_visibility = KeyVisibility::Public;
+        // If master key set visibility to public
+        if sig_request_account != master_key.encode() {
+            let user_details =
+                get_registering_user_details(&api, &sig_request_address.clone(), rpc).await?;
+            key_visibility = user_details.key_visibility.0;
+        };
```
HCastano, line 339:
```suggestion
        let key_visibility = if sig_request_account == master_key.encode() {
            KeyVisibility::Public
        } else {
            let user_details =
                get_registering_user_details(&api, &sig_request_address.clone(), rpc).await?;
            user_details.key_visibility.0
        };

```

This makes the assignment a bit more clear, and we can also get rid of that comment

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -325,21 +330,31 @@ async fn setup_dkg(
...
             data.block_number,
         )
         .await?;
         let verifying_key = key_share.verifying_key().to_encoded_point(true).as_bytes().to_vec();
-        let string_verifying_key = hex::encode(verifying_key.clone()).to_string();
+        let mut string_verifying_key = hex::encode(verifying_key.clone()).to_string();
+        // If jump start store key under the master key (zero address)
+        if sig_request_account == master_key.encode() {
+            string_verifying_key = hex::encode(master_key).to_string();
+        }
```
HCastano, line 356: Wouldn't this lead to a mismatch between the `verifying_key` (which is used in the `confirm_registered` call and what gets stored in the database + used in the `send_key` call?

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -325,21 +330,31 @@ async fn setup_dkg(
...
             &app_state.listener_state,
             sig_request_address.clone(),
-            *user_details.key_visibility,
+            key_visibility,
             data.block_number,
         )
         .await?;
         let verifying_key = key_share.verifying_key().to_encoded_point(true).as_bytes().to_vec();
-        let string_verifying_key = hex::encode(verifying_key.clone()).to_string();
+        let mut string_verifying_key = hex::encode(verifying_key.clone()).to_string();
```
HCastano, line 352: We can also simplify this logic similar to the above (expressions instead of using `mut`)

HCastano (commented):
There's a couple more things I need to look at, in particular the use of the zero-address around the different calls

crates/threshold-signature-server/src/user/api.rs
JesseAbram: so we don't call confirm_register when jump starting we call jump_start_results

pallets/registry/src/lib.rs
```diff
@@ -128,6 +139,18 @@ pub mod pallet {
         pub registered_accounts: Vec<(T::AccountId, u8, Option<[u8; 32]>, VerifyingKey)>,
     }
 
+    #[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+    pub enum JumpStartStatus {
```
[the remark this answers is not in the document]
JesseAbram: mmmm idk depends on the implementation, it can very well be that register takes no communication between threshold servers and when they want the veryfying key it is just a call to a tss server, or maybe it does, in this regard that should be applied when new register flow is implemented

pallets/registry/src/lib.rs
```diff
@@ -128,6 +139,18 @@ pub mod pallet {
...
+        Ready,
+        InProgress(u32),
```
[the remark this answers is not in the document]
JesseAbram: will do, to answer it tho, it is the block number when jump start was called (in case of failure and retry)

pallets/registry/src/lib.rs
```diff
@@ -225,18 +259,113 @@ pub mod pallet {
...
+    /// Allows anyone to create a master key for the network if the network is read and a master key does not exist
     #[pallet::call]
     impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight({
+            <T as Config>::WeightInfo::jump_start_network()
+        })]
+        pub fn jump_start_network(origin: OriginFor<T>) -> DispatchResult {
+            let _who = ensure_signed(origin)?;
+            let network_account = H256::zero();
```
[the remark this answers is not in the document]
JesseAbram: should be easily fixed with a guard on register, good find

pallets/registry/src/lib.rs
```diff
@@ -225,18 +259,113 @@ pub mod pallet {
...
+            Ok(())
+        }
+
+        /// Allows validators to signal a succecful network jumpstart
+        #[pallet::call_index(1)]
+        #[pallet::weight({
+                <T as Config>::WeightInfo::jump_start_results_confirm(SIGNING_PARTY_SIZE as u32)
+                .max(<T as Config>::WeightInfo::jump_start_results_done(SIGNING_PARTY_SIZE as u32))
+        })]
+        pub fn jump_start_results(
```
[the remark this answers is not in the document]
JesseAbram: same concept a % or all of the signing nodes will confirm, the subgroup is irrelevant

pallets/registry/src/lib.rs
```diff
@@ -225,18 +259,113 @@ pub mod pallet {
...
+                _ => return Err(Error::<T>::JumpStartNotInProgress.into()),
+            };
+
+            ensure!(
+                !jump_start_info.confirmations.contains(&signing_subgroup),
+                Error::<T>::AlreadyConfirmed
+            );
+            let confirmation_length = jump_start_info.confirmations.len() as u32;
+
+            // TODO  some sort of test I guess like a sign from this account or check the verifying keys
```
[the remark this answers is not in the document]
JesseAbram: same as we do in confirm registered make sure all the verifying key from the nodes are the same

pallets/registry/src/lib.rs
```diff
@@ -225,18 +259,113 @@ pub mod pallet {
...
+            // If failed unlock the locks and allow another jumpstart
```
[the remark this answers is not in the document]
JesseAbram: this is related to the comment and the issue to be created above, how we define a success and a failure, will be pretty obvious when decided

pallets/registry/src/tests.rs
```diff
@@ -141,6 +142,92 @@ fn it_registers_a_user() {
...
+        System::set_block_number(100);
+
+        assert_ok!(Registry::jump_start_network(RuntimeOrigin::signed(1)));
+        assert_eq!(
+            Registry::jump_start_progress(),
+            JumpStartDetails {
+                jump_start_status: JumpStartStatus::InProgress(100),
+                confirmations: vec![]
+            },
+            "ensures jump start is called again if too many blocks passed"
```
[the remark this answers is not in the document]
JesseAbram: yes mostly, if we define like requires a valid sig from each tss or the same verifying key there are some ways we can know it failed.......but the more likely ways is a message dropped and no on responds and ya, like 50 blocks later we should be safe to retry

pallets/registry/src/lib.rs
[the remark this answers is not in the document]
JesseAbram: https://github.com/entropyxyz/entropy-core/issues/927

ameba23:
Ive had a quick look at this and it looks great.  Since it is a step towards having a network parent key but not fully implemented, Im trying to figure out what exactly the scope of this PR is. 

This is what i am getting, please correct if wrong:

- This adds a 'parent' DKG for the network - but does not yet remove per-user DKG.
- The registry pallet tracks whether a network parent key exists (`Done`), does not exist (`Ready`) or DKG is in progress (`InProgress`), but there are not yet restrictions on users signing or registering when the network parent key does not exist.
- The network DKG is activated by submitting a `jump_start_network` extrinsic - but there are currently no restrictions as to who can submit this or how big the network should be (https://github.com/entropyxyz/entropy-core/issues/923).
- The network can only be jump started if it is either not yet jump started, or jump starting has been in-progress for at least 50 blocks.
- The network is considered successfully jump started on getting register confirmations from all but one of the TSS nodes who participated in the network DKG (i am not totally clear on this, and i feel like it should be 100%).
- Signature requests for the parent key account will result in an error.

@JesseAbram @HCastano

JesseAbram:
> Ive had a quick look at this and it looks great. Since it is a step towards having a network parent key but not fully implemented, Im trying to figure out what exactly the scope of this PR is.
> 
> This is what i am getting, please correct if wrong:
> 
> * This adds a 'parent' DKG for the network - but does not yet remove per-user DKG.

yes

> * The registry pallet tracks whether a network parent key exists (`Done`), does not exist (`Ready`) or DKG is in progress (`InProgress`), but there are not yet restrictions on users signing or registering when the network parent key does not exist.

also yes because we don't use it yet.....and as we move the registery flow to the new flow this, is where having a block before jumpstart should be addressed

> * The network DKG is activated by submitting a `jump_start_network` extrinsic - but there are currently no restrictions as to who can submit this or how big the network should be ([Jump Start network requirements #923](https://github.com/entropyxyz/entropy-core/issues/923)).
> * The network can only be jump started if it is either not yet jump started, or jump starting has been in-progress for at least 50 blocks.
> * The network is considered successfully jump started on getting register confirmations from all but one of the TSS nodes who participated in the network DKG (i am not totally clear on this, and i feel like it should be 100%).

No it is all of them...... If you are referring to this line https://github.com/entropyxyz/entropy-core/pull/918/files#diff-1e0acd338c23affee5840b8efb06bd46be87c49819841bdcf3ce804631a40b92R340 it is because the last confirmation sets it to done as if there were 4 out of 5 done, this one is the 5th

> * Signature requests for the parent key account will result in an error.
> 
> @JesseAbram @HCastano

crates/threshold-signature-server/src/user/api.rs
HCastano: In `setup_dkg` we call `confirm_register` at the end of the function

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -500,17 +517,25 @@ pub async fn confirm_registered(
...
+        let tx_request = entropy::tx().registry().confirm_register(
+            who,
+            subgroup,
+            entropy::runtime_types::bounded_collections::bounded_vec::BoundedVec(verifying_key),
+        );
+        submit_transaction(api, rpc, signer, &tx_request, Some(nonce)).await?;
+    } else {
+        let tx_request = entropy::tx().registry().jump_start_results(subgroup);
+        submit_transaction(api, rpc, signer, &tx_request, Some(nonce)).await?;
+    }
```
HCastano, line 538:
```suggestion
    if who.encode() == NETWORK_PARENT_KEY.encode() {
        let jump_start_request = entropy::tx().registry().jump_start_results(subgroup);
        submit_transaction(api, rpc, signer, &jump_start_request, Some(nonce)).await?;
    } else {
        let confirm_register_request = entropy::tx().registry().confirm_register(
            who,
            subgroup,
            entropy::runtime_types::bounded_collections::bounded_vec::BoundedVec(verifying_key),
        );
        submit_transaction(api, rpc, signer, &confirm_register_request, Some(nonce)).await?;
    }

```

HCastano: And this way we can also get rid of the `parent_key` argument into `confirm_registered`

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -325,21 +330,31 @@ async fn setup_dkg(
...
         )
         .await?;
         let verifying_key = key_share.verifying_key().to_encoded_point(true).as_bytes().to_vec();
-        let string_verifying_key = hex::encode(verifying_key.clone()).to_string();
+        let mut string_verifying_key = hex::encode(verifying_key.clone()).to_string();
+        // If jump start store key under the master key (zero address)
+        if sig_request_account == NETWORK_PARENT_KEY.encode() {
+            string_verifying_key = hex::encode(*NETWORK_PARENT_KEY).to_string();
+        }
+
```
HCastano, line 357:
```suggestion
        let string_verifying_key = if sig_request_account == NETWORK_PARENT_KEY.encode() {
            hex::encode(*NETWORK_PARENT_KEY)
        } else {
            hex::encode(verifying_key.clone())
        }
        .to_string();

```

HCastano: I still don't understand this substitution though.

Why do we need to change the key in the DB under which we store the parent key's keyshare? Like why not just use the actual verifying key instead of the `NETWORK_PARENT_KEY`?

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -500,17 +517,25 @@ pub async fn confirm_registered(
     signer: &PairSigner<EntropyConfig, sr25519::Pair>,
     verifying_key: Vec<u8>,
     nonce: u32,
+    parent_key: Vec<u8>,
```
HCastano, line 520:
```suggestion
```

crates/threshold-signature-server/src/user/errors.rs
```diff
@@ -151,6 +151,8 @@ pub enum UserErr {
     EncryptionOrAuthentication(#[from] EncryptedSignedMessageErr),
     #[error("Custom hash choice out of bounds")]
     CustomHashOutOfBounds,
+    #[error("No signing from master key")]
+    NoSigningFromMasterKey,
```
HCastano, line 155: Should be `parent`

pallets/registry/src/lib.rs
```diff
@@ -225,18 +259,115 @@ pub mod pallet {
...
+            Ok(())
+        }
+
+        /// Allows validators to signal a successful network jumpstart
+        #[pallet::call_index(1)]
+        #[pallet::weight({
+                <T as Config>::WeightInfo::jump_start_results_confirm(SIGNING_PARTY_SIZE as u32)
+                .max(<T as Config>::WeightInfo::jump_start_results_done(SIGNING_PARTY_SIZE as u32))
+        })]
+        pub fn jump_start_results(
```
HCastano, line 312: This should probably be renamed to something like: 

```suggestion
        pub fn confirm_jump_start(
```

pallets/registry/src/lib.rs
```diff
@@ -225,18 +259,115 @@ pub mod pallet {
...
+            );
+
+            ensure!(
+                !jump_start_info.confirmations.contains(&signing_subgroup),
+                Error::<T>::AlreadyConfirmed
+            );
+            let confirmation_length = jump_start_info.confirmations.len() as u32;
+
+            // TODO: some sort of test I a sign from this account or check the verifying keys see :https://github.com/entropyxyz/entropy-core/issues/927
+            // If failed unlock the locks and allow another jumpstart
```
HCastano, line 339:
```suggestion
            // TODO (#927): Add another check, such as a signature or a verifying key comparison, to
            // ensure that registration was indeed successful.
            //
            // If it fails we'll need to allow another jumpstart.
```

pallets/registry/src/lib.rs
```diff
@@ -225,18 +259,115 @@ pub mod pallet {
...
+                JumpStartStatus::InProgress(started_block_number) => {
+                    if converted_block_number.saturating_sub(started_block_number)
+                        < BLOCKS_TO_RESTART_JUMP_START
+                    {
+                        return Err(Error::<T>::JumpStartProgressNotReady.into());
+                    };
+                },
+                _ => return Err(Error::<T>::JumpStartProgressNotReady.into()),
+            };
+            // TODO: Add checks for network state see https://github.com/entropyxyz/entropy-core/issues/923
```
HCastano, line 293:
```suggestion
            // TODO (#923): Add checks for network state.
```

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -774,6 +786,82 @@ async fn test_store_share() {
...
+
+    assert_eq!(user_registration_response.text().await.unwrap(), "");
+    // wait for jump start event check that key exists in kvdb
+    for _ in 0..45 {
+        std::thread::sleep(std::time::Duration::from_millis(1000));
+        let block_hash = rpc.chain_get_block_hash(None).await.unwrap();
+        let events = EventsClient::new(api.clone()).at(block_hash.unwrap()).await.unwrap();
+        let jump_start_event = events.find::<entropy::registry::events::FinishedNetworkJumpStart>();
+        for event in jump_start_event.flatten() {
+            dbg!(event);
```
HCastano, line 844:
```suggestion
```

pallets/registry/src/lib.rs
```diff
@@ -247,7 +378,11 @@ pub mod pallet {
             programs_data: BoundedVec<ProgramInstance<T>, T::MaxProgramHashes>,
         ) -> DispatchResultWithPostInfo {
             let sig_req_account = ensure_signed(origin)?;
-
+            let encoded_sig_req_account = sig_req_account.encode();
+            ensure!(
+                encoded_sig_req_account != NETWORK_PARENT_KEY.encode(),
+                Error::<T>::NoRegisteringFromParentKey
+            );
```
HCastano, line 385: I think this could be useful to have. E.g we can have the Entropy network control an account on other chains like this.

Obvs this wouldn't work if the parent key address is the zero-address though.

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -142,6 +143,11 @@ pub async fn sign_tx(
         .number;
 
     check_stale(user_sig_req.block_number, block_number).await?;
+    // Probably impossible but block signing from master key anyways
+    if user_sig_req.signature_verifying_key == H256::zero().0.to_vec() {
```
HCastano, line 147:
```suggestion
    if user_sig_req.signature_verifying_key == NETWORK_PARENT_KEY {
```

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -492,6 +508,7 @@ pub async fn is_registering(
 }
 
 /// Confirms that a address has finished registering on chain.
+#[allow(clippy::too_many_arguments)]
```
HCastano, line 511:
```suggestion
```

HCastano (commented):
Getting close!

Couple things:
- It's unclear to me why we need to substitute the parent verifying key in the KVDB with `0x00...`
- I'm still slightly worried about the use of the zero address here, but we could kick this down the road to at least get things moving (maybe we open an issue for right now)
- We need to rename comments, errors, etc. to `parent`

JesseAbram:
> Getting close!
> 
> Couple things:
> 
> * It's unclear to me why we need to substitute the parent verifying key in the KVDB with `0x00...`

In theory we don't and we can change it, but when we keep it a set amount, we can and it is easier to make it a global constant and have if parent key do x code easier (which we can also add extra security on it too) 
> * I'm still slightly worried about the use of the zero address here, but we could kick this down the road to at least get things moving (maybe we open an issue for right now)

Global constant, I know your impulse, I think it is fine we can make it any random key tbh
> * We need to rename comments, errors, etc. to `parent`

crates/threshold-signature-server/src/user/errors.rs
```diff
@@ -151,6 +151,8 @@ pub enum UserErr {
     EncryptionOrAuthentication(#[from] EncryptedSignedMessageErr),
     #[error("Custom hash choice out of bounds")]
     CustomHashOutOfBounds,
+    #[error("No signing from parent key")]
+    NoSigningFromMasterKey,
```
HCastano, line 155:
```suggestion
    NoSigningFromParentKey,
```

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -325,21 +330,32 @@ async fn setup_dkg(
...
             sig_request_address.clone(),
-            *user_details.key_visibility,
+            key_visibility,
             data.block_number,
         )
         .await?;
         let verifying_key = key_share.verifying_key().to_encoded_point(true).as_bytes().to_vec();
-        let string_verifying_key = hex::encode(verifying_key.clone()).to_string();
+        let string_verifying_key = if sig_request_account == NETWORK_PARENT_KEY.encode() {
+            hex::encode(*NETWORK_PARENT_KEY)
```
HCastano, line 353: One suggestion: instead of overwriting the parent's verifying key to be `0x00` we could
have a separate "forbidden key" entry, `PARENT_VERIFYING_KEY`, which points to the parent
verifying key.

For those checks you mention, we could achieve them the same way, but doing an extra
lookup.

That way we keep the verifying key entries consistent.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ At the moment this project **does not** adhere to
 - Add `blake2` as built in hash function and make `HashingAlgorithm` non-exhaustive ([#881](https://github.com/entropyxyz/entropy-core/pull/881))
 - Add sort to subgroup signer selection ([#900](https://github.com/entropyxyz/entropy-core/pull/900))
 - Create four node Docker Compose chainspec ([#902](https://github.com/entropyxyz/entropy-core/pull/902))
+- Jumpstart network ([#918](https://github.com/entropyxyz/entropy-core/pull/918))
```
HCastano, line 38: Oh actually, this should go under `Unreleased`

HCastano (approved):
Few things we need to follow up on, but looks good otherwise!

crates/threshold-signature-server/src/user/api.rs
JesseAbram: sure at this point tho im in favour of opening an issue

HCastano: Yeah an issue is fine

JesseAbram: https://github.com/entropyxyz/entropy-core/issues/931

PR #922, opened 2024-07-03
Oracle data integration

### LINKED ISSUE #789: Integrate oracle data into programs
#788 starts this process a bit early to make sure we have backwards compatibility with future updates

- [x] Create a new runtime that allows evaluate to take oracle data 
- [x] add oracle data pointer to program data 
- [x] Create an oracle pallet to hold incoming data 
- [x] Allow oracle data pointer to point to said data 
- [x] Grab data from pointer and pass it to program execution 

After doing this we can modularly add new "oracle data" by having pallets update the oracle mapping with new and proper keys

JesseAbram:
This PR integrates oracle data into programs and implements block number as the first use case

* All oracle data is determined by the chain (how is up to specific implementation) 
* Oracle data is stored in oracle_data in the Oracle pallet under a scale encoded key and scale encoded value (bounded vecs for both
* upon running a sig request the oracle data location is pulled from the ```ProgramInfo``` in the ```Programs``` pallet to know what heading the specific oracle data is under
* The data pulled from the chain is passed to the program for evaluation 

Closes #789


pallets/oracle/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+        type MaxOracleKeyLength: Get<u32>;
+        /// The maximum amount of owned programs.
+        type MaxOracleValueLength: Get<u32>;
+        /// Weight information for the extrinsics in this module.
+        type WeightInfo: WeightInfo;
+    }
+
+    #[pallet::storage]
+    #[pallet::getter(fn oracle_data)]
+    // TODO: parameterize bounded vec constants
```
ameba23, line 55: Maybe make an issue

ameba23 (approved):
I've read over this and it all looks good. 

But on a high level im still a bit unclear on the use case for oracles and how users who want to add an oracle data source would do that.

Are oracles supposed to be something which is fetched or computed at the point of making a signature request, or would the chain constantly retrieve data regardless of whether signature requests are made?

JesseAbram:
> I've read over this and it all looks good.
> 
> But on a high level im still a bit unclear on the use case for oracles and how users who want to add an oracle data source would do that.
> 
> Are oracles supposed to be something which is fetched or computed at the point of making a signature request, or would the chain constantly retrieve data regardless of whether signature requests are made?

the second one, then it gets fed into the program based on how the program is written. It can be a lot of data, but the "end game" for this as of now would be super cool to get ibc proofs on chain and then be able to give a proof of the state of a chain verify it in a program and use it as a constraint

PR #930, opened 2024-07-11
Release `v0.2.0`

HCastano:
Follow up to #910.

Only changes since then are some dependency bumps from Dependabot and #929.

:shipit:


frankiebee (approved):
Thank you!

PR #933, opened 2024-07-12
Change parent key heading

### LINKED ISSUE #931: Parent Key Value
In #918 the network creates a parent key and currently uses the 0x00 address

However in ed25519 and sr25519 that is an insecure account, actions on the address are blocked anyways, however there could be potential secondary effects we are not aware of (probably not but still why tempt fate) 

The shard can sit under any key value in the kvdb and having it known actually helps, changing it to "NETWORK_PARENT_KEY" seems to make sense

JesseAbram:
Closes #931


crates/shared/src/constants.rs
```diff
@@ -55,3 +53,6 @@ pub const VERIFICATION_KEY_LENGTH: u32 = 33;
 
 /// `device_key_proxy.wasm` from the `programs` repo.
 pub const DEVICE_KEY_PROXY: &[u8] = include_bytes!("../device_key_proxy.wasm");
+
+/// Network parent key specific size to fit into [u8; 32] to save extra code
```
HCastano, line 57: Can you elaborate on this comment? Where are we saving code?

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -49,8 +50,12 @@ pub const LATEST_BLOCK_NUMBER_PROACTIVE_REFRESH: &str = "LATEST_BLOCK_NUMBER_PRO
 #[cfg(test)]
 pub const DEFAULT_ENDPOINT: &str = "ws://localhost:9944";
 
-pub const FORBIDDEN_KEYS: [&str; 3] =
-    [FORBIDDEN_KEY_MNEMONIC, FORBIDDEN_KEY_SHARED_SECRET, FORBIDDEN_KEY_DIFFIE_HELLMAN_PUBLIC];
+pub const FORBIDDEN_KEYS: [&str; 4] = [
+    FORBIDDEN_KEY_MNEMONIC,
+    FORBIDDEN_KEY_SHARED_SECRET,
+    FORBIDDEN_KEY_DIFFIE_HELLMAN_PUBLIC,
+    NETWORK_PARENT_KEY,
```
HCastano, line 57: So I think we should have something like this added below:

```rust
pub const FORBIDDEN_NETWORK_PARENT_KEY: &str = entropy_shared::NETWORK_PARENT_KEY;
```

and then we add it to the array.

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -817,7 +817,7 @@ async fn test_jumpstart_network() {
         },
     ];
     let onchain_user_request = OcwMessageDkg {
-        sig_request_accounts: vec![H256::zero().encode()],
+        sig_request_accounts: vec![NETWORK_PARENT_KEY.encode()],
```
HCastano, line 820: Just curious, do you know what `AccountId` this ends up corresponding to?

HCastano (approved):

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -49,8 +50,12 @@ pub const LATEST_BLOCK_NUMBER_PROACTIVE_REFRESH: &str = "LATEST_BLOCK_NUMBER_PRO
...
+];
 
 pub const FORBIDDEN_KEY_MNEMONIC: &str = "MNEMONIC";
```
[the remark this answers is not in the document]
JesseAbram: I don't think it should be since it needs to be pulled into the chain too so no_std and I don't wanna pull this whole crate into core

crates/shared/src/constants.rs
JesseAbram: specifically here https://github.com/entropyxyz/entropy-core/blob/master/crates/threshold-signature-server/src/user/api.rs#L328 where we assume all incoming keys are [u8; 32] probably also lower in the stack too

crates/threshold-signature-server/src/helpers/launch.rs
JesseAbram: I mean I can but is that really more descriptive?

crates/threshold-signature-server/src/user/tests.rs
JesseAbram: 4e4554574f524b5f504152454e545f4b45595f464f525f454e54524f50595f

crates/threshold-signature-server/src/helpers/launch.rs
HCastano: It's more so about being consistent with naming of "Forbidden" keys, while still allowing you to refer to the parent key as an account in different contexts

PR #934, opened 2024-07-12
Make TSS startup checks more forgiving

HCastano:
Right now the TSS panics and crashes if it's not able to establish a connection with a
Substrate node or if the TSS account doesn't meet a minimum balance.

This PR makes it so that the TSS server retries establishing a connection for ~15 mins
before calling it quits. It also doesn't crash if the TSS account doesn't have a minimum
balance, instead it prints a warning and continues.

### DIFF
--- a/crates/threshold-signature-server/src/helpers/launch.rs
+++ b/crates/threshold-signature-server/src/helpers/launch.rs
@@ -241,21 +241,23 @@ pub struct StartupArgs {
     pub mnemonic_file: Option<PathBuf>,
 }
 
-pub async fn has_mnemonic(kv: &KvManager) -> (bool, String) {
+pub async fn threshold_account_id(kv: &KvManager) -> String {
+    let mnemonic = kv.kv().get(FORBIDDEN_KEY_MNEMONIC).await.expect("Issue getting mnemonic");
+    let pair = <sr25519::Pair as Pair>::from_phrase(
+        &String::from_utf8(mnemonic).expect("Issue converting mnemonic to string"),
+        None,
+    )
+    .expect("Issue converting mnemonic to pair");
+    AccountId32::new(pair.0.public().into()).to_ss58check()
+}
+
+pub async fn has_mnemonic(kv: &KvManager) -> bool {
     let exists = kv.kv().exists(FORBIDDEN_KEY_MNEMONIC).await.expect("issue querying DB");
-    let mut account_id = "".to_string();
     if exists {
         tracing::debug!("Existing mnemonic found in keystore.");
-        let mnemonic = kv.kv().get(FORBIDDEN_KEYS[0]).await.expect("Issue getting mnemonic");
-        let pair = <sr25519::Pair as Pair>::from_phrase(
-            &String::from_utf8(mnemonic).expect("Issue converting mnemonic to string"),
-            None,
-        )
-        .expect("Issue converting mnemonic to pair");
-        account_id = AccountId32::new(pair.0.public().into()).to_ss58check();
     }
 
-    (exists, account_id)
+    exists
 }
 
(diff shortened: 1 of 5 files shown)


crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -370,3 +372,63 @@ pub async fn setup_only(kv: &KvManager) {
...
+
+            match balance_query {
+                Ok(has_minimum_balance) => {
+                    if has_minimum_balance {
+                        tracing::info!(
+                            "The account `{}` has enough funds for submitting extrinsics.",
+                            &account_id
+                        )
+                    } else {
+                        tracing::warn!(
```
JesseAbram, line 417: I really think we should crash the node here, it is to me the same theory as a compile error vs a runtime error, what this does is allow validators to set up a node that will fail eventually

HCastano: Disagree. An operator could see the warning on startup and do something about it later on. I could be swayed here though.

Maybe we defer the decision to @vitropy? td;dr for you Vi, do we allow a TSS to start up without a funded TSS account?

HCastano: Thanks for the feedback @vitropy! What would happen if the TSS started without any funds is that if it tried to participate in operations (e.g registration or signing) it would error out with something like: "Insufficient funds for operation".

However, as soon as it was funded it would operate normally.

The point about not having to restart Docker containers is a good one, and to me that makes a clear case for not crashing the process straight away.

@JesseAbram do you have a strong opinion against this? If not I say we merge this as is and we can revisit later if needed

JesseAbram: I do, as we personally have made this exact mistake before and it was pretty hard to diagnose. Instead of letting an operator run a node that would fail during a critical process (registering) we should force them to restart the docker container

JesseAbram: you can't read logs of other operators in a decentralized system, you are assuming that after a validator who is not us spins up a validator, they read and understand the warning and or monitor their logs and understand the error message

JesseAbram: if they can't figure out how to fund an account they should not be a validator hard stop

JesseAbram: > > if they can't figure out how to fund an account they should not be a validator hard stop
> 
> Would a TSS account that had no funds interfere with or cause issues for other validators on the network or otherwise be able to function as a validator in the first place? If not, what does it matter to you if their container is running or not?

yes in the case where they are in the signing group they would not be able to signal that they have completed the dkg to the chain and cause the whole registering process to fail (affects user and other validators) 
> 
> Also,
> 
> > we personally have made this exact mistake before and it was pretty hard to diagnose.
> 
> Why was it hard to diagnose?

This was before we set up loki so I had to read all the individual logs.......however in the case of it not being our validator and some random other persons it would be extremely challenging without having their logs

HCastano: My 2 cents to close this out:

With the logging we have it should not be hard for an operator to find out that their
node does not have sufficient funds at process startup and address this without needing
to muck around with their infrastructure again.

Having Docker containers crash to me seems like the wrong way to tell them of this, and I
don't want to have to couple process startup with having a funded account.

The implications of running a node with a non-funded account are not problematic on a full
network. It's been problematic for us during genesis specifically because there's been no
redundancy, i.e we've needed every machine we've spun up to work. On a live network this
isn't the case, some machines can be down and the network can continue to operate.

If we find that operations are unable to figure this out on their own or if the
implications of machines running without funds increases we can revisit this.

JesseAbram (approved):

PR #938, opened 2024-07-16
Add Signer groups and rotation

JesseAbram:
- [x] Set parameters for size of signer group and T of N
- [x] Add Signer groups to chain 
- [x] Get randomness to select new signer from validators on session refresh
- [x] Rotate Signers


runtime/src/lib.rs
```diff
@@ -714,6 +714,7 @@ parameter_types! {
 impl pallet_staking_extension::Config for Runtime {
     type Currency = Balances;
     type MaxEndpointLength = MaxEndpointLength;
+    type Randomness = pallet_babe::RandomnessFromOneEpochAgo<Runtime>;
```
JesseAbram, line 717: https://github.com/entropyxyz/entropy-core/issues/944

pallets/staking/src/lib.rs
```diff
@@ -151,6 +160,14 @@ pub mod pallet {
     #[pallet::getter(fn proactive_refresh)]
     pub type ProactiveRefresh<T: Config> = StorageValue<_, RefreshInfo, ValueQuery>;
 
+    #[pallet::storage]
+    #[pallet::getter(fn signers)]
+    pub type Signers<T: Config> = StorageValue<_, Vec<T::ValidatorId>, ValueQuery>;
+
+    #[pallet::storage]
+    #[pallet::getter(fn next_signers)]
+    pub type NextSigners<T: Config> = StorageValue<_, Vec<T::ValidatorId>, ValueQuery>;
```
HCastano, line 169: Can you add docs to these?

pallets/staking/Cargo.toml
```diff
@@ -56,7 +58,8 @@ std=[
   'pallet-session/std',
   'pallet-staking/std',
   'scale-info/std',
+  'sp-consensus-babe/std',
   'sp-runtime/std',
-
+  "rand_chacha/std",
```
HCastano, line 63: Why use `chaha20` here instead of something else?

pallets/staking/src/lib.rs
```diff
@@ -382,9 +401,45 @@ pub mod pallet {
             Ok(ledger.stash)
         }
 
+        pub fn get_randomness() -> ChaCha20Rng {
+            let phrase = b"signer_rotation";
+            // TODO: Is randomness freshness an issue here
+            // https://github.com/paritytech/substrate/issues/8312
+            let (seed, _) = T::Randomness::random(phrase);
```
HCastano, line 408: So does this mean that we should also be checking how fresh the randomness is here before committing to use it ?

pallets/staking/src/lib.rs
```diff
@@ -382,9 +401,45 @@ pub mod pallet {
...
+        }
+
         pub fn new_session_handler(
-            _validators: &[<T as pallet_session::Config>::ValidatorId],
+            validators: &[<T as pallet_session::Config>::ValidatorId],
         ) -> Result<(), DispatchError> {
+            let mut current_signers = Self::signers();
+            // Since not enough validators do not allow rotation
+            // TODO: https://github.com/entropyxyz/entropy-core/issues/943
+            if validators.len() <= current_signers.len() {
```
HCastano, line 421: In the case that `validators.len() == current_signers.len()` why not allow rotation?

pallets/staking/src/lib.rs
```diff
@@ -382,9 +401,45 @@ pub mod pallet {
...
+                return Ok(());
+            }
+            let mut randomness = Self::get_randomness();
+            // grab a current signer to initiate value
+            let mut next_signer_up = &current_signers[0].clone();
+            let mut index;
+            // loops to find signer in validator that is not already signer
+            while current_signers.contains(next_signer_up) {
```
HCastano, line 429: This is a pretty expensive lookup to do on a `Vec`, esp with random indexes, can we get away with making `Signers` a map?

pallets/staking/src/lib.rs
```diff
@@ -382,9 +401,45 @@ pub mod pallet {
...
+                index = randomness.next_u32() % validators.len() as u32;
+                next_signer_up = &validators[index as usize];
+            }
+
+            // removes first signer and pushes new signer to back
+            current_signers.remove(0);
```
HCastano, line 435: Similar thing here, removing the first entry in a Vec will cause us to reallocate basically the whole thing. Maybe we can get away with a `VecDeque` if a `Map` doesn't work.

HCastano: Another thing, why remove the first signer instead of a random signer here?

pallets/staking/src/lib.rs
```diff
@@ -382,9 +401,45 @@ pub mod pallet {
...
+            current_signers.push(next_signer_up.clone());
+            NextSigners::<T>::put(current_signers);
+
+            // for next PR
+            // tell signers to do new key rotation with new signer group (dkg)
+            // confirm action has taken place
```
HCastano, line 441: Can you remove this or add a `TODO` with an issue?

pallets/parameters/src/lib.rs
```diff
@@ -133,5 +151,20 @@ pub mod module {
             });
             Ok(())
         }
+
+        #[pallet::call_index(2)]
+        #[pallet::weight(T::WeightInfo::change_signers_info())]
+        pub fn change_signers_info(
```
HCastano, line 157: Docs pls

pallets/parameters/src/lib.rs
```diff
@@ -80,19 +82,30 @@ pub mod module {
...
         }
     }
 
     #[pallet::error]
     pub enum Error<T> {}
 
+    #[derive(Clone, Encode, Decode, Eq, PartialEqNoBound, RuntimeDebug, TypeInfo, Default)]
+    pub struct SignersSize {
+        pub signers_size: u8,
+        pub threshold: u8,
```
HCastano, line 97:
```suggestion
        pub threshold: u8,
        pub total_signers: u8,
```

pallets/parameters/src/lib.rs
```diff
@@ -80,19 +82,30 @@ pub mod module {
...
+    }
+
     #[pallet::event]
     #[pallet::generate_deposit(fn deposit_event)]
     pub enum Event<T: Config> {
         /// Request limit changed
         RequestLimitChanged { request_limit: u32 },
         /// Max instructions per program changes
         MaxInstructionsPerProgramsChanged { max_instructions_per_programs: u64 },
+        /// Max instructions per program changes
```
HCastano, line 107: Comment needs updating

node/cli/src/chain_spec/testnet.rs
```diff
@@ -442,6 +446,9 @@ pub fn testnet_genesis_config(
         "parameters": ParametersConfig {
             request_limit: 20,
             max_instructions_per_programs: INITIAL_MAX_INSTRUCTIONS_PER_PROGRAM,
+            // TODO: figure out init values and const them
+            signers_size: 5,
+            threshold: 3,
```
HCastano, line 451: Should probably be `2-of-3` at the moment since that's how it's being done in the current integration tests

pallets/parameters/src/lib.rs
```diff
@@ -71,6 +71,8 @@ pub mod module {
     pub struct GenesisConfig<T: Config> {
         pub request_limit: u32,
         pub max_instructions_per_programs: u64,
+        pub signers_size: u8,
+        pub threshold: u8,
```
HCastano, line 75:
```suggestion
        pub threshold: u8,
        pub total_signers: u8,
```

HCastano (commented):
Don't think this needs an `impl_version` bump since we're already bumping the `spec_version`. I also don't think the `transaction_version` needs to be bumped since only new stuff is being added, as opposed to existing extrinsics being changed

pallets/staking/Cargo.toml
JesseAbram: saw it used in the society pallet, when I have a choice like this to make I default to auditted production code, open to other ideas

pallets/staking/src/lib.rs
JesseAbram: ummmmmmm no I don't believe so this is like more of a commit reveal thing where this can be useful, as in like idk a lottery, bets are placed, there is a clear end time then you grab randomness, but the way we use randomness is in an ever flowing system where validators are swapped in and out, potentially we can grab the randomness from a previous session, some more investitgation on how validators are ordered may help here, but good randomness is probably enough

JesseAbram: In any case this should be hashed out in #944

pallets/staking/src/lib.rs
JesseAbram: do have #943 open for discussion about this, but if we do not let signers leave validators while in the signer group (which is also a discussion lol) then we can assume all validators are signers here and we can not let them leave in fear of losing the parent key

pallets/staking/src/lib.rs
JesseAbram: mmmmm ok so we do know signers is relatively small and this would cause other operations (coming in future PRs) to increase complexity (happy to explain them but honestly may be better to see in a few days) that being said Im open to making this an issue and going back over to optimise this, but for now I feel strongly about leaving this as is to push through the flow

node/cli/src/chain_spec/testnet.rs
JesseAbram: good point

JesseAbram: I should probably also make them a global const

pallets/staking/src/lib.rs
JesseAbram: good question and saves a theoretical operation, a really cheap and low code way to make sure we only keep all signers in for max signer_size as in a new signer gets added to the back and the next signer removed gets taken from the front, I do want to test this more later to make sure they don't get sorted but my tests currently show they do not

pallets/staking/src/lib.rs
JesseAbram: oh this is all tracked in #941 I just like to write my next steps in comments in the code I mean im literally doing this next

ameba23 (approved):
Looks great :100: (pending CI)

Just to confirm the scope of this PR is to add a signing committee concept to the staking pallet, but not yet actually use them.

I haven't looked into the issue around 'freshness' relating to randomness, and obviously having a good source of randomness is pretty critical here, but since there is a comment and https://github.com/entropyxyz/entropy-core/issues/944 i think we should move ahead with this.

As discussed elsewhere, as well as randomness there are other factors we could use in this selection process eg: stake, age of validator, and number of blocks since a validator was last in the signing committee.

As for the frequency of rotation (moving one validator in/out per session), i think this is too frequent, and should probably also be configurable as a parameter, or even depend on this size of the validator set (the more validators, the more frequent the rotation).  As i mentioned before, my main concern is that the more frequent the rotation, the higher the chance that a malicious actor who controls `n` nodes will control the complete signing committee. But for simplicity i think we should leave this as it is for now and merge this.

As for having t and n configurable as a parameter.  I totally agree with this, and i can't see a reason why changing them with a running network would cause issues.  But of course issues with this could well pop up.

pallets/staking/src/tests.rs
```diff
@@ -323,3 +323,25 @@ fn it_declares_synced() {
...
+
+#[test]
+fn it_tests_new_session_handler() {
+    new_test_ext().execute_with(|| {
+        // start with current validators as 6 and 5
+
+        // no next signers at start
+        assert_eq!(Staking::next_signers().len(), 0);
+
+        Staking::new_session_handler(&vec![1u64, 2u64, 3u64]);
```
HCastano, line 335: You shouldn't need to explicitly specify the `u64` type in this whole test

pallets/staking/src/tests.rs
```diff
@@ -323,3 +323,25 @@ fn it_declares_synced() {
         assert!(Staking::is_validator_synced(5));
     });
 }
+
+#[test]
+fn it_tests_new_session_handler() {
+    new_test_ext().execute_with(|| {
+        // start with current validators as 6 and 5
```
HCastano, line 330:
```suggestion
        // Start with current validators as 5 and 6 based off the Mock `GenesisConfig`.
```

node/cli/src/chain_spec/testnet.rs
```diff
@@ -413,6 +414,10 @@ pub fn testnet_genesis_config(
                 })
                 .collect::<Vec<_>>(),
             proactive_refresh_data: (vec![], vec![]),
+            inital_signers: initial_authorities.iter().map(|auth| {
+                auth.0.clone()
+            })
+            .collect::<Vec<_>>(),
```
HCastano, line 420: For the testnet config we may want it to be empty actually, and instead have it be populated through a different jumpstart related mechanism.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ At the moment this project **does not** adhere to
 
 ### Added 
 - Jumpstart network ([#918](https://github.com/entropyxyz/entropy-core/pull/918))
+- Add Signer groups and rotation ([#938](https://github.com/entropyxyz/entropy-core/pull/938))
+
+### Breaking Changes
+- In [#938](https://github.com/entropyxyz/entropy-core/pull/938) As the network readys for signer subgroups `Signers` are added to the chain config, currently the creation of the config file should select and assign them
```
HCastano, line 17:
```suggestion
- In [#938](https://github.com/entropyxyz/entropy-core/pull/938), the chainspec got a couple of new
  fields, `pallet_staking_extension::initial_signers`, `pallet_parameters::total_signers`, and
  `pallet_parameters::threshold`, which are used to set up the initial threshold signing
  configuration for the network.
```

HCastano (approved):

pallets/parameters/src/lib.rs
```diff
@@ -133,5 +151,21 @@ pub mod module {
...
+        #[pallet::call_index(2)]
+        #[pallet::weight(T::WeightInfo::change_signers_info())]
+        pub fn change_signers_info(
+            origin: OriginFor<T>,
+            total_signers: u8,
+            threshold: u8,
+        ) -> DispatchResult {
+            T::UpdateOrigin::ensure_origin(origin)?;
+            let signer_info = SignersSize { total_signers, threshold };
+            // TODO: add checks to make sure threshold is not bigger then signature size
```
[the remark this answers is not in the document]
JesseAbram: ya imma add them now

node/cli/src/chain_spec/testnet.rs
JesseAbram: actually covered in #941

PR #946, opened 2024-07-17
Bump Synedrion to latest `master`

HCastano:
The latest `HEAD` of Synedrion includes child key derivation functionality that I need
for the new registration flow, so I had to update the dependency.

Two things to note:
- I'm using my own branch of Synedrion until
  https://github.com/entropyxyz/synedrion/pull/133 gets merged since that adds some
  `serde` support that I need
- There are one or two `TODO`s in places where I want to double check that functionality
  remained the same, but I'll do that tomorrow. I'd still consider the PR as a whole
  ready for review

### DIFF
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -25,7 +25,7 @@ use synedrion::{
     k256::EncodedPoint,
     make_aux_gen_session, make_interactive_signing_session, make_key_init_session,
     make_key_resharing_session,
-    sessions::{FinalizeOutcome, Session},
+    sessions::{FinalizeOutcome, Session, SessionId as SynedrionSessionId},
     signature::{self, hazmat::RandomizedPrehashSigner},
     AuxInfo, KeyResharingInputs, KeyShare, NewHolder, OldHolder, PrehashedMessage,
     RecoverableSignature, ThresholdKeyShare,
@@ -178,7 +178,7 @@ pub async fn execute_signing_protocol(
 
     let session = make_interactive_signing_session(
         &mut OsRng,
-        &session_id_hash,
+        SynedrionSessionId::from_seed(session_id_hash.as_slice()),
         pair,
         &party_ids,
         key_share,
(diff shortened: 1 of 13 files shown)


JesseAbram (approved):
looks good from my end but should also check with @fjarri before merging

crates/protocol/src/execute_protocol.rs
```diff
@@ -357,10 +378,11 @@ pub async fn execute_proactive_refresh(
 fn get_key_init_parties(
     my_party_id: &PartyId,
     threshold: usize,
-    validators: &[PartyId],
+    validators: &BTreeSet<PartyId>,
     session_id_hash: &[u8],
-) -> Result<(Vec<PartyId>, bool), ProtocolExecutionErr> {
-    let mut parties = vec![];
+) -> Result<(BTreeSet<PartyId>, bool), ProtocolExecutionErr> {
+    let validators = validators.iter().cloned().collect::<Vec<PartyId>>();
```
ameba23, line 384: i guess for this to work we need this to turn the set into a vector with consistent ordering. Since the tests pass this must be the case.

crates/protocol/tests/protocol.rs
```diff
@@ -159,7 +161,15 @@ async fn test_dkg_and_sign_with_parties(num_parties: usize) {
         pairs.iter().map(|pair| ValidatorSecretInfo::pair_only(pair.clone())).collect();
     let session_id = SessionId::Dkg { user: AccountId32([0; 32]), block_number: 0 };
     let outputs = test_protocol_with_parties(dkg_parties, session_id, threshold).await;
-    let signing_committee = &ids[..threshold];
+
+    // TODO (Nando): Double check to see if this is actually equivalent to previous code
```
ameba23, line 165: Pretty sure this is equivalent.

ameba23 (approved):
Looks great.  And once this is merged we get some other nice things such as being able to get the session id from a synedrion protocol message. 

Normally i would discourage using different versions of synedrion in different crates but since you are planning to fix as soon i am approving.

crates/protocol/src/execute_protocol.rs
HCastano: I believe this works because `BTreeSet` requires that any item inserted into the set implements `Ord`, meaning that the contents of the set will be ordered at any point in time.

crates/protocol/src/execute_protocol.rs
```diff
@@ -206,24 +209,26 @@ pub async fn execute_dkg(
     tracing::debug!("Executing DKG");
     let broadcaster = chans.0.clone();
 
-    let mut party_ids: Vec<PartyId> =
+    let party_ids: BTreeSet<PartyId> =
```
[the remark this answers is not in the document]
HCastano: 🙌 thank mr bogdan 🙌

PR #952, opened 2024-07-18
Split jumpstart and register flows

HCastano:
Right now the jumpstart flow makes use of the registration flow's DKG infrastructure.

Since registration is not going to be using this in the future it makes sense to split
them up and give jumpstart a bit more of a dedicated codepath.

To do so I've added a new endpoint to the TSS, `/generate_network_key`, which behind the
scenes uses the same code that the `/user/new` endpoint does.


pallets/propagation/src/lib.rs
```diff
@@ -83,12 +84,70 @@ pub mod pallet {
...
+                return Err(http::Error::Unknown);
+            }
+            let _res_body = response.body().collect::<Vec<u8>>();
+
+            Self::deposit_event(Event::DkgMessagePassed(req_body));
+
+            Ok(())
+        }
+
+        pub fn post_user_registration(block_number: BlockNumberFor<T>) -> Result<(), http::Error> {
```
JesseAbram, line 144: docs

pallets/propagation/src/lib.rs
```diff
@@ -83,12 +84,70 @@ pub mod pallet {
 
     impl<T: Config> Pallet<T> {
         pub fn post_dkg(block_number: BlockNumberFor<T>) -> Result<(), http::Error> {
```
JesseAbram, line 86: docs

pallets/registry/src/lib.rs
```diff
@@ -180,6 +180,11 @@ pub mod pallet {
     pub type Registering<T: Config> =
         StorageMap<_, Blake2_128Concat, T::AccountId, RegisteringDetails<T>, OptionQuery>;
 
+    #[pallet::storage]
```
JesseAbram, line 183: docs

JesseAbram (approved):

PR #954, opened 2024-07-19
Fix warnings when running Staking tests

HCastano:
There were some warnings around unused `Result`s and functions, so I've fixed those. I also did a
tiny cleanup with the types.


ameba23 (approved):
:100:

PR #955, opened 2024-07-19
New on-chain registration flow

### LINKED ISSUE #961: Using synedrion as a dependency of a pallet
@HCastano wants to use synedrion as a dependency of a pallet - probably the registry pallet.

This means it needs to work on wasm with no-std - which it should do since it doesn't use the standard library and is tested on wasm.

Hernando tried making an empty `#![no_std]` crate with the following two dependencies:

```toml
synedrion = { path="<my local copy>", default-features=false }
frame-support = { version="29.0.0", default-features=false }
```

But on compiling gets the following error:

```
❯ cargo check --no-default-features --target wasm32-unknown-unknown
    Checking sp-io v31.0.0
error[E0152]: found duplicate lang item `panic_impl`
    --> /Users/hcastano/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sp-io-31.0.0/src/lib.rs:1768:1
     |
1768 | / pub fn panic(info: &core::panic::PanicInfo) -> ! {
1769 | |     let message = sp_std::alloc::format!("{}", info);
1770 | |     #[cfg(feature = "improved_panic_error_reporting")]
1771 | |     {
...    |
1778 | |     }
1779 | | }
     | |_^
     |
     = note: the lang item is first defined in crate `std` (which `serde` depends on)
     = note: first definition in `std` loaded from /Users/hcastano/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libstd-2bf3225a2841264f.rlib
     = note: second definition in the local crate (`sp_io`)

For more information about this error, try `rustc --explain E0152`.
error: could not compile `sp-io` (lib) due to 1 previous error
```

Doing `cargo tree --edges features` shows that serde's `std` feature is introduced by dependencies of `frame-support`, not `synedrion` - which is why i have made this issue here rather than in synedrion. But strangely, if we remove synedrion as a dependency, the crate will build, even though we can see that the `std` feature of serde is still present with cargo-tree.

But `frame-support` is used by our pallets with no issues.

And if i add synedrion as a dependency of the registry pallet, the pallet compiles, as does the `entropy` binary.  So although this seems strange, im not sure its an issue.  Or maybe i just can't see the issue.

HCastano:
This PR adds a new extrinsic, `Registry::register_on_chain()`, which uses the new child
key derivation feature in Synedrion in order to simplify the registration process.

Since we can derive a child verification key from a parent verification key, all we need
to do for user registration is derive a user's key from the network key and store it
on-chain (alongside some program info). We don't need to reach out to the TSS to do any
cryptographic operations for us.

This PR is the first of a few that will be required to fully integrate the new
registration flow. In follow up PRs I plan to:

- Improve the mechanism used to derive accounts
- Remove old code related to registration (on the chain and TSS sides)
- Consolidate our old tests to use the new extrinsic and ensure we don't lose any
  coverage

Closes #961.


HCastano:
The CI is failing because Synedrion isn't able to compile to the `wasm32-unknown-unknown` target right now. Talking with Bogdan about this and will hopefully have it sorted soon 🤞

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+
+            let network_verifying_key =
+                if let Some(key) = <JumpStartProgress<T>>::get().verifying_key {
+                    SynedrionVerifyingKey::try_from(key.as_slice())
+                        .expect("The network verifying key must be valid.")
+                } else {
+                    return Err(Error::<T>::JumpStartNotCompleted.into());
+                };
+
+            // TODO (Nando): For a `CountedStorageMap` there is the possibility that the counter
```
HCastano, line 782: I'll swap this out with an issue after we go through the review process a bit in case there are discussions around this that pop up

pallets/registry/Cargo.toml
```diff
@@ -12,9 +12,11 @@ publish   =false
 targets=['x86_64-unknown-linux-gnu']
 
 [dependencies]
+bip32     ={ version="0.5.2", default-features=false, features=["alloc"] }
 codec     ={ package="parity-scale-codec", version="3.6.3", default-features=false, features=["derive"] }
-scale-info={ version="2.11", default-features=false, features=["derive"] }
 log       ={ version="0.4.22", default-features=false }
+scale-info={ version="2.11", default-features=false, features=["derive"] }
+synedrion ={ git="https://github.com/entropyxyz/synedrion/", branch="hc/bump-deps-for-no-std-builds", default-features=false }
```
HCastano, line 19: Waiting on https://github.com/entropyxyz/synedrion/pull/140 to get merged before changing this

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+                            Err(Error::<T>::NoProgramSet)
+                        }
+                    },
+                )?;
+            }
+
+            let network_verifying_key =
+                if let Some(key) = <JumpStartProgress<T>>::get().verifying_key {
+                    SynedrionVerifyingKey::try_from(key.as_slice())
+                        .expect("The network verifying key must be valid.")
```
ameba23, line 777: Did you mean to leave this `expect` in? I guess it should be impossible to get into a state where it is invalid

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+            // can decrease as storage entries are removed from the map. In our case we don't ever
+            // remove entries from the `RegisteredOnChain` map so the counter should never
+            // decrease. If it does we will end up with the same verifying key for different
+            // accounts, which would be bad.
+            //
+            // For a V1 of this flow it's fine, but we'll need to think about a better solution
+            // down the line.
+            let count = RegisteredOnChain::<T>::count();
+            let path =
+                bip32::DerivationPath::from_str(&scale_info::prelude::format!("m/0/{}", count))
```
ameba23, line 792: im guessing the `m/0` bit is like a version number that we can bump if the derivation process changes?

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+            program_modification_account: T::AccountId,
+            programs_data: BoundedVec<ProgramInstance<T>, T::MaxProgramHashes>,
+        ) -> DispatchResultWithPostInfo {
+            use core::str::FromStr;
+            use synedrion::{ecdsa::VerifyingKey as SynedrionVerifyingKey, DeriveChildKey};
+
+            let signature_request_account = ensure_signed(origin)?;
+
+            ensure!(
+                signature_request_account.encode() != NETWORK_PARENT_KEY.encode(),
```
ameba23, line 752: I dont think we need this check anymore, as it was to stop the network parent keyshare getting clobbered by a user keyshare in the kvdb.  But here the user has no keyshare to be stored.

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+                bip32::DerivationPath::from_str(&scale_info::prelude::format!("m/0/{}", count))
+                    .map_err(|_| Error::<T>::InvalidBip32DerivationPath)?;
+            let child_verifying_key = network_verifying_key
+                .derive_verifying_key_bip32(&path)
+                .map_err(|_| Error::<T>::Bip32AccountDerivationFailed)?;
+
+            let child_verifying_key = BoundedVec::try_from(
+                child_verifying_key.to_encoded_point(true).as_bytes().to_vec(),
+            )
+            .expect("Synedrion must have returned a valid verifying key.");
```
ameba23, line 801: agree that this could never happen but i still kinda don't like having `expect`s

ameba23 (approved):
:100: looks great.

Only concern i have about using a counter as input is that anybody can know what the next entropy account public key is going to be before it is made. As soon as the network jump starts everyone can know the complete set of all the entropy account public keys there will ever be.

I can't think of a reason why this is bad - but my hunch is that concatenating something from the user (eg: signature request account id and/or program modification account id) to the counter in the derivation path feels safer.

pallets/registry/src/lib.rs
```diff
@@ -200,6 +210,16 @@ pub mod pallet {
...
         StorageMap<_, Blake2_128Concat, VerifyingKey, RegisteredInfo<T>, OptionQuery>;
 
+    /// An item tracking all the users registered on the Entropy network.
+    ///
+    /// Notice that the registration state does not depend on any Substrate account being
+    /// registered, but rather a _verifying key_, which represents the user beyond the scope of the
+    /// Entropy network itself (e.g it can be an account on Bitcoin or Ethereum).
+    #[pallet::storage]
+    #[pallet::getter(fn registered_on_chain)]
+    pub type RegisteredOnChain<T: Config> =
```
JesseAbram, line 220: Will the name of this change when we remove the old code? If so I suggest something more terrible so we make sure to change it like Registered_2 or something dumb like that

pallets/registry/src/lib.rs
```diff
@@ -438,7 +461,7 @@ pub mod pallet {
             }
 
             Dkg::<T>::try_mutate(block_number, |messages| -> Result<_, DispatchError> {
-                messages.push(encoded_sig_req_account);
+                messages.push(sig_req_account.encode());
```
JesseAbram, line 464: In this case here we do the encoding twice instead of once, is this really the best way to handle this?

pallets/registry/src/lib.rs
JesseAbram: yaaaa but this may be one of those I don't know what this could lead to but I have a bad feeling about it not being in here

pallets/registry/src/lib.rs
JesseAbram: ya def remove it for proper error handling

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+            let num_programs = programs_data.len();
+            ensure!(num_programs != 0, Error::<T>::NoProgramSet);
+
+            // Change program ref counter
+            for program_instance in &programs_data {
+                pallet_programs::Programs::<T>::try_mutate(
+                    program_instance.program_pointer,
+                    |maybe_program_info| {
+                        if let Some(program_info) = maybe_program_info {
+                            program_info.ref_counter = program_info.ref_counter.saturating_add(1);
```
JesseAbram, line 765: ok this is a really really nit but I should bring it up, saturating can overflow back to 0 and then program can be pulled.....really unlikely but probably should be a checked_add here

pallets/registry/src/lib.rs
JesseAbram: ya also are we not moving from a counter to using the account or is there a possibility we would keep the counter?

pallets/registry/src/lib.rs
JesseAbram, line 792: This looks good but what path is the parent key at and we def need to block it, if the parent key is at 0 we need to start the count from 1

pallets/registry/src/lib.rs
JesseAbram: same

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+            RegisteredOnChain::<T>::insert(
+                child_verifying_key.clone(),
+                RegisteredInfo {
+                    programs_data,
+                    program_modification_account: program_modification_account.clone(),
+                    version_number: T::KeyVersionNumber::get(),
+                },
+            );
+
+            ModifiableKeys::<T>::try_mutate(
```
JesseAbram, line 812: This should go above RegisteredOnChain because it might fail, pretty sure atomicity now but optimization

pallets/registry/src/benchmarking.rs
```diff
@@ -163,6 +163,91 @@ benchmarks! {
...
+        balance,
+    );
+  }: _(
+      RawOrigin::Signed(signature_request_account.clone()),
+      program_modification_account,
+      programs_info
+  )
+  verify {
+    use core::str::FromStr;
+    use synedrion::DeriveChildKey;
```
JesseAbram, line 219: nit: gross put this at the top of the file

JesseAbram (commented):
looks good in flow, won't approve till you do negative tests

pallets/registry/src/lib.rs
HCastano: Yeah, the idea would be to turn this into the `Registered` struct

pallets/registry/src/lib.rs
HCastano: We're only encoding 32 bytes, so it's not that expensive. I personally find this more legible, but if you really think it's an issue I can revert this change.

pallets/registry/src/lib.rs
```diff
@@ -704,6 +727,105 @@ pub mod pallet {
...
+        #[pallet::weight({
+            <T as Config>::WeightInfo::register_on_chain(<T as Config>::MaxProgramHashes::get())
+        })]
+        pub fn register_on_chain(
+            origin: OriginFor<T>,
+            program_modification_account: T::AccountId,
+            programs_data: BoundedVec<ProgramInstance<T>, T::MaxProgramHashes>,
+        ) -> DispatchResultWithPostInfo {
+            use core::str::FromStr;
+            use synedrion::{ecdsa::VerifyingKey as SynedrionVerifyingKey, DeriveChildKey};
```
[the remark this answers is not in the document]
HCastano: We only use these traits and structs in this specific extrinsic, so I wanted to avoid littering the whole module with them 🤷‍♂️

pallets/registry/src/lib.rs
HCastano: Yes this is intentional. If this `expect` is hit it means there's a bug somewhere in the code (we're storing an invalid verifying key), and I would like users to log a report about it.

pallets/registry/src/lib.rs
HCastano: Same thing here, this would be a bug on the Synedrion side and I would like a bug report about it instead of "handling" it.

pallets/registry/src/lib.rs
HCastano: All extrinsics are transactional, so the ordering doesn't matter anymore.

I'd like to keep it here because it make more sense to me in terms of the flow, and it doesn't break up the "core" registration code

pallets/registry/src/benchmarking.rs
HCastano: Same thing as in the other comment, we only use these here so I didn't want to litter all the imports

pallets/registry/src/lib.rs
HCastano: If we use `m/0/0` it still ends up with a different private key. From this [bip32 website](http://bip32.org/):

image

JesseAbram: true Im fine, should be a really easy test, just one assert statment in the test, probably worth it?

HCastano: No, the `m/.../...` bit is the actual derivation path.

If you take a look at the [graphic here](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#master-key-generation) you can see that it's sort of like a namespace. So accounts under `m/0/*` are all associated with each other, while those under `m/1/*` are distinct from `m/0/*` (but associated with each other).

pallets/registry/src/lib.rs
```diff
@@ -158,13 +158,23 @@ pub mod pallet {
     #[pallet::genesis_build]
     impl<T: Config> BuildGenesisConfig for GenesisConfig<T> {
         fn build(&self) {
-            for account_info in &self.registered_accounts {
-                assert!(account_info.1.len() as u32 == VERIFICATION_KEY_LENGTH);
+            for (account_id, verifying_key) in &self.registered_accounts {
```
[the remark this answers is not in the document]
HCastano: 🙏

pallets/registry/src/lib.rs
HCastano: There is the possibility we keep the counter, but it would probably be in the context of something like `hash(account ID) ++ counter`

pallets/registry/src/lib.rs
HCastano: A `saturating_add()` would end up saturating at the upper bound of the type being added. In this case that means that this would have a max value of `u128::MAX`, no overflow would occur.

pallets/registry/src/lib.rs
HCastano: In the jumpstart PR where this was introduced [I mentioned](https://github.com/entropyxyz/entropy-core/pull/918#discussion_r1672631761) that allowing registration from the parent key could be cool. 

While a bit weird, it would allow the network to have a registered account for use on other networks. Maybe it could allow for governance actions on Bitcoin or Ethereum for example. I'd be down to remove this check as long as its cool with both of you

pallets/registry/src/lib.rs
HCastano: I don't think this is possible as long as _some_ sort of derivation step exists in the extrinsic.

What do you think a test for this should look like though? Would we just check the `m/0/0` path, or enumerate all the paths and check those?

pallets/registry/src/lib.rs
JesseAbram: I think leave it in make an issue and this is actually a good retreat topic

pallets/registry/src/lib.rs
JesseAbram: leaving it up to you on both of these comments, I guess the question is what is more important slightly better optimization or slightly better readability

pallets/registry/src/lib.rs
JesseAbram: just in the positive register test you can take the key you put as the parent network key derived off and assert there is not program data for that, one line no?

JesseAbram (approved):

ameba23:
@HCastano @JesseAbram - one thing that just occurred to me about this:

When we come to do the signing flow, i assume the signature request is going to specify the verifying key, and that to sign a message the threshold server will need to know the derivation path.

So we would need a way of looking up what count number a given verifying key has.  Maybe `CountedStorageMap` gives us some easy way to get the count of a given key - but i can't see it. So maybe we need to store the count in `RegisteredInfo`.

HCastano:
@ameba23 

> Only concern i have about using a counter as input is that anybody can know what the next entropy account public key is going to be before it is made. As soon as the network jump starts everyone can know the complete set of all the entropy account public keys there will ever be.
>
> I can't think of a reason why this is bad - but my hunch is that concatenating something from the user (eg: signature request account id and/or program modification account id) to the counter in the derivation path feels safer.

Yeah knowing the public key of an account isn't inherently dangerous. Maybe there are some scenarios where this could be used in some way, but I don't think it's any different from having access to a vanity generator (e.g if you want an account like `0xbeef...`. We should definitely incorporate some element other than the counter though.

> When we come to do the signing flow, i assume the signature request is going to specify the verifying key, and that to sign a message the threshold server will need to know the derivation path.
>
> So we would need a way of looking up what count number a given verifying key has. Maybe CountedStorageMap gives us some easy way to get the count of a given key - but i can't see it. So maybe we need to store the count in RegisteredInfo.

You're right here, and it's something I was going to deal with in the signing flow PRs.

pallets/registry/src/lib.rs
HCastano: Opened https://github.com/entropyxyz/entropy-core/issues/983.

pallets/registry/src/lib.rs
HCastano: Opened https://github.com/entropyxyz/entropy-core/issues/984.

PR #956, opened 2024-07-20
Remove session hash from protocol message 

ameba23:
Merging https://github.com/entropyxyz/entropy-core/pull/946 means we can now get the session Id from synedrion protocol messages and don't need to store it a second time in `ProtocolMessage`.


crates/protocol/src/execute_protocol.rs
```diff
@@ -74,6 +74,7 @@ async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
     session: Session<Res, sr25519::Signature, PairWrapper, PartyId>,
     session_id_hash: [u8; 32],
 ) -> Result<(Res::Success, mpsc::Receiver<ProtocolMessage>), GenericProtocolError<Res>> {
+    let session_id = synedrion::SessionId::from_seed(&session_id_hash);
```
HCastano, line 77: I didn't realize this was being re-exported at the top level so I have it imported as `SynedrionSessionId`. Let's use the top level import though, as you have here, elsewhere

HCastano (approved):

JesseAbram (approved):

Issue #941, opened 2024-07-16
Validator joining flow


JesseAbram:
## Core
- [x] Set parameters for size of signer group and T of N
- [x] Add Signer groups to chain 
- [x] Get randomness to select new signer from validators on session refresh
- [x] Rotate Signers 
- [x] Create ocw message to do reshare
- [x] Allow nodes to call in to confirm (holding new key share in seperate storage slot) 
- [x] On confirmation have chain update new signers 
- [ ] When all confirmed (or T idk) create another message to rotate new keyshare to old keyshare, delete old keyshare #968 
- [x] Validate new_reshare pathway
- [x] Delete old key if not a new signer (includes making the test for reshare in TSS actually rotate the signers not just do a reshare)
- [x] Replace SIGNING_PARTY_SIZE to TOTAL_SIGNERS or SIGNER_THRESHOLD

## Extras

- [x] Have next_signers and signers bounded by T and N in parameters
- [x] Store current threshold info with verifying key onchain
- [ ] Add checks to what a signer needs to have before selected (balance etc) #998 
- [ ] Have a signature or something in confirming reshare done correctly #981 
- [x] Determine how initial signers are selected  and set parameter bound on configs (should be in jumpstart)
- [ ] Do not allow validator unbonding if they are a signer #942 
- [x] Refactor reshare to be more resuable (reshare and refresh pathways should be mostly the same
- [x] Add test for new signer without key added to network (currently they all have keys

JesseAbram:
first four handled in #938

JesseAbram:
As all of the task still left are open for input, have been pulled out to their own issues and as I would consider them enhancements I am closing this as done

PR #970, opened 2024-07-30
validate new reshare

JesseAbram:
Related #941 

- [x] Adds validate reshare

### DIFF
--- a/crates/threshold-signature-server/src/helpers/launch.rs
+++ b/crates/threshold-signature-server/src/helpers/launch.rs
@@ -45,6 +45,8 @@ pub const DEFAULT_DAVE_MNEMONIC: &str =
 pub const DEFAULT_EVE_MNEMONIC: &str =
     "impact federal dish number fun crisp various wedding radio immense whisper glue";
 pub const LATEST_BLOCK_NUMBER_NEW_USER: &str = "LATEST_BLOCK_NUMBER_NEW_USER";
+pub const LATEST_BLOCK_NUMBER_RESHARE: &str = "LATEST_BLOCK_NUMBER_RESHARE";
+
 pub const LATEST_BLOCK_NUMBER_PROACTIVE_REFRESH: &str = "LATEST_BLOCK_NUMBER_PROACTIVE_REFRESH";
 
 #[cfg(any(test, feature = "test_helpers"))]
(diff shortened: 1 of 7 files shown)


crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -373,6 +375,19 @@ pub async fn setup_latest_block_number(kv: &KvManager) -> Result<(), KvError> {
             .await
             .expect("failed to update latest block number");
     }
+    let exists_result_new_user =
```
ameba23, line 378: nitpick:

```suggestion
    let exists_result_reshare =
```

pallets/staking/src/lib.rs
```diff
@@ -231,12 +233,22 @@ pub mod pallet {
                 proactive_refresh_keys: self.proactive_refresh_data.1.clone(),
             };
             ProactiveRefresh::<T>::put(refresh_info);
-
+            // mocks a signer rotation for tss new_reshare tests
```
ameba23, line 236: If this is staying long term i think it would be ideal if it was kept out of the production binary. Normally i would say put it in a `if cfg!(test) {}` - but we can't do that because we test against the release version of the chain node.
Maybe it could go behind a feature flag.

ameba23 (approved):
:1st_place_medal:

pallets/staking/src/lib.rs
JesseAbram: ya but then we would have to build it like that for tests, this line of code goes into the genesis file but only if you specify in the chainspec file. The only chainspec file that does this is integration test, so this does stay out of the production binary

JesseAbram: that being said this is because I need state that isn't changeable (only on session and idk if a 4 hour long test makes sense) if we go the feature flag route we can make a generalizable test function that changes any state......it would get rid of this, idk if we are cool with that im fine with it as a refactor (looping in @HCastano)

HCastano: @JesseAbram I do agree with Peg here, I would like to keep test related code more clearly marked as such for future readers. 

To be fair to Jesse though, the compiler would remove this code for the condition that `mock_signer_rotate` is `false`, which should be the case for production builds.

On the session length, our tests should have this configured to be less than four hours (like a few blocks, maybe?). And we can always force a session rotation in test code as well

Issue #969, opened 2024-07-30
Confusion around constants for t and n


ameba23:
`entropy-shared` has the following constants:
https://github.com/entropyxyz/entropy-core/blob/f712a8524e2d0ab1b0598fcc727c36f00e5b154c/crates/shared/src/constants.rs#L36

https://github.com/entropyxyz/entropy-core/blob/f712a8524e2d0ab1b0598fcc727c36f00e5b154c/crates/shared/src/constants.rs#L67

https://github.com/entropyxyz/entropy-core/blob/f712a8524e2d0ab1b0598fcc727c36f00e5b154c/crates/shared/src/constants.rs#L70

`SIGNING_PARTY_SIZE` is used by the registry pallet to check we have the right number of confirmations following network DKG.

But i would expect this to be 3, not 2.  I think it would be good to have just two constants to avoid confusion.

JesseAbram:
yes Signing party size should be replaced by total signers and signer threshold, will add this to #941

ameba23:
Fixed in https://github.com/entropyxyz/entropy-core/pull/971

PR #971, opened 2024-07-31
Set inital signers

JesseAbram:
Related #941 

- [x] Set inital signers in jumpstart call
- [x] Remove SIGNING_PARTY_SIZE


crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -104,6 +104,7 @@ pub async fn new_reshare(
         .any(|validator_info| validator_info.tss_account == *signer.account_id());
 
     if !is_proper_signer {
+        dbg!("not proper");
```
ameba23, line 107: Either remove or change to `tracing::warn` or whatever

pallets/staking/src/lib.rs
```diff
@@ -234,20 +232,21 @@ pub mod pallet {
...
                 NextSigners::<T>::put(NextSignerInfo {
-                    next_signers: self.inital_signers.clone(),
+                    next_signers: self.mock_signer_rotate.clone().1,
                     confirmations: vec![],
                 });
 
                 ReshareData::<T>::put(ReshareInfo {
                     // To give enough time for test_reshare setup
                     block_number: TEST_RESHARE_BLOCK_NUMBER.into(),
                     // Alice signer public key
```
ameba23, line 248: Not sure if this comment should stay - is it still always alice here?

pallets/staking/src/lib.rs
```diff
@@ -199,11 +199,10 @@ pub mod pallet {
     #[derive(DefaultNoBound)]
     pub struct GenesisConfig<T: Config> {
         pub threshold_servers: Vec<ThresholdServersConfig<T>>,
-        pub inital_signers: Vec<T::ValidatorId>,
         /// validator info and accounts to take part in proactive refresh
         pub proactive_refresh_data: (Vec<ValidatorInfo>, Vec<Vec<u8>>),
         /// validator info and account to take part in a reshare
```
ameba23, line 204: This doccoment could be improved because its not very clear what the second and third things in the tuple are.

It looks like the first one is the set of initial signers, and the second one is the first new next signer which will be chosen.  But i am not sure why it is a vec if we only ever use the first element.

crates/shared/src/constants.rs
```diff
@@ -33,8 +33,6 @@ lazy_static! {
...
 }
 
-pub const SIGNING_PARTY_SIZE: usize = 2;
```
ameba23, line 36: :+1:

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -785,7 +785,7 @@ async fn test_jumpstart_network() {
 
     // wait for jump start event check that key exists in kvdb
     let mut got_jumpstart_event = false;
-    for _ in 0..45 {
+    for _ in 0..75 {
```
ameba23, line 788: Just to check - did you bump this when you were having some issue which you then resolved, or does the initial DKG just take longer as there are now 3 parties rather than 2.

ameba23 (approved):
:100: looks great.  we can close https://github.com/entropyxyz/entropy-core/issues/969

pallets/staking/src/lib.rs
JesseAbram: no you are right

crates/threshold-signature-server/src/user/tests.rs
JesseAbram: ya I was getting some inconsistent failings in my last PR for this on the CI, I couldn't always hit it, so I added an assert statment here in the last PR and if I saw it again I would now have an error and know it was here

PR #974, opened 2024-08-01
Add parent key threshold dynamically

JesseAbram:
Related #941 

- [x] Threshold stored with parent key
- [x] Reshare uses said storing


pallets/staking/src/lib.rs
```diff
@@ -169,6 +175,36 @@ pub mod pallet {
...
+        RuntimeDebug,
+        TypeInfo,
+        frame_support::DefaultNoBound,
+    )]
+    #[scale_info(skip_type_params(T))]
+    pub struct JumpStartDetails<T: Config> {
+        pub jump_start_status: JumpStartStatus,
+        pub confirmations: Vec<T::ValidatorId>,
+        pub verifying_key: Option<VerifyingKey>,
+        pub parent_key_threhsold: u8,
```
ameba23, line 205:
```suggestion
        pub parent_key_threshold: u8,
```

ameba23 (approved):
Looks good - i don't know if the decision to move the jump start stuff to staking pallet was because it logically belongs there or because in needed to be there in order to mutate `JumpStartProgress`. But fine by me either way.

I still think we should treat changing the threshold with extreme caution until we have tested it a lot but as i said before i can't see why it wouldn't work just fine.

JesseAbram:
> Looks good - i don't know if the decision to move the jump start stuff to staking pallet was because it logically belongs there or because in needed to be there in order to mutate `JumpStartProgress`. But fine by me either way.
> 
It was the second one although since it is related to the parent key it may make sense in there but registry takes in staking so that means staking can't pull in registry or else we get circular dependency issues 

> I still think we should treat changing the threshold with extreme caution until we have tested it a lot but as i said before i can't see why it wouldn't work just fine.

Agreed

pallets/staking/src/lib.rs
JesseAbram: sometimes I think Im dyslexic, who knows

Issue #976, opened 2024-08-02
Both `KeyShare` and `AuxInfo` need to be stored on key reshare


ameba23:
This is something i missed when reviewing the key reshare PRs: https://github.com/entropyxyz/entropy-core/pull/970 and https://github.com/entropyxyz/entropy-core/pull/965

To sign a message with synedrion we need both a `ThresholdKeyShare` and `AuxInfo`.  `AuxInfo` is created as part of our DKG process, and stays the same following a refresh (a reshare with the parties staying the same). 

The type we store following a DKG is:
https://github.com/entropyxyz/entropy-core/blob/43061486dd8f10a24b0c733eff9f5e1cf6a2313a/crates/protocol/src/lib.rs#L135

That is, we store a tuple with both the keyshare and aux info. 

However when we do a reshare where the parties change, there will be a party who doesn't yet have any `aux_info`.  This means we need to run the aux gen protocol, ideally in the same protocol session as the reshare, as we do in DKG (where we run key init, reshare and aux gen all in one protocol session).  

To expose issues like this, we should really be signing a message after re-sharing in our reshare tests.

PR #977, opened 2024-08-02
Run the aux gen protocol when resharing to a new signing committee

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/976

When running the reshare protocol, we also need to re-generate aux info and store the aux info together with the new keyshare.

When running the refresh protocol (which is just a reshare with the parties not changing), we need to store the old aux info together with the new keyshare - as both are stored under the same key in the key-value db.


crates/protocol/src/lib.rs
```diff
@@ -214,9 +214,9 @@ impl SessionId {
     }
 }
 
-/// A sub-protocol of the DKG protocol
+/// A sub-protocol of the DKG or reshare protocols
 #[derive(Debug)]
-pub enum DkgSubsession {
+pub enum Subsession {
```
ameba23, line 219: I renamed this enum because it is no longer only used for DKG

crates/testing-utils/Cargo.toml
```diff
@@ -20,7 +20,7 @@ tokio             ={ version="1.39", features=["macros", "fs", "rt-multi-thread"
 axum              ={ version="0.7.5" }
 entropy-shared    ={ version="0.2.0", path="../shared" }
 entropy-kvdb      ={ version="0.2.0", path="../kvdb", default-features=false }
-entropy-tss       ={ version="0.2.0", path="../threshold-signature-server" }
+entropy-tss       ={ version="0.2.0", path="../threshold-signature-server", features = ["test_helpers"] }
```
ameba23, line 23: Unrelated to this PR but i needed this for testing locally

JesseAbram (approved):

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ At the moment this project **does not** adhere to
 - Reshare confirmation ([#965](https://github.com/entropyxyz/entropy-core/pull/965))
 - Set inital signers ([#971](https://github.com/entropyxyz/entropy-core/pull/971))
 - Add parent key threshold dynamically ([#974](https://github.com/entropyxyz/entropy-core/pull/974))
+- Run the aux gen protocol when resharing to a new signing committee ([#977](https://github.com/entropyxyz/entropy-core/pull/977))
```
HCastano, line 25: This is more of an internal change, don't think we need to have this

crates/threshold-signature-server/src/signing_client/api.rs
```diff
@@ -112,7 +112,18 @@ pub async fn proactive_refresh(
                 ocw_data.block_number,
             )
             .await?;
-            let serialized_key_share = key_serialize(&new_key_share)
+
+            // Get aux info from existing entry
+            let aux_info = {
+                let existing_entry = app_state.kv_store.kv().get(&key).await?;
+                let (_old_key_share, aux_info): KeyShareWithAuxInfo = deserialize(&existing_entry)
```
HCastano, line 119: I would add a comment here indicating that we intentionally want to use the old `aux_info` in the case of a refresh

HCastano (approved):
Good stuff 👍

PR #979, opened 2024-08-02
Update `cargo-deny` to version `2`

HCastano:
This new version removed some keys that we were using, namely the `unlicense` and `copyleft` keys.
These fields are now set to `deny` by default as long as `version = 2` is set.

See https://github.com/EmbarkStudios/cargo-deny/pull/611.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,6 +1,7 @@
 [licenses]
-# The lint level for crates which do not have a detectable license
-unlicensed="deny"
+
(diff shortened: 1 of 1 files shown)


HCastano:
Looks like this was fixed in https://github.com/entropyxyz/entropy-core/pull/974.

@JesseAbram in the future do you mind pulling CI fixes like this into their own PRs? Thanks!

PR #990, opened 2024-08-07
Signing flow with derived accounts

HCastano:
This PR allows signature requests to happen using the new derived account flow that was
implemented in #955. I've implemented it in such a way that the old registration and signing
flow is also valid as I didn't want to break everything.

I want to have a few follow up PRs related to this in order to keep the review queue
manageable:
- After this gets merged I want to update the `entropy-test-cli` to support the new
  derived flows and ensure everything works
- Following that we can start getting rid of old code
- Related to the above, at that point I want to start improving the test coverage for
  both the registration and signing flows, using the tests from the old flows as a
  reference to try and avoid losing coverage
- Implement the message passing as described in the Tofino spec


pallets/registry/src/lib.rs
```diff
@@ -119,6 +119,8 @@ pub mod pallet {
     pub struct RegisteredInfo<T: Config> {
         pub programs_data: BoundedVec<ProgramInstance<T>, T::MaxProgramHashes>,
         pub program_modification_account: T::AccountId,
+        /// The SCALE encoded BIP-32 `DerivationPath` used to register this account.
+        pub derivation_path: Option<Vec<u8>>,
```
HCastano, line 123: @JesseAbram do you know what kind of runtime version bump we'd need for this? For the release it doesn't make a difference since we're bumping everything anyways, but I'm more curious in case we ever do this as a one-off thing

crates/client/src/user.rs
```diff
@@ -36,27 +36,39 @@ pub struct UserSignatureRequest {
...
     pub hash: HashingAlgorithm,
-    /// The veryfying key for the signature requested
+    /// The verifying key for the signature requested
     pub signature_verifying_key: Vec<u8>,
 }
 
 pub async fn get_signers_from_chain(
     api: &OnlineClient<EntropyConfig>,
     rpc: &LegacyRpcMethods<EntropyConfig>,
+    with_parent_key: bool,
```
JesseAbram, line 46: ehhh maybe a get signers from chain and a get validators from chain function makes more sense

pallets/registry/src/lib.rs
JesseAbram: tbh I just click all of them and let you tell me which one it was

crates/threshold-signature-server/Cargo.toml
```diff
@@ -57,8 +57,9 @@ uuid                    ={ version="1.10.0", features=["v4"] }
 
 # Misc
 tokio-tungstenite="0.23.1"
-bip39            ={ version="2.0.0", features=["zeroize"] }
 bincode          ="1.3.3"
+bip32            ={ version="0.5.2" }
+bip39            ={ version="2.0.0", features=["zeroize"] }
```
JesseAbram, line 62: why both?

JesseAbram (approved):

crates/client/src/user.rs
HCastano: I don't disagree, but for the case of this function it is always getting the signers. The difference between both arms here is that the signers are either: queried directly or grabbed from the validator set.

crates/threshold-signature-server/Cargo.toml
HCastano: `bip39` was already in the codebase, I just moved it in a sorted order. The crate is used for the TSS mnemonic

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -207,22 +213,41 @@ pub async fn sign_tx(
         request_author,
     };
 
-    let _has_key = check_for_key(&string_verifying_key, &app_state.kv_store).await?;
+    // In the new registration flow we don't store the verifying key in the KVDB, so we only do this
+    // check if we're using the old registration flow
```
ameba23, line 217: But we could still check here if we are currently in the signing committee. But i guess theres no point as we are about to attempt to get the parent keyshare in `get_sign_context` anyway

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -54,7 +54,10 @@ async fn test_reshare() {
     let alice = AccountKeyring::AliceStash;
 
     let cxt = test_node_process_testing_state(true).await;
-    let (_validator_ips, _validator_ids) = spawn_testing_validators(true).await;
+
+    let add_parent_key_to_kvdb = true;
```
ameba23, line 58: much clearer

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -362,6 +364,134 @@ async fn test_sign_tx_no_chain() {
     clean_tests();
 }
 
+#[tokio::test]
+#[serial]
+async fn signature_request_with_derived_account_works() {
```
ameba23, line 369: Could we now remove the new registration flow test below?  Is there anything in it which is not repeated here?

ameba23 (approved):
Great to see this working.  This all happened faster than i imagined it would.

And im looking forwards to seeing the old stuff removed and hopefully everything a bit cleaner.  Also looking forwards to being able to try this out with docker-compose and the test cli.

crates/threshold-signature-server/src/user/tests.rs
HCastano: I want to keep them separate. We already have enough tests that conflate what their actual purpose is. What I'd want to do instead is trim down some of the code in the signing test to be the bare minimum setup we need - but I wasn't sure what the best way to do that yet so I left it in full.

crates/threshold-signature-server/src/user/api.rs
HCastano: We weren't checking for inclusion in the signing committee here anyways, so maybe I'm misunderstanding your comment

crates/client/src/user.rs
```diff
@@ -36,27 +36,39 @@ pub struct UserSignatureRequest {
...
+
+        validators.sort();
+        validators
+    };
 
     // TODO #898 For now we use a fix proportion of the number of validators as the threshold
     let threshold = (validators.len() as f32 * 0.75) as usize;
+
     // TODO #899 For now we just take the first t validators as the ones to perform signing
-    validators.sort();
```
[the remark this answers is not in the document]
HCastano: I think it will still be called by both. Probably safer to keep the sorting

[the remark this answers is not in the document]
HCastano: @ameba23 when I sorted the signers coming from the on-chain flow the signing test started failing. Not entirely sure why 🤷

PR #991, opened 2024-08-08
Migrate circle-ci workflow to github actions

cooldracula:
This PR adds a number of github workflows that are a near direct port of the analogous jobs in the existing `.circleci/then.yml` file.

The specific mapping:

| circleci | gh action|
---------|----------|
| [lint](https://github.com/entropyxyz/entropy-core/blob/4316480dc79028d381bab36826ce1c6cff50d74c/.circleci/then.yml#L172) | [lint-and-check-licenses.yaml](https://github.com/entropyxyz/entropy-core/compare/zach/migrate-circleci-jobs?expand=1#diff-60f46ff916bf65cef79d0a98a4f93b20be02c472920779cf4ff85678f4055039) |
| [test](https://github.com/entropyxyz/entropy-core/blob/4316480dc79028d381bab36826ce1c6cff50d74c/.circleci/then.yml#L175) | [test-runtime-benchmarks.yaml](https://github.com/entropyxyz/entropy-core/compare/zach/migrate-circleci-jobs?expand=1#diff-af679a38508589612e5fc871291218a0cdc887d0aaaf8744dcdca7b5cfb20c90) |
| [build](https://github.com/entropyxyz/entropy-core/blob/4316480dc79028d381bab36826ce1c6cff50d74c/.circleci/then.yml#L178) | [build-and-run-node-test.yaml](https://github.com/entropyxyz/entropy-core/compare/zach/migrate-circleci-jobs?expand=1#diff-8f0de702d1c093acb696d3313281aaf0f64161d430f621228a280b9d2f197c1f) |
| [chainspecs](https://github.com/entropyxyz/entropy-core/blob/4316480dc79028d381bab36826ce1c6cff50d74c/.circleci/then.yml#L187) | [check-chainspecs.yaml](https://github.com/entropyxyz/entropy-core/compare/zach/migrate-circleci-jobs?expand=1#diff-ffb27273818e8752620d7fe6046b73c1441b222deaa3b8c1ce85ba6e1fe3963e) |
| [documentation](https://github.com/entropyxyz/entropy-core/blob/4316480dc79028d381bab36826ce1c6cff50d74c/.circleci/then.yml#L190) | [check-doc-build.yaml](https://github.com/entropyxyz/entropy-core/compare/zach/migrate-circleci-jobs?expand=1#diff-3fc19da8a7a008436f9ec1736f0ac3e75175f083f27120088d4479d8c8e90a3a) |

This PR would resolve one of the tasks in our devops-infra issue [#59](https://github.com/entropyxyz/devops-infrastructure/issues/59)

---

note: I did not remove the .circleci directory as part of this PR as there are other commands in the then.yml that, while not used in the jobs, might still be used or useful somewhere else? For example: [the comment-on-pr command](https://github.com/entropyxyz/entropy-core/blob/4316480dc79028d381bab36826ce1c6cff50d74c/.circleci/then.yml#L64)


ameba23 (approved):
:fire: great to have this. and kinda lucky that CI is failing because we can see that the same errors are coming up in circle CI as gh actions.

Re: the `comment-on-pr` thing, i don't know what that is for, i guess it makes a pr comment based on ci output? it looks like john added it.

Was there anything else in the circle CI file that isn't migrated?  To me it looks like thats everything.  If not then fine by me to go ahead and remove the circle ci script.

.github/actions/install-dependencies/action.yaml
```diff
@@ -0,0 +1,20 @@
+---
+name: 'Install Dependencies'
+description: 'Install the dependencies and rust components used across jobs'
```
HCastano, line 3:
```suggestion
description: 'Install the dependencies and Rust components used across jobs'
```

.github/actions/install-dependencies/action.yaml
```diff
@@ -0,0 +1,20 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Install dependencies
+      run: |
+        sudo sed -i "/#\$nrconf{restart} = 'i';/s/.*/\$nrconf{restart} = 'a';/" /etc/needrestart/needrestart.conf
+        sudo apt-get update
+        sudo apt install -y libssl-dev clang libclang-dev protobuf-compiler
+      shell: bash
+    - name: Add rust components
```
HCastano, line 13:
```suggestion
    - name: Add Rust components
```

.github/workflows/build-and-run-node-test.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Install dependencies
+        uses: ./.github/actions/install-dependencies/
+      - name: Build entropy-protocol nodejs package
+        run: |
+          cd crates/protocol
+          make build-nodejs-testing
+          cd nodejs-test
+          yarn
+          cd ../../..
+      - name: Run cargo test
```
HCastano, line 33:
```suggestion
      - name: Run `cargo build && cargo test`
```

.github/workflows/build-and-run-node-test.yaml
```diff
@@ -0,0 +1,40 @@
+---
+name: "Build and run node-test"
```
HCastano, line 2:
```suggestion
name: "Build and test"
```

Since we also test the TSS here, not just the node

HCastano (approved):
Thanks for this 🤩 

I haven't checked to see what the CI failures are about, but let me know if you need help addressing them

cooldracula:
> Thanks for this 🤩
> 
> I haven't checked to see what the CI failures are about, but let me know if you need help addressing them

Thank you (and @ameba23 ) for the review! For the CI failures, they are all failing during compilation due to the following Rust error:

```sh
error[E0282]: type annotations needed for `Box<_>`
  --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/time-0.3.34/src/format_description/parse/mod.rs:83:9
   |
83 |     let items = format_items
   |         ^^^^^
...
86 |     Ok(items.into())
   |              ---- type must be known at this point
   |
help: consider giving `items` an explicit type, where the placeholders `_` are specified
   |
83 |     let items: Box<_> = format_items
   |              ++++++++
```

It seems to be an error bubbling up from a dependency?  I would love help addressing it though.

ameba23:
@cooldracula this needs the `time` crates to be bumped to 0.3.36.  It looks like this is fixed now in master (there are constantly things being updated by dependabot). So i think if you merge or rebase master CI should pass.

HCastano:
Yeah Peg is right, this was fixed in `master` like two weeks ago 😄

HCastano:
@cooldracula looks like the GHA job failed with a timeout now. Are we using different runners for Circle CI and GHA? 

Ideally we don't just work around this by bumping the timeout to, say, 60 minutes

cooldracula:
> @cooldracula looks like the GHA job failed with a timeout now. Are we using different runners for Circle CI and GHA?
> 
> Ideally we don't just work around this by bumping the timeout to, say, 60 minutes

We are using different runners for the two services.  In Circle-Ci, we are using a self-hosted runner ([Linux X-Large](https://app.circleci.com/pipelines/github/entropyxyz/entropy-core/4939/workflows/b5fe191b-1a3c-4be2-bf01-a82a5248dcdf/jobs/18536)), while in this github action, we are using the [standard github hosted runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories).   If we wanted, we could use the custom runner we're currently using for the release builds for this workflow too?  

For the failed job, I think this was due to a timeout I should not have added to the step.  The original circle-ci job had the property [no_output_timeout: 45m](https://github.com/entropyxyz/entropy-core/blob/15794e6d82cfc935e3a2e4ea585e97868e49b73e/.circleci/then.yml#L127), so I added the most closely equivalent github property `timeout-minutes`.  However, github's timeout is based on the total execution time of the step, regardless of if the step was still producing output.  So the job running on smaller machine, with a hard timeout, would fail.  I removed the timeout property in my most recent commit (https://github.com/entropyxyz/entropy-core/pull/991/commits/b6a3eb2f77424f873526cbf0651d69c13b991e8c).

This job should complete successfully now, though it would be slower than on circle-ci.  If we want it to be faster, it might be good to use our larger self-hosted machine for it.

HCastano:
@cooldracula yeah ideally we keep runners with similar specs. 

I don't see the GHA build and test step for the tests runs though 🤔 I'd be curious to know how long that step takes with the default runners

cooldracula:
>  ideally we keep runners with similar specs.

I agree, @HCastano .  I updated the job to use our custom runner(`core-build-runner`), whose specs are roughly equivalent to our circle-ci runner. 

> I don't see the GHA build and test step for the tests runs though 🤔 I'd be curious to know how long that step takes with the default runners

This is the correct behaviour for this PR, as the job is set to only run if there was a change to files within the `crates`,`pallets`,`node`, or `runtime` directories. 

I [pushed a small edit to crates/README.md](https://github.com/entropyxyz/entropy-core/pull/991/commits/35f47d56ec2408b39d02f0a746a2cd7e2eb48154) to trigger the job.  It [completed successfully in 39 minutes](https://github.com/entropyxyz/entropy-core/actions/runs/10365576753/job/28692986738?pr=991).  This is about the same as circle-ci. In [a recent circle-ci run](https://app.circleci.com/pipelines/github/entropyxyz/entropy-core/4955/workflows/77fdee6e-40b4-4c80-8d87-ca6a8a3bcc2f/jobs/18616), the equivalent job completed in 40 minutes.

.github/workflows/build-and-run-node-test.yaml
```diff
@@ -0,0 +1,39 @@
+---
+name: "Build and test"
+on:
+  push:
+    paths:
+      - "node/**"
+      - "crates/**"
+      - "pallets/*"
+      - "runtime/*"
```
HCastano, line 9: Ah, okay I didn't read this carefully enough then. Nice that it doesn't do the full build for things like the `README`, that's something that annoyed me about the Circle CI config.

Let's change this to the following, since we have some nested folders in `pallets` and `runtime` where we'd want to run the CI if they changed (e.g `runtime/src/weights`):

```suggestion
name: "Build and test"
on:
  push:
    paths:
      - "node/**"
      - "crates/**"
      - "pallets/**"
      - "runtime/**"
```

PR #993, opened 2024-08-08
Fix TSS `AccountId` keys in chainspec

JesseAbram:
Keys were generated incorrectly with subkey when we needed to use hpke

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ At the moment this project **does not** adhere to
 - Add parent key threshold dynamically ([#974](https://github.com/entropyxyz/entropy-core/pull/974))
 
 ### Changed
-- Fix tss keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
+- Fix TSS `AccountId` keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
 
 ## [0.2.0](https://github.com/entropyxyz/entropy-core/compare/release/v0.1.0...release/v0.2.0) - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ At the moment this project **does not** adhere to
 - Set inital signers ([#971](https://github.com/entropyxyz/entropy-core/pull/971))
 - Add parent key threshold dynamically ([#974](https://github.com/entropyxyz/entropy-core/pull/974))
 
+### Changed
+- Fix tss keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
+
 ## [0.2.0](https://github.com/entropyxyz/entropy-core/compare/release/v0.1.0...release/v0.2.0) - 2024-07-11
 
 ### Breaking Changes
--- a/node/cli/src/chain_spec/mod.rs
+++ b/node/cli/src/chain_spec/mod.rs
@@ -75,15 +75,15 @@ pub mod tss_account_id {
     pub static ref CHARLIE: sp_runtime::AccountId32  =
             super::hex!["946140d3d5ddb980c74ffa1bb64353b5523d2d77cdf3dc617fd63de9d3b66338"].into();
 
-    /// Not sure what mnemonic is used to derive the following `AccountId`.
+     /// The `DEFAULT_DAVE_MNEMONIC` is used to derive the following `AccountId`.
     /// Mnemonic: "beef dutch panic monkey black glad audit twice humor gossip wealth drive"
     pub static ref DAVE: sp_runtime::AccountId32  =
-            super::hex!["12bf1c8e365c20cc2af606e3814b98b192857d85d182dac5e33fb90d0380ca75"].into();
+            super::hex!["0a9054ef6b6b8ad0dd2c89895b2515583f2fbf1edced68e7328ae456d86b9402"].into();
 
-    /// The `DEFAULT_CHARLIE_MNEMONIC` is used to derive the following `AccountId`.
+    /// The `DEFAULT_EVE_MNEMONIC` is used to derive the following `AccountId`.
     /// Mnemonic: "impact federal dish number fun crisp various wedding radio immense whisper glue"
     pub static ref EVE: sp_runtime::AccountId32  =
-            super::hex!["f2b4113735e988f662fe45e97b39770e804ebcd893ad0ab7cd8b7c5b5dcfff22"].into();
+            super::hex!["ac0d9030598f1722ff7c6a2a3043fa65903448dcc7a23011ec06c1c31cdad120"].into();
 
     }
 }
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ At the moment this project **does not** adhere to
 - Set inital signers ([#971](https://github.com/entropyxyz/entropy-core/pull/971))
 - Add parent key threshold dynamically ([#974](https://github.com/entropyxyz/entropy-core/pull/974))
 
+### Changed
+- Fix tss keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
```
HCastano, line 28:
```suggestion
- Fix TSS `AccountId` keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
```

node/cli/src/chain_spec/mod.rs
```diff
@@ -75,15 +75,15 @@ pub mod tss_account_id {
     pub static ref CHARLIE: sp_runtime::AccountId32  =
             super::hex!["946140d3d5ddb980c74ffa1bb64353b5523d2d77cdf3dc617fd63de9d3b66338"].into();
 
-    /// Not sure what mnemonic is used to derive the following `AccountId`.
+    /// The `DEFAULT_DAVE_MNEMONIC` is used to derive the following `AccountId`.
     /// Mnemonic: "beef dutch panic monkey black glad audit twice humor gossip wealth drive"
     pub static ref DAVE: sp_runtime::AccountId32  =
-            super::hex!["12bf1c8e365c20cc2af606e3814b98b192857d85d182dac5e33fb90d0380ca75"].into();
+            super::hex!["0a9054ef6b6b8ad0dd2c89895b2515583f2fbf1edced68e7328ae456d86b9402"].into();
```
HCastano, line 81: So I thought we were just supposed to be using `subkey` for these, not the HPKE stuff.

This is because an operator would be passing this in using the `--mnemonic` flag, which ideally they get by running `subkey generate`.

In this case then, the public key here should be `0x12bf1c8e365c20cc2af606e3814b98b192857d85d182dac5e33fb90d0380ca75` iiuc.

```
❮ ./target/release/entropy key inspect "beef dutch panic monkey black glad audit twice humor gossip wealth drive"
Secret phrase:       beef dutch panic monkey black glad audit twice humor gossip wealth drive
  Network ID:        substrate
  Secret seed:       0xb7bb3269a4a184529445a7fc91b03a87c17be3309cfa23adbcccb0f0fbc7f087
  Public key (hex):  0x12bf1c8e365c20cc2af606e3814b98b192857d85d182dac5e33fb90d0380ca75
  Account ID:        0x12bf1c8e365c20cc2af606e3814b98b192857d85d182dac5e33fb90d0380ca75
  Public key (SS58): 5CVHVSDm8643E1JWesWUjt7GZR6zMQa5oytg9TkBTMJ8T3cq
  SS58 Address:      5CVHVSDm8643E1JWesWUjt7GZR6zMQa5oytg9TkBTMJ8T3cq
```

node/cli/src/chain_spec/mod.rs
```diff
@@ -75,15 +75,15 @@ pub mod tss_account_id {
     pub static ref CHARLIE: sp_runtime::AccountId32  =
```
HCastano, line 75: How come Alice, Bob, and Charlie don't need to change?

node/cli/src/chain_spec/mod.rs
```diff
@@ -75,15 +75,15 @@ pub mod tss_account_id {
...
     pub static ref DAVE: sp_runtime::AccountId32  =
-            super::hex!["12bf1c8e365c20cc2af606e3814b98b192857d85d182dac5e33fb90d0380ca75"].into();
+            super::hex!["0a9054ef6b6b8ad0dd2c89895b2515583f2fbf1edced68e7328ae456d86b9402"].into();
 
-    /// The `DEFAULT_CHARLIE_MNEMONIC` is used to derive the following `AccountId`.
+    /// The `DEFAULT_EVE_MNEMONIC` is used to derive the following `AccountId`.
     /// Mnemonic: "impact federal dish number fun crisp various wedding radio immense whisper glue"
     pub static ref EVE: sp_runtime::AccountId32  =
-            super::hex!["f2b4113735e988f662fe45e97b39770e804ebcd893ad0ab7cd8b7c5b5dcfff22"].into();
+            super::hex!["ac0d9030598f1722ff7c6a2a3043fa65903448dcc7a23011ec06c1c31cdad120"].into();
```
HCastano, line 86: Similar thing here

```
❮ ./target/release/entropy key inspect "impact federal dish number fun crisp various wedding radio immense whisper glue"
Secret phrase:       impact federal dish number fun crisp various wedding radio immense whisper glue
  Network ID:        substrate
  Secret seed:       0x0b22c9142215fbb5de12d6f372e3a0819887945209c98405787a68c10a9ed4fb
  Public key (hex):  0xf2b4113735e988f662fe45e97b39770e804ebcd893ad0ab7cd8b7c5b5dcfff22
  Account ID:        0xf2b4113735e988f662fe45e97b39770e804ebcd893ad0ab7cd8b7c5b5dcfff22
  Public key (SS58): 5HYvwGKUxxCLWnhhGXbonLdgQ3wTLopUVyMvjAAo4o2j3XQi
  SS58 Address:      5HYvwGKUxxCLWnhhGXbonLdgQ3wTLopUVyMvjAAo4o2j3XQi
```

ameba23 (approved):
Im approving but i haven't actually checked the keys. Should be from this fn right?
https://github.com/entropyxyz/entropy-core/blob/d970930dae48f988045210a85a0260c98c8b17cd/crates/threshold-signature-server/src/helpers/validator.rs#L93

HCastano (commented):
I need a bit more clarity on where the accounts are coming from

node/cli/src/chain_spec/mod.rs
JesseAbram: so you can get the phrase and pass it in then on spin up the tss server will generate the keys from the seed and spit it out in the .production file under account_id

JesseAbram:
to answer both comments, yes I got these keys from actually running the TSS server and grabbing the output of what is generated

node/cli/src/chain_spec/mod.rs
ameba23: The auditors didn't like us deriving the encryption keys from signing keys, so now both signing keypair and encryption keys are derived from the mnemonic, using a different derivation process to that of subkey.

HCastano: Okay maybe I need to follow up with that issue again 🤔

JesseAbram: that is cool gonna merge in the meantime since it is blocking

PR #994, opened 2024-08-08
Refactor reshare

JesseAbram:
Related #941 

Refactors reshare to share code between proactive refresh and reshare, mainly by pulling key reshare inputs out to a hire level and passed through

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,6 @@ At the moment this project **does not** adhere to
 
 ### Changed
 - Fix TSS `AccountId` keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
-- Refactor reshare ([#994](https://github.com/entropyxyz/entropy-core/pull/994))
 
 ## [0.2.0](https://github.com/entropyxyz/entropy-core/compare/release/v0.1.0...release/v0.2.0) - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ At the moment this project **does not** adhere to
 
 ### Changed
 - Fix TSS `AccountId` keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
+- Refactor reshare ([#994](https://github.com/entropyxyz/entropy-core/pull/994))
 
 ## [0.2.0](https://github.com/entropyxyz/entropy-core/compare/release/v0.1.0...release/v0.2.0) - 2024-07-11
 
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -333,7 +333,6 @@ pub async fn execute_reshare(
     session_id: SessionId,
     chans: Channels,
     threshold_pair: &sr25519::Pair,
-    threshold_accounts: Vec<AccountId32>,
     inputs: KeyResharingInputs<KeyParams, PartyId>,
     aux_info_option: Option<AuxInfo<KeyParams, PartyId>>,
 ) -> Result<
@@ -343,8 +342,6 @@ pub async fn execute_reshare(
     tracing::debug!("Executing proactive refresh");
     tracing::debug!("Signing with {:?}", &threshold_pair.public());
 
-    let party_ids: BTreeSet<PartyId> =
-        threshold_accounts.iter().cloned().map(PartyId::new).collect();
     let pair = PairWrapper(threshold_pair.clone());
 
     let session_id_hash = session_id.blake2(None)?;
@@ -353,13 +350,12 @@ pub async fn execute_reshare(
         &mut OsRng,
         SynedrionSessionId::from_seed(session_id_hash.as_slice()),
         pair,
-        &party_ids,
-        inputs,
+        &inputs.new_holders,
+        inputs.clone(),
     )
     .map_err(ProtocolExecutionErr::SessionCreation)?;
 
     let (new_key_share, chans) = execute_protocol_generic(chans, session, session_id_hash).await?;
-
     let aux_info = if let Some(aux_info) = aux_info_option {
         aux_info
     } else {
@@ -369,11 +365,11 @@ pub async fn execute_reshare(
             &mut OsRng,
             SynedrionSessionId::from_seed(session_id_hash_aux_data.as_slice()),
             PairWrapper(threshold_pair.clone()),
-            &party_ids,
+            &inputs.new_holders,
         )
         .map_err(ProtocolExecutionErr::SessionCreation)?;
 
-        execute_protocol_generic(chans, session, session_id_hash).await?.0
+        execute_protocol_generic(chans, session, session_id_hash_aux_data).await?.0
     };
 
     Ok((new_key_share.ok_or(ProtocolExecutionErr::NoOutputFromReshareProtocol)?, aux_info))
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -73,8 +73,7 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
     mut chans: Channels,
     session: Session<Res, sr25519::Signature, PairWrapper, PartyId>,
     session_id_hash: [u8; 32],
-) -> Result<(Res::Success, Broadcaster, mpsc::Receiver<ProtocolMessage>), GenericProtocolError<Res>>
-{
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ At the moment this project **does not** adhere to
 
 ### Changed
 - Fix TSS `AccountId` keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
+- Refactor reshare ([#994](https://github.com/entropyxyz/entropy-core/pull/994))
```
HCastano, line 29:
```suggestion
```

Internal change

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -166,54 +165,34 @@ pub async fn new_reshare(
...
-    let (new_key_share_option, rx) = execute_protocol_generic(channels, session, session_id_hash)
-        .await
-        .map_err(|_| ValidatorErr::ProtocolError("Error executing protocol".to_string()))?;
-
-    let new_key_share = new_key_share_option.ok_or(ValidatorErr::NoOutputFromReshareProtocol)?;
+    .await?;
 
-    // Setup channels for the next session
-    let channels = Channels(broadcaster, rx);
+    // // Setup channels for the next session
```
HCastano, line 187:
```suggestion
    // Setup channels for the next session
```

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -166,54 +165,34 @@ pub async fn new_reshare(
...
+    let channels = Channels(brodcaster, rx);
 
-    // Now run an aux gen session
+    // // Now run an aux gen session
```
HCastano, line 190:
```suggestion
    // Now run an aux gen session
```

crates/protocol/src/execute_protocol.rs
```diff
@@ -73,7 +73,8 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
     mut chans: Channels,
     session: Session<Res, sr25519::Signature, PairWrapper, PartyId>,
     session_id_hash: [u8; 32],
-) -> Result<(Res::Success, mpsc::Receiver<ProtocolMessage>), GenericProtocolError<Res>> {
+) -> Result<(Res::Success, Broadcaster, mpsc::Receiver<ProtocolMessage>), GenericProtocolError<Res>>
```
HCastano, line 76: What's the advantage of passing the `Broadcaster` back to the caller?

crates/threshold-signature-server/src/signing_client/api.rs
```diff
@@ -102,7 +103,7 @@ pub async fn proactive_refresh(
             ) = deserialize(&old_key_share)
                 .ok_or_else(|| ProtocolErr::Deserialization("Failed to load KeyShare".into()))?;
 
-            let new_key_share = do_proactive_refresh(
+            let (new_key_share, _broadcaster, _rx) = do_proactive_refresh(
```
HCastano, line 106: If we ignore both of the new return values on the only invocation of this function, are they worth returning back?

crates/threshold-signature-server/src/signing_client/api.rs
```diff
@@ -152,6 +153,7 @@ async fn handle_socket_result(socket: WebSocket, app_state: AppState) {
     };
 }
 
+#[allow(clippy::type_complexity)]
```
HCastano, line 156: Please avoid ignoring lints if possible. We can either change the return values to something more manageable (e.g do we need to pass the channel info back?), or we can typedef this and have a single type here

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -632,7 +632,7 @@ async fn test_store_share() {
 
     let mut new_verifying_key = vec![];
     // wait for registered event check that key exists in kvdb
-    for _ in 0..45 {
```
HCastano, line 635: Any idea what changes from the PR caused this to require an increase?

HCastano (commented):
So tbh I don't really understand the advantages gained by these changes. Maybe you can elaborate what problem this is solving for you?

Otherwise I'll defer to Peg here since he's more familiar with the protocol loop.

JesseAbram:
> So tbh I don't really understand the advantages gained by these changes. Maybe you can elaborate what problem this is solving for you?
> 
> Otherwise I'll defer to Peg here since he's more familiar with the protocol loop.

ya for sure so proactive refresh and reshare are kinda the same code in the execution step, but they take different inputs, in reshare we have n - 1 nodes with keyshares and 1 new joiner, in reshare it is n nodes, originally the input generation happened lower down in the stack so all the code up to that point needed to be duplicated, by moving it up the stack a lot of code can now be shared, that said probably gonna remove reshare eventually but who knows

crates/protocol/src/execute_protocol.rs
JesseAbram: So in a reshare it is a two step process (also in jumpstart) where you do the reshare then we reshare the aux data, by returning brodcaster we can use the same channel that was already setup instead of creating a new channel

crates/threshold-signature-server/src/signing_client/api.rs
JesseAbram: we ignore it here but to make this function generalizable to reshare too we can return these

crates/threshold-signature-server/src/user/tests.rs
JesseAbram: mmm no not sure why, I think it was just a random race condition, sometimes blocks take longer since we are only running one validator and so bob gets assigned slots, based on chain randomness

crates/threshold-signature-server/src/signing_client/api.rs
JesseAbram: idk these rules are pretty arbitrary and IMO here it would decrease readability and cleanliness

crates/protocol/src/execute_protocol.rs
ameba23: We don't need to return the broadcaster, only the `mpsc::Receiver`, as we can clone the broadcaster before passing it in (which is what we were doing before).

If you want to return the broadcaster anyway for convenience, for clarity i would return the `Channels` struct, which is just a tuple of broadcaster and mpsc receiver:
https://github.com/entropyxyz/entropy-core/blob/15794e6d82cfc935e3a2e4ea585e97868e49b73e/crates/protocol/src/execute_protocol.rs#L48 

This would probably remove the need for allowing `type_complexity`.

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -166,54 +165,34 @@ pub async fn new_reshare(
...
-    let new_key_share = new_key_share_option.ok_or(ValidatorErr::NoOutputFromReshareProtocol)?;
+    .await?;
 
     // Setup channels for the next session
-    let channels = Channels(broadcaster, rx);
+    let channels = Channels(brodcaster, rx);
 
     // Now run an aux gen session
     let session_id_hash = session_id.blake2(Some(Subsession::AuxGen))?;
     let session = make_aux_gen_session(
```
ameba23, line 192: Just an opinion but i would keep anything which deals directly with synedrion sessions inside entropy-protocol's `execute_protocol.rs`. So either move this into `execute_proactive_refresh` inside an `if` checking if the parties have changed. Or have separate execute refresh / execute reshare functions.

crates/threshold-signature-server/src/signing_client/api.rs
```diff
@@ -273,3 +279,34 @@ pub async fn validate_proactive_refresh(
     kv_manager.kv().put(reservation, latest_block_number.to_be_bytes().to_vec()).await?;
     Ok(())
 }
+
+pub async fn get_channels(
```
ameba23, line 283: This is good, probably we can also use it for the signing / DKG sessions

crates/protocol/src/execute_protocol.rs
```diff
@@ -337,29 +340,20 @@ pub async fn execute_proactive_refresh(
     chans: Channels,
```
ameba23, line 340: Just an opinion but if using this function for both reshare and refresh i would name it `execute_reshare`.

ameba23 (approved):
:+1:
This all works.  I would have done things a little bit differently, but its purely aesthetics with no right or wrong answer.

crates/threshold-signature-server/src/signing_client/api.rs
JesseAbram: ya I was thinking the same but gonna wait till active development on them is done

PR #997, opened 2024-08-09
Add new registration flow test to TSS side

HCastano:
While working on the signing flow I realized that there wasn't a test checking the new
registration flow on the TSS side (which is a pre-requisite to doing signing).

I've added a mostly happy-path test for registration. I don't really want to get bogged
down in checking all the failure scenarios until after the signing flow is done.

### DIFF
--- a/crates/threshold-signature-server/src/user/tests.rs
+++ b/crates/threshold-signature-server/src/user/tests.rs
@@ -1471,3 +1471,25 @@ pub async fn get_sign_tx_data(
 
     (validators_info, generic_msg, validator_ips_and_keys)
 }
+
+pub async fn jump_start_network(
+    api: &OnlineClient<EntropyConfig>,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+) {
+    let alice = AccountKeyring::Alice;
+    let signer = PairSigner::<EntropyConfig, sr25519::Pair>::new(alice.clone().into());
+
+    let jump_start_request = entropy::tx().registry().jump_start_network();
+    let _result = submit_transaction(api, rpc, &signer, &jump_start_request, None).await.unwrap();
+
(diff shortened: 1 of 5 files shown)


crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -828,6 +829,32 @@ pub async fn put_register_request_on_chain(
...
+        PairSigner::<EntropyConfig, sp_core::sr25519::Pair>::new(signature_request_account.pair());
+
+    let registering_tx =
+        entropy::tx().registry().register_on_chain(program_modification_account, program_instance);
+
+    let events =
+        submit_transaction(api, rpc, &signature_request_account, &registering_tx, None).await?;
+
+    // Since we're only submitting one request above, looking for the first event as opposed to
+    // say, all events, should be fine.
```
ameba23, line 851: This is nice.  But in our actual production client code we will still need a loop to filter out the events of other users, right?

ameba23 (approved):
:100: nice clear comments i gotta say

JesseAbram (approved):

crates/threshold-signature-server/src/user/tests.rs
JesseAbram: yes that is correct

pallets/registry/src/lib.rs
```diff
@@ -143,15 +143,6 @@ pub mod pallet {
                         version_number: T::KeyVersionNumber::get(),
                     },
                 );
-
-                RegisteredOnChain::<T>::insert(
```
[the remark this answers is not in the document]
HCastano: For the moment no. I wrote a little thing out in the commit message removing this why it doesn't make sense to have them. tl;dr registration requires a jumpstarted network as a pre-req and that's a bit of a tricky setup using the Genesis config code.

It's something we might want to look at in the future if we find that manually setting everything up with extrinsics becomes cumbersome

PR #999, opened 2024-08-10
Delete old keyshare if not in next_signers

JesseAbram:
Related #941 

- [x] Delete old key if not a new signer (includes making the test for reshare in TSS actually rotate the signers not just do a reshare)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,6 @@ At the moment this project **does not** adhere to
 - Reshare confirmation ([#965](https://github.com/entropyxyz/entropy-core/pull/965))
 - Set inital signers ([#971](https://github.com/entropyxyz/entropy-core/pull/971))
 - Add parent key threshold dynamically ([#974](https://github.com/entropyxyz/entropy-core/pull/974))
-- Delete old keyshare if not in next_signers ([#999](https://github.com/entropyxyz/entropy-core/pull/999))
 - TSS attestation endpoint ([#1001](https://github.com/entropyxyz/entropy-core/pull/1001))
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ At the moment this project **does not** adhere to
 - Reshare confirmation ([#965](https://github.com/entropyxyz/entropy-core/pull/965))
 - Set inital signers ([#971](https://github.com/entropyxyz/entropy-core/pull/971))
 - Add parent key threshold dynamically ([#974](https://github.com/entropyxyz/entropy-core/pull/974))
+- Delete old keyshare if not in next_signers ([#999](https://github.com/entropyxyz/entropy-core/pull/999))
 
 ### Changed
 - Fix TSS `AccountId` keys in chainspec ([#993](https://github.com/entropyxyz/entropy-core/pull/993))
--- a/crates/threshold-signature-server/src/validator/api.rs
+++ b/crates/threshold-signature-server/src/validator/api.rs
@@ -99,8 +99,12 @@ pub async fn new_reshare(
     )
     .map_err(|e| ValidatorErr::VerifyingKeyError(e.to_string()))?;
 
-    let is_proper_signer =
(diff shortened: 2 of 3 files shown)


crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -360,3 +359,22 @@ pub async fn prune_old_holders(
...
+) -> Result<bool, ValidatorErr> {
+    let is_proper_signer =
+        &validators_info.iter().any(|validator_info| validator_info.tss_account == *account_id);
+    if *is_proper_signer {
+        Ok(true)
+    } else {
+        // delete old keyshare if has it and not next_signer
+        if kv_manager.kv().exists(&hex::encode(NETWORK_PARENT_KEY)).await? {
+            kv_manager.kv().delete(&hex::encode(NETWORK_PARENT_KEY)).await?
+        }
```
HCastano, line 377: I would say it's misleading to have an `is_()` function which has a side effect - and especially in this case where it's deleting a key from the DB.

Since this function is only called once, I would just inline it at the call site so it's at least transparent to any reader that there are side effects in the case that the validator isn't a "proper" signer.

JesseAbram: that was my first instinct however since this is a simple test for functionality, I moved it out for testing. 

To move it in I would have to rewrite the state of the mock_signers (breaking an earlier test......or put it in the dev chain init state.......or I could manually swap out the TSS key by accessing the kvdb directly) 

which I could do, but it would like bloat the test code plus require a chain spin up and at least add 60 seconds to the test which seemed a lot worse 

I could change the function name to account for the secondary effects tho like ```is_signer_or_delete_parent_key```

HCastano: Even in the function name you're proposal it seems like you're trying to couple two different concepts. I do see what you mean about the ease of testing though.

I would try and think about a few other options here, but I'll approve to unblock you.

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -360,3 +359,22 @@ pub async fn prune_old_holders(
...
+pub async fn is_proper_signer(
+    account_id: &AccountId32,
+    validators_info: Vec<ValidatorInfo>,
+    kv_manager: &KvManager,
+) -> Result<bool, ValidatorErr> {
+    let is_proper_signer =
+        &validators_info.iter().any(|validator_info| validator_info.tss_account == *account_id);
+    if *is_proper_signer {
+        Ok(true)
+    } else {
```
HCastano, line 373:
```suggestion
    let is_proper_signer =
        validators_info.iter().any(|validator_info| validator_info.tss_account == *account_id);
    if is_proper_signer {
        Ok(true)
    } else {
```

Don't need the reference here

crates/threshold-signature-server/src/validator/api.rs
HCastano, line 377:
```suggestion
        // delete old keyshare if has it and not next_signer
        let network_key = hex::encode(NETWORK_PARENT_KEY);
        if kv_manager.kv().exists(&network_key).await? {
            kv_manager.kv().delete(&network_key).await?
        }
```

HCastano (approved):

CHANGELOG.md
```diff
@@ -23,7 +23,8 @@ At the moment this project **does not** adhere to
 - Reshare confirmation ([#965](https://github.com/entropyxyz/entropy-core/pull/965))
 - Set inital signers ([#971](https://github.com/entropyxyz/entropy-core/pull/971))
 - Add parent key threshold dynamically ([#974](https://github.com/entropyxyz/entropy-core/pull/974))
-- TSS attestation endpoint ([#1001](https://github.com/entropyxyz/entropy-core/pull/1001)
+- Delete old keyshare if not in next_signers ([#999](https://github.com/entropyxyz/entropy-core/pull/999))
```
HCastano, line 26:
```suggestion
```

This is an internal change, we don't need it here

Issue #982, opened 2024-08-05
DCAP attestation feature design


ameba23:
As part of our `tofino` design, the entropy threshold signature server is going to be run in a trusted execution environment.  This is a proposed feature for attestation that validators have correctly set up a trusted domain and are running our release VM image with the current release of the `entropy-tss` binary, 

This comes from my notes which were originally going to be incorporated into the [tofino-spec](https://github.com/entropyxyz/tofino-spec). But since it has become a bit implementation-oriented i am putting it here as a feature design issue. Once we are happy with it i'll make a PR to the spec repo with a more high-level description.

## `pallet-attestation`

The attestation pallet will be responsible for requesting attestations via the propagation pallet, and verifying the resulting quotes submitted by TS servers using the `attest` extrinsic. 

Quote verification will involve:
1. Verifying the ECDSA P-256 signature.
2. Checking the attestation key with the associated PCK (Provisioning Certification Key) certificate. The PCK needs to be submitted by the validator when they join, and will be handled by the staking extension pallet.
3. Checking the build-time measurement register (MRTD) value matches the current release of the threshold server virtual machine image. I think this value should be hard-coded into the chain runtime and so that it can only be updated with a runtime update. 
4. Checking the `REPORTDATA` for the block number associated with the attestation request and additional details [described below](##secure-channel-establishment). In the case that this data structure exceeds 64 bytes it can be hashed, and the hashes compared. This data structure will need to live in `entropy-shared` as is will be needed by both the attestation pallet and the threshold server. 

Since quote verification happens in a pallet, we probably cant use one of intel's crates with bindings to the C quote verification library (eg: [`intel-tee-quote-verification-sys`](https://crates.io/crates/intel-tee-quote-verification-sys)). This means we will probably have to use a pure rust library. There exists [`sgx-quote`](https://crates.io/crates/sgx-quote) which i think has only minor differences to the TDX quote format. The quote format is very simple and is well documented in appendix 3 of the [Intel TDX DCAP Quoting Library API reference](https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_TDX_DCAP_Quoting_Library_API.pdf).

As for when or how often to do on-chain attestation - i am not sure.  But it should definitely be triggered when a new validator initially calls the staking extension's `validate` extrinsic, and the account ID and x25519 public key of the new validator should not be stored until the quote is successfully verified.  At the point of network jump start would also be a good idea.

## The threshold server `/attest` endpoint

[...]

PR #1001, opened 2024-08-12
TSS attestation endpoint

ameba23:
This is part of https://github.com/entropyxyz/entropy-core/issues/982

This adds a HTTP endpoint `/attest` to the TS server.

With the `unsafe` feature enabled, this will create a mock TDX quote for testing the attestation logic on non-TDX hardware.

Without the `unsafe` feature enabled, this will currently just respond with an error. Eventually though, it will create an actual TDX quote.

The idea is that once there is a pallet handling attestation, this endpoint will be called by an offchain worker.  The endpoint will create the quote and submit it to the chain in an extrinsic. That does not yet happen, as we don't yet have a pallet to handle this.


crates/threshold-signature-server/src/lib.rs
```diff
@@ -155,7 +155,7 @@ use crate::{
     r#unsafe::api::{delete, put, remove_keys, unsafe_get},
     signing_client::{api::*, ListenerState},
     user::api::*,
-    validator::api::new_reshare,
+    validator::api::{attest, new_reshare},
```
ameba23, line 158: I'm not sure if `validator::api` is the right home for this.  Open to suggestions.  Could make an `attestation` module.

crates/shared/src/types.rs
```diff
@@ -99,3 +100,22 @@ pub enum HashingAlgorithm {
 
 /// A compressed, serialized [synedrion::ecdsa::VerifyingKey<k256::Secp256k1>]
 pub type EncodedVerifyingKey = [u8; VERIFICATION_KEY_LENGTH as usize];
+
+/// Input data to be included in a TDX attestation
+pub struct QuoteInputData(pub [u8; 64]);
```
ameba23, line 105: This is in `entropy-shared` because it will also be used by the chain when verifying quote input

crates/threshold-signature-server/src/lib.rs
JesseAbram: ya I think an attestation module would be cleaner

crates/threshold-signature-server/src/lib.rs
```diff
@@ -178,6 +178,7 @@ pub fn app(app_state: AppState) -> Router {
         .route("/user/sign_tx", post(sign_tx))
         .route("/signer/proactive_refresh", post(proactive_refresh))
         .route("/validator/reshare", post(new_reshare))
+        .route("/attest", post(attest))
```
JesseAbram, line 181: this should go lower into the unsafe endpoints

JesseAbram (approved):
just the two things but looks good

crates/threshold-signature-server/src/lib.rs
ameba23: i can put it there for now but the idea is this same endpoint will be the real thing when `unsafe` feature is not present

JesseAbram: idk we can always change it when we remove the unsafe, it makes sense to me but your call

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -360,3 +360,47 @@ pub async fn prune_old_holders(
...
+    // Non-mock attestation (the real thing) will go here
+    Err(ValidatorErr::NotImplemented)
+}
+
+#[cfg(any(test, feature = "unsafe"))]
+pub async fn attest(
+    State(app_state): State<AppState>,
+    input: Bytes,
+) -> Result<(StatusCode, Bytes), ValidatorErr> {
+    // TODO #982 confirm with the chain that an attestation should be happenning
```
HCastano, line 378:
```suggestion
    // TODO (#982): confirm with the chain that an attestation should be happening
```

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -360,3 +360,47 @@ pub async fn prune_old_holders(
...
+    let nonce = input[..].try_into()?;
```
HCastano, line 379:
```suggestion
    let nonce = input.as_slice().try_into()?;
```

crates/threshold-signature-server/Cargo.toml
```diff
@@ -83,6 +84,7 @@ ethers-core ="2.0.14"
 schnorrkel  ={ version="0.11.4", default-features=false, features=["std"] }
 schemars    ={ version="0.8.21" }
 subxt-signer="0.35.3"
+tdx-quote   ={ git="https://github.com/entropyxyz/tdx-quote", features=["mock"] }
```
HCastano, line 87: Do we need to have this in both norm deps and dev deps?

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -360,3 +360,47 @@ pub async fn prune_old_holders(
...
+        signer.signer().public().into(),
+        *public_key.as_bytes(),
+        nonce,
+        block_number,
+    );
+
+    let quote = tdx_quote::Quote::mock(signing_key.clone(), input_data.0);
+    // Here we would submit an attest extrinsic to the chain - but for now we just include it in the
+    // response
+    Ok((StatusCode::OK, Bytes::from(quote.as_bytes().to_vec())))
```
HCastano, line 405: Just a small thing for the future - would be nice to have a `From/Into` implementation for turning these quotes into `Bytes`

crates/threshold-signature-server/src/lib.rs
HCastano: FWIW I think it's fine as is

HCastano (approved):
Nice stuff 👍

crates/threshold-signature-server/Cargo.toml
ameba23: the problem is if we don't, it wont compile in test mode without the `unsafe` feature present. and ideally i don't want tdx-quote with the `mock` feature available in production.

crates/threshold-signature-server/src/validator/api.rs
ameba23: I think thats only worth doing if we continue to include the quote in the http response. Initially i am proposing to only have on-chain quote verification, so the response body here will be empty and the quote will be included in an extrinsic instead. I want to keep the dependencies of `tdx-quote` as minimal as possible as its used in a pallet.

But if we decide to do quote verification by other TS servers, then we will want this.

crates/threshold-signature-server/src/validator/api.rs
ameba23: i went for `as_ref` as `as_bytes` seems to be only available on nightly:
https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.as_bytes

PR #1003, opened 2024-08-13
Attestation pallet

ameba23:
This is part of https://github.com/entropyxyz/entropy-core/issues/982

This adds a pallet for TDX attestation.

The pallet stores the nonces of all pending (requested) attestations, storing them under associated TSS account ID. So there may be at most one pending attestation per TS server. The nonce is just a random 32 bytes, which is included in the input data to the TDX quote, to prove that this is a freshly made quote.

An attestation request is responded to by submitting the quote using the `attest` extrinsic. If there was a pending attestation for the caller, the quote is verified.  Verification currently just means checking that the quote parses correctly and has a valid signature. But this would eventually also check the build-time or run-time measurement details match our current release and that the public key matches the corresponding PCK certificate. PCK certificates will be handled in a separate PR.

If the quote fails to verify, something should happen to the validator - eg: remove them from the signing committee or block them from joining.  This is outside of the scope of this PR and can be handled as part of the slashing feature.

The attestation pallet also stores a mapping of block number to TSS account IDs of nodes for who an attestation request should be initiated. This is used by the propagation pallet to make a POST request to the TS server's `/attest` endpoint whenever there are requests to be made.  Currently, the only place these attestation requests are initiated is in the genesis config for testing.


crates/shared/src/types.rs
```diff
@@ -105,14 +114,14 @@ pub type EncodedVerifyingKey = [u8; VERIFICATION_KEY_LENGTH as usize];
 pub struct QuoteInputData(pub [u8; 64]);
 
 impl QuoteInputData {
-    pub fn new(
-        tss_account_id: [u8; 32],
+    pub fn new<T: Encode>(
```
ameba23, line 117: This is just so it can be used with mock account ids in the attestation pallet test.

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -13,57 +13,98 @@
...
 use axum::{body::Bytes, extract::State, http::StatusCode};
+use entropy_shared::OcwMessageAttestationRequest;
+use parity_scale_codec::Decode;
+use sp_core::Pair;
+use subxt::tx::PairSigner;
+use x25519_dalek::StaticSecret;
 
 /// HTTP POST endpoint to initiate a TDX attestation.
-/// Not yet implemented.
-#[cfg(not(any(test, feature = "unsafe")))]
```
ameba23, line 21: Before we had a production and mock version of the endpoint function - but that is now replaced with a common endpoint function which internally calls either a mock or production version of the `create_quote` function

crates/threshold-signature-server/src/attestation/tests.rs
```diff
@@ -27,25 +31,34 @@ async fn test_attest() {
...
-    // Check the input data of the quote
-    let expected_input_data =
-        QuoteInputData::new(TSS_ACCOUNTS[0].0, X25519_PUBLIC_KEYS[0], nonce, 0);
-    assert_eq!(quote.report_input_data(), expected_input_data.0);
+        // There should be no more pending attestation as the attestation has been handled
+        let pending_attestation_query =
+            entropy::storage().attestation().pending_attestations(&TSS_ACCOUNTS[0]);
+        if query_chain(&api, &rpc, pending_attestation_query, None).await.unwrap().is_none() {
+            return;
+        }
```
ameba23, line 61: This test isn't ideal - but currently the only outcome of a successful attestation is that there is no longer a pending attestation.

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -13,57 +13,98 @@
...
 /// [Index TDX DCAP Quoting Library API](https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_TDX_DCAP_Quoting_Library_API.pdf).
-#[cfg(any(test, feature = "unsafe"))]
 pub async fn attest(
     State(app_state): State<AppState>,
     input: Bytes,
-) -> Result<(StatusCode, Bytes), AttestationErr> {
-    use crate::{chain_api::get_rpc, get_signer_and_x25519_secret};
-    use rand_core::OsRng;
-    use sp_core::Pair;
+) -> Result<StatusCode, AttestationErr> {
```
JesseAbram, line 38: we will have to validate this message agains blockchain data (can be in a later PR but needs an issue

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -13,57 +13,98 @@
...
+    let api = get_api(&app_state.configuration.endpoint).await?;
     let rpc = get_rpc(&app_state.configuration.endpoint).await?;
 
+    // Get the input nonce for this attestation
+    let nonce = {
+        let pending_attestation_query =
+            entropy::storage().attestation().pending_attestations(signer.account_id());
+        query_chain(&api, &rpc, pending_attestation_query, None)
+            .await?
+            .ok_or_else(|| AttestationErr::Unexpected)?
```
JesseAbram, line 56: wouldn't hate a better error message here, a generalize one for querying chains maybe that takes a string of a better description

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    #[pallet::storage]
+    #[pallet::getter(fn pending_attestations)]
+    pub type PendingAttestations<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::AccountId, [u8; 32], OptionQuery>;
+
+    /// A mapping between block numbers and TSS nodes for who we want to make a request for
+    /// attestation, used to make attestation requests via an offchain worker
+    #[pallet::storage]
+    #[pallet::getter(fn attestation_requests)]
+    pub type AttestationRequests<T: Config> =
```
JesseAbram, line 83: idea here, we can make a trailing request of X blocks on an on init hook that boots a validator if they did not complete their attestation

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+            // Check report input data matches the nonce, TSS details and block number
+            let expected_input_data =
+                QuoteInputData::new(&who, x25519_public_key, nonce, block_number);
+            ensure!(
+                quote.report_input_data() == expected_input_data.0,
+                Error::<T>::IncorrectInputData
+            );
+
+            // Remove the entry from PendingAttestations
+            PendingAttestations::<T>::remove(&who);
```
JesseAbram, line 140: this should be below the other two probably

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+
```
JesseAbram, line 141: Idea: we put in the validator info when their last successful attestation was made to chain

JesseAbram (approved):
Looks good a few issues need to be made

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+//! # Attestation Pallet
```
HCastano, line 16: Some docs about what this is would be nice

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+        /// Describes the weights of the dispatchables exposed by this pallet.
+        type WeightInfo: WeightInfo;
+    }
+
+    #[pallet::genesis_config]
+    #[derive(frame_support::DefaultNoBound)]
+    pub struct GenesisConfig<T: Config> {
+        pub initial_pending_attestations: Vec<(T::AccountId, [u8; 32])>,
+        pub initial_attestation_requests: Vec<(BlockNumberFor<T>, Vec<Vec<u8>>)>,
```
HCastano, line 58:
```suggestion
        pub initial_attestation_requests: Vec<(BlockNumberFor<T>, Vec<T::AccountId>)>,
```

Why not do something like this instead?

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+                AttestationRequests::<T>::insert(block_number, account_ids);
+            }
+        }
+    }
+
+    /// A map of TSS account id to quote nonce for pending attestations
+    #[pallet::storage]
+    #[pallet::getter(fn pending_attestations)]
+    pub type PendingAttestations<T: Config> =
+        StorageMap<_, Blake2_128Concat, T::AccountId, [u8; 32], OptionQuery>;
```
HCastano, line 77: We could typedef this as `type Nonce = [u8; 32]`

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        AttestationMade,
+    }
+
+    #[pallet::error]
+    pub enum Error<T> {
+        BadQuote,
+        UnexpectedAttestation,
+        IncorrectInputData,
+        NoStashAccount,
+        NoServerInfo,
```
HCastano, line 98: I know we don't do it for the other pallets, but let's try and add doc comments before it's too late lol

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight({100})]
+        pub fn attest(origin: OriginFor<T>, quote: Vec<u8>) -> DispatchResult {
```
HCastano, line 105: Missing doc comments. Would be helpful to know what a `quote` should look like for example

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        UnexpectedAttestation,
+        IncorrectInputData,
+        NoStashAccount,
+        NoServerInfo,
+    }
+
+    #[pallet::call]
+    impl<T: Config> Pallet<T> {
+        #[pallet::call_index(0)]
+        #[pallet::weight({100})]
```
HCastano, line 104: Should be using the benchmarked weights

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        pub fn attest(origin: OriginFor<T>, quote: Vec<u8>) -> DispatchResult {
+            let who = ensure_signed(origin)?;
```
HCastano, line 106:
```suggestion
            let who = ensure_signed(origin)?;

```

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+
+            // TODO Check measurements match current release of entropy-tss
+            let _mrtd = quote.mrtd();
+
+            // TODO Check that the attestation public key matches that from PCK certificate
+            let _attestation_key = quote.attestation_key;
+
+            // TODO If anything fails, don't just return an error - do something mean
+
+            Self::deposit_event(Event::AttestationMade);
```
HCastano, line 150:
```suggestion
            Self::deposit_event(Event::AttestationMade);

```

pallets/attestation/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+                QuoteInputData::new(&who, x25519_public_key, nonce, block_number);
+            ensure!(
+                quote.report_input_data() == expected_input_data.0,
+                Error::<T>::IncorrectInputData
+            );
+
+            // Remove the entry from PendingAttestations
+            PendingAttestations::<T>::remove(&who);
+
+            // TODO Check measurements match current release of entropy-tss
```
HCastano, line 142: I would tie these `TODO` to the main attestation issue you have open

HCastano: Also, maybe a topic for discussion elsewhere but I would want to have bit of wiggle room for the release version. It makes sense to be that there could be slightly different yet compatible versions of the TSS, e.g `X.Y.0` vs `X.Y.1`

pallets/attestation/src/benchmarking.rs
```diff
@@ -0,0 +1,76 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+// This is a randomly generated secret p256 ECDSA key
+const ENCLAVE_SIGNING_KEY: [u8; 32] = [
+    167, 184, 203, 130, 240, 249, 191, 129, 206, 9, 200, 29, 99, 197, 64, 81, 135, 166, 59, 73, 31,
+    27, 206, 207, 69, 248, 56, 195, 64, 92, 109, 46,
+];
```
HCastano, line 20: nit: I'd put this after the imports

pallets/attestation/src/benchmarking.rs
```diff
@@ -0,0 +1,76 @@
...
+        tss_account: attestee.clone(),
+        x25519_public_key: [0; 32],
+        endpoint: b"http://localhost:3001".to_vec(),
+    });
+
+  }: _(RawOrigin::Signed(attestee.clone()), quote.clone())
+  verify {
+    assert_last_event::<T>(
+        Event::<T>::AttestationMade.into()
+    );
```
HCastano, line 72: Here we can also test for expected behaviour. 

For example, we can check that `PendingAttestations` doesn't have the validator ID anymore, or that expected events were emitted

pallets/propagation/src/lib.rs
```diff
@@ -304,5 +310,49 @@ pub mod pallet {
...
+            if let Some(attestations_to_request) =
+                pallet_attestation::Pallet::<T>::attestation_requests(block_number)
+            {
+                if attestations_to_request.is_empty() {
+                    return Ok(());
+                }
+
+                let deadline = sp_io::offchain::timestamp().add(Duration::from_millis(2_000));
+                let kind = sp_core::offchain::StorageKind::PERSISTENT;
+                let from_local = sp_io::offchain::local_storage_get(kind, b"attest")
```
HCastano, line 327: I think we also need to reflect this in the [`service`](https://github.com/entropyxyz/entropy-core/blob/0abb8bea4f5e1deb18da2e3af9c3b45e90137ac5/node/cli/src/service.rs#L370)

node/cli/src/chain_spec/dev.rs
```diff
@@ -336,5 +336,9 @@ pub fn development_genesis_config(
                 10,
             )],
         },
+        "attestation": AttestationConfig {
+            initial_attestation_requests: vec![(3, vec![crate::chain_spec::tss_account_id::ALICE.to_raw_vec()])],
```
HCastano, line 340: Do you mind using an account instead of `3` here? Either a typedef or the result from one of the other helpful functions is fine

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -13,57 +13,98 @@
...
 pub async fn attest(
     State(app_state): State<AppState>,
     input: Bytes,
-) -> Result<(StatusCode, Bytes), AttestationErr> {
-    use crate::{chain_api::get_rpc, get_signer_and_x25519_secret};
-    use rand_core::OsRng;
-    use sp_core::Pair;
+) -> Result<StatusCode, AttestationErr> {
+    let (signer, x25519_secret) = get_signer_and_x25519_secret(&app_state.kv_store).await?;
+    let attestaion_requests = OcwMessageAttestationRequest::decode(&mut input.as_ref())?;
```
HCastano, line 40:
```suggestion
    let attestation_requests = OcwMessageAttestationRequest::decode(&mut input.as_ref())?;
```

crates/threshold-signature-server/src/attestation/tests.rs
```diff
@@ -27,25 +31,34 @@ async fn test_attest() {
     initialize_test_logger().await;
     clean_tests();
 
-    let _cxt = test_node_process_testing_state(false).await;
+    let cxt = test_node_process_stationary().await;
     let (_validator_ips, _validator_ids) = spawn_testing_validators(false).await;
+    let api = get_api(&cxt.ws_url).await.unwrap();
+    let rpc = get_rpc(&cxt.ws_url).await.unwrap();
+
+    // Check that there is an attestation request at block 3 from the genesis config
```
HCastano, line 39: Why block 3?

HCastano (approved):
Left some comments and questions, but nothing major. Good work 👍 

Make sure to add a `CHANGELOG` entry before merging

crates/threshold-signature-server/src/attestation/tests.rs
ameba23: This was one of those problems which kind of got sorted but i don't completely understand why.

The propagation pallet's off chain worker hooks which run every block don't run on the genesis block.  It seems to me that they also don't run on block one - although me and Jesse were getting different behavior. In the end i've ended up with block 3, and also am not subtracting 1 from the block number. Probably this could be either changed to 2, or the subtract 1 could be put back in.

node/cli/src/chain_spec/dev.rs
ameba23: This is a block number, not an account id

pallets/attestation/src/benchmarking.rs
ameba23: Good idea.  I've added the check that there is no longer a pending attestation. Not sure what other events i could check for.

pallets/attestation/src/lib.rs
ameba23: Good point. I thought it is not possible to have a <T::AccountId> as a value in a `StorageMap`.  As in all our other pallets, we use `Vec<u8>`.  But actually i just tried it, and it compiled fine.

So now the issue is that in `entropy_shared::OcwMessageAttestationRequest` we represent this as a `[u8; 32]`. Since theres no way to turn a generic `T::AccountId` into a `[u8; 32]`, we would have to make `OcwMessageAttestationRequest` generic for chain config.

I'd be happy to do this, but i have the feeling it there might be issues with deriving `Serialize` and `Deserialize` and probably it would make sense to also do it to all the other types we have in `entropy-shared` which represent account ids with `Vec<u8>`. So maybe its for another PR.

pallets/attestation/src/lib.rs
ameba23: I've moved it lower down for now but Im not sure about this.

I think once we have some specific outcome for a failed attestation, we will still want to remove the pending attestation even if it is bad. So pending attestations will just be ones that haven't been responded to yet and we will have some other way of representing ones which have been responded to but were bad.

pallets/attestation/src/lib.rs
ameba23: Will add this to the issue - i am going to make a list of things in the main issue rather than separate issues for now.

pallets/propagation/src/lib.rs
ameba23: Ah ok yep will do.  

Although i find it strange that we store a bunch of these urls for each endpoint, rather than just a top level one.

crates/threshold-signature-server/src/attestation/api.rs
ameba23: I'll put it in the issue - i am adding a bunch of things into the once issue for this feature.

But since we check below with the chain that there is a pending attestation, im not sure what we gain by also checking the `OcwMessageAttestationRequest` matches on chain.

pallets/attestation/src/lib.rs
ameba23: I've added this to the TODOs in the issue

PR #1004, opened 2024-08-13
Add `network-jumpstart` command to `entropy-test-cli`

HCastano:
This PR adds a way to trigger a network jumpstart from the test CLI. This is useful for ensuring the
network is in the correct state before registering using the new registration flow.

### DIFF
--- a/crates/client/Cargo.toml
+++ b/crates/client/Cargo.toml
@@ -33,7 +33,7 @@ anyhow          ="1.0.86"
 
 # Only for the browser
 js-sys={ version="0.3.70", optional=true }
-tokio ="1.39"
+tokio ={ version="1.39", features=["time"] }
 
 [dev-dependencies]
 serial_test          ="3.1.1"
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -369,6 +369,18 @@ pub async fn jumpstart_network(
     rpc: &LegacyRpcMethods<EntropyConfig>,
     signer: sr25519::Pair,
 ) -> Result<(), ClientError> {
+    // We split the implementation out into an inner function so that we can more easily pass a
+    // single future to the `timeout`
+    tokio::time::timeout(std::time::Duration::from_secs(45), jumpstart_inner(api, rpc, signer))
+        .await
+        .map_err(|_| ClientError::JumpstartTimeout)?
+}
+
+async fn jumpstart_inner(
+    api: &OnlineClient<EntropyConfig>,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+    signer: sr25519::Pair,
+) -> Result<(), ClientError> {
     // In this case we don't care too much about the result because we're more interested in the
     // `FinishedNetworkJumpStart` event, which happens later on.
     let jump_start_request = entropy::tx().registry().jump_start_network();
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -360,3 +360,31 @@ pub async fn change_threshold_accounts(
         .ok_or(anyhow!("Error with transaction"))?;
     Ok(result_event)
 }
+
+/// Trigger a network wide distributed key generation (DKG) event.
+///
+/// Fails if the network has already been jumpstarted.
+pub async fn jumpstart_network(
+    api: &OnlineClient<EntropyConfig>,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+    signer: sr25519::Pair,
+) -> Result<(), ClientError> {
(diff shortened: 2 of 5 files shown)


crates/client/src/client.rs
```diff
@@ -360,3 +360,31 @@ pub async fn change_threshold_accounts(
...
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+    signer: sr25519::Pair,
+) -> Result<(), ClientError> {
+    // In this case we don't care too much about the result because we're more interested in the
+    // `FinishedNetworkJumpStart` event, which happens later on.
+    let jump_start_request = entropy::tx().registry().jump_start_network();
+    let _result =
+        submit_transaction_with_pair(api, rpc, &signer, &jump_start_request, None).await?;
+
+    let mut blocks_sub = api.blocks().subscribe_finalized().await?;
```
ameba23, line 378: Ah i didn't know about this, this is so much nicer that polling with a timer.

crates/client/src/client.rs
```diff
@@ -360,3 +360,31 @@ pub async fn change_threshold_accounts(
...
+
+    while let Some(block) = blocks_sub.next().await {
```
ameba23, line 380: What happens if the jump start DKG fails?  Im guessing we just wont get confirmations and this will hang forever.

crates/test-cli/src/lib.rs
```diff
@@ -143,6 +143,17 @@ enum CliCommand {
...
+    /// Triggers the network wide distributed key generation process.
+    ///
+    /// A fully jumpstarted network is required for the on-chain registration flow to work
+    /// correctly.
+    ///
+    /// Note: Any account may trigger the jumpstart process.
+    JumpstartNetwork {
+        /// The mnemonic for the signer which will trigger the jumpstart process.
+        #[arg(short, long)]
+        mnemonic_option: Option<String>,
```
ameba23, line 155: unrelated to this PR but i still think its insane that this isn't a positional argument

crates/client/src/client.rs
```diff
@@ -360,3 +360,31 @@ pub async fn change_threshold_accounts(
...
+/// Fails if the network has already been jumpstarted.
+pub async fn jumpstart_network(
+    api: &OnlineClient<EntropyConfig>,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+    signer: sr25519::Pair,
+) -> Result<(), ClientError> {
+    // In this case we don't care too much about the result because we're more interested in the
+    // `FinishedNetworkJumpStart` event, which happens later on.
+    let jump_start_request = entropy::tx().registry().jump_start_network();
+    let _result =
```
ameba23, line 375: It says in the doccomment that this fails if the network has already been jumpstarted. I can't see where that would happen.  Isn't there also a case where the network is not yet ready to be jumpstarted because of too few validators?  Or do we just manually decide when its ready by running this command.

ameba23 (approved):
Great. Feels like we are getting close to be able to try all this out

crates/client/src/client.rs
JesseAbram: the fails happens on chain here https://github.com/entropyxyz/entropy-core/blob/master/pallets/registry/src/lib.rs#L270

crates/test-cli/src/lib.rs
HCastano: I actually don't think it should be positional 😉  But we do need to refactor the CLI UX at some point, it's not as consistent or friendly as it can be

crates/client/src/client.rs
HCastano: Yeah, so what would happen is when we submit the `jump_start_network()` extrinsic we'd get back one of those errors instead of a success event (`StartedNetworkJumpStart`), and then we'd bail out of this function

crates/client/src/client.rs
HCastano: Yeah right now it would just sit waiting for the event 😅

crates/client/Cargo.toml
```diff
@@ -33,7 +33,7 @@ anyhow          ="1.0.86"
 
 # Only for the browser
 js-sys={ version="0.3.70", optional=true }
-tokio ="1.39"
+tokio ={ version="1.39", features=["time"] }
```
HCastano, line 36: @ameba23 not sure if adding this feature will be problematic for Wasm compilation. In the docs it seems like it's fine as long as the Wasm environment is in the web, so not `wasm32-unknown-unknown`.

ameba23: i guess if it turns out to be a problem we can find a way to fix it

PR #1006, opened 2024-08-14
Clean attestation pallet

JesseAbram:
Just a few things I saw, feel free to close if not the vibe


ameba23 (approved):
Thank you. I think there might be other things to tidy up - a lot of stuff in the mock runtime together with dependencies were copied from the registry pallet.  and probably a lot of it isn't needed.

PR #1012, opened 2024-08-16
Add test showing register and sign with test client

ameba23:
@HCastano you don't need to merge this, i just wanted to see it passing.

I really wanted to make this an integration test, but it didn't work out.  The cool thing about the new registration flow is that we can test it in integration because it doesn't rely on there being multiple chain nodes.

But not quite. We have the pre-existing parent keyshares - but we don't have a way to get the chain to a state where jumpstart is completed. The staking extension pallet's genesis config would need to let us put the chain into a jump start completed state.  Something for another PR.

### DIFF
--- a/crates/threshold-signature-server/src/helpers/tests.rs
+++ b/crates/threshold-signature-server/src/helpers/tests.rs
@@ -32,6 +32,7 @@ use crate::{
         },
         logger::{Instrumentation, Logger},
         substrate::{query_chain, submit_transaction},
+        validator::get_signer_and_x25519_secret_from_mnemonic,
     },
     signing_client::ListenerState,
     AppState,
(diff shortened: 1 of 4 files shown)


JesseAbram:
> @HCastano you don't need to merge this, i just wanted to see it passing.
> 
> I really wanted to make this an integration test, but it didn't work out. The cool thing about the new registration flow is that we can test it in integration because it doesn't rely on there being multiple chain nodes.
> 
> But not quite. We have the pre-existing parent keyshares - but we don't have a way to get the chain to a state where jumpstart is completed. The staking extension pallet's genesis config would need to let us put the chain into a jump start completed state. Something for another PR.

ya we do, you do it here https://github.com/entropyxyz/entropy-core/pull/1012/files#diff-e1d8c0027f2fda5b3998b919bb461f5935ad1f6690b719c637968150be6f5e76R1581 this "tricks" the network into that state

ameba23:
> ya we do, you do it here https://github.com/entropyxyz/entropy-core/pull/1012/files#diff-e1d8c0027f2fda5b3998b919bb461f5935ad1f6690b719c637968150be6f5e76R1581 this "tricks" the network into that state

i thought we probably wouldn't be able to call that function without `cfg(test)` because it uses test sort of stuff like [`development_mnemonic`](https://github.com/entropyxyz/entropy-core/blob/685a1bbe4e84a495a862f72d94f744e0887ccfce/crates/threshold-signature-server/src/helpers/launch.rs#L256). But actually probably we can and that would be an easier way to do it.  We could re-export it in `testing-utils` to save having to make it public without cfg(test).

HCastano:
@ameba23 this is pretty cool!

I say it would be worth it to spend a little bit of time trying to turn this into an integration test if possible. If not, maybe a another place for it could be the `entropy_client` tests?

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -264,3 +265,25 @@ pub async fn unsafe_get(client: &reqwest::Client, query_key: String, port: u32)
 
     get_result.bytes().await.unwrap().into()
 }
+
+/// Mock the network being jump started by confirming a jump start even though no DKG took place,
+/// so that we can use pre-store parent keyshares for testing
+pub async fn jump_start_network_with_signer(
+    api: &OnlineClient<EntropyConfig>,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+    signer: &PairSigner<EntropyConfig, sr25519::Pair>,
```
ameba23, line 274: This was needed because `sp_keyring` is a development dependency and we now need this fn in release mode if the `test_helpers` feature is present. So here we pass a signer in instead of getting it from the keyring, and the unit tests use a wrapped version of this function which grabs alice from the keyring.

ameba23:
So @JesseAbram you were right the mock jump start will work fine outside of `cfg(test)`.

So i have made this into an integration test.

HCastano (approved):
👍

Issue #1009, opened 2024-08-15
Benchmarks for new_session and on_init in propagation repo 


JesseAbram:
Looks like they are not being done properly need to fix

PR #1013, opened 2024-08-16
Add benchmarks to pallet propagation

JesseAbram:
Related #1009


ameba23 (approved):
Looks good to me - so if i understand right, even though the propagation pallet has no extrinsics, we still need a weight for the `on_initialize` hook, as it makes writes to runtime storage.

Approving pending fixing CI - looks like a trivial fix.

JesseAbram:
> Looks good to me - so if i understand right, even though the propagation pallet has no extrinsics, we still need a weight for the `on_initialize` hook, as it makes writes to runtime storage.
> 
> Approving pending fixing CI - looks like a trivial fix.

yes but not runtime storage, pretty much every operation that takes any form of time should be benchmarked (even if it doesn't require a fee) because it informs the chain how much time the operation takes up in a block, as block space is finite

PR #1016, opened 2024-08-19
Benchmark for new session

### LINKED ISSUE #1009: Benchmarks for new_session and on_init in propagation repo

JesseAbram:
Closes #1009


pallets/staking/src/lib.rs
```diff
@@ -600,9 +605,18 @@ pub mod pallet {
...
+                    log::warn!(
+                        "Error splitting validators, Session: {:?}, reason: {:?}",
+                        new_index,
+                        why
+                    )
+                } else {
+                    frame_system::Pallet::<T>::register_extra_weight_unchecked(
+                        result_weight.expect("Error unwraping non error value"),
+                        DispatchClass::Mandatory,
+                    );
```
HCastano, line 619: This seems really risky to use. From [the docs](https://docs.rs/frame-system/latest/frame_system/pallet/struct.Pallet.html#method.register_extra_weight_unchecked): 

> Even more dangerous is to note that this function does NOT take any action if the new sum of block weight is more than the block weight limit. This is what the unchecked.

This means we could stall the chain at a session boundary

HCastano (changes requested):
Requesting changes because we really need to think about the impact of this.

I haven't been able to find any examples of implementations of `SessionManager` which deal with any weights. I only found one for [`NoteHistoricalRoot`](https://docs.rs/pallet-session/37.0.0/src/pallet_session/historical/mod.rs.html#209) which doesn't deal with weights.

Do you have any other counter examples here? Otherwise I would just leave this out tbh

JesseAbram:
> Requesting changes because we really need to think about the impact of this.
> 
> I haven't been able to find any examples of implementations of `SessionManager` which deal with any weights. I only found one for [`NoteHistoricalRoot`](https://docs.rs/pallet-session/37.0.0/src/pallet_session/historical/mod.rs.html#209) which doesn't deal with weights.
> 
> Do you have any other counter examples here? Otherwise I would just leave this out tbh

ya, used it before in statemine, in the collator selection pallet, if we don't do weights here on session change we would have larger blocks we can add an if statment for the max weight of a block and be like hey don't add the weight if it is greater then a max block weight and have a larger block instead of a chain stall

ameba23 (commented):
Im leaving this as a comment so that @HCastano gets the say on this. I would need you to sit on a call and talk me through this to really get it.

But the fact that the same thing has been done in another pallet which has presumably been audited makes me think its probably fine.

HCastano (approved):

PR #1018, opened 2024-08-21
Rotate parent network key takes 2 steps

### LINKED ISSUE #968: Should deleting and rotating a parent key be a one or two step process
Currently at the end of the reshare, there are two steps, one is the network updates and then two the TSS rotates their key, currently it is a one step process, that could cause out of sync where for whatever reason the call back into the chain fails and now the TSS signers and the chain are out of sync, this could be a two step process to rotate the Parent key

JesseAbram:
Closes #968


crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -176,20 +176,69 @@ pub async fn new_reshare(
 
     let serialized_key_share = key_serialize(&(new_key_share, aux_info))
         .map_err(|_| ProtocolErr::KvSerialize("Kv Serialize Error".to_string()))?;
-    let network_parent_key = hex::encode(NETWORK_PARENT_KEY);
-    // TODO: should this be a two step process? see # https://github.com/entropyxyz/entropy-core/issues/968
-    if app_state.kv_store.kv().exists(&network_parent_key).await? {
-        app_state.kv_store.kv().delete(&network_parent_key).await?
+    let next_network_parent_key = hex::encode(NEXT_NETWORK_PARENT_KEY);
```
ameba23, line 179: Why does this need to be encoded as hex?

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -102,14 +102,94 @@ async fn test_reshare() {
             deserialize(&key_share_and_aux_data_after).unwrap();
 
         // Check key share has changed
```
ameba23, line 104:
```suggestion
        // Check key share has not yet changed
```

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -102,14 +102,94 @@ async fn test_reshare() {
...
-        assert_ne!(serialize(&key_share_before).unwrap(), serialize(&key_share_after).unwrap());
+        assert_eq!(serialize(&key_share_before).unwrap(), serialize(&key_share_after).unwrap());
         // Check aux info has changed
```
ameba23, line 106:
```suggestion
        // Check aux info has not yet changed
```

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -102,14 +102,94 @@ async fn test_reshare() {
...
+            serialize(&key_share_before).unwrap(),
+            serialize(&key_share_after_rotate).unwrap()
+        );
+        // Check aux info has changed
+        assert_ne!(
+            serialize(&aux_info_before).unwrap(),
+            serialize(&aux_info_after_rotate).unwrap()
+        );
+
+        // calling twice doesn't do anything
```
ameba23, line 131: :+1:

ameba23 (approved):
:+1:

crates/shared/src/constants.rs
```diff
@@ -61,6 +61,9 @@ pub const DEVICE_KEY_PROXY: &[u8] = include_bytes!("../device_key_proxy.wasm");
 /// Network parent key specific size to fit into [u8; 32] to save extra code
 pub const NETWORK_PARENT_KEY: &str = "NETWORK_PARENT_KEY_FOR_ENTROPY_";
 
+// Next Network parent key ready to be swapped in to new one
```
ameba23, line 64: Maybe something like
```suggestion
// Lookup key for the next network parent keyshare ready to be swapped in to new one
```

HCastano:
Will take a look at this on Monday

crates/threshold-signature-server/src/validator/api.rs
JesseAbram: the type that the kv takes needs it, but tbh should probably make the const that

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -176,20 +176,69 @@ pub async fn new_reshare(
...
+    let new_parent_key = app_state.kv_store.kv().get(&next_network_parent_key_heading).await?;
+
+    if app_state.kv_store.kv().exists(&network_parent_key_heading).await? {
+        app_state.kv_store.kv().delete(&network_parent_key_heading).await?;
+    };
+
+    app_state.kv_store.kv().delete(&next_network_parent_key_heading).await?;
+
+    let reservation = app_state.kv_store.kv().reserve_key(network_parent_key_heading).await?;
+    app_state.kv_store.kv().put(reservation, new_parent_key).await?;
```
HCastano, line 238: Does this mean we should get rid of the `is_signer_or_delete_parent_key` step in the other flow? We want to wait for this to be called before deleting the parent key

HCastano (commented):
So I don't necessarily think that a two phase approach is bad, but I'm also not sure if we should be adding an extra OCW flow just yet.

I'd either want to see us explore other approaches (e.g maybe just having the TSS subscribe to an on-chain event post-confirmation before deleting), or see where this actually ends up failing in practice with the current implementation.

crates/threshold-signature-server/src/validator/api.rs
JesseAbram: mmm yes that is a good point, that should be moved to the second step

JesseAbram:
> So I don't necessarily think that a two phase approach is bad, but I'm also not sure if we should be adding an extra OCW flow just yet.
> 
> I'd either want to see us explore other approaches (e.g maybe just having the TSS subscribe to an on-chain event post-confirmation before deleting), or see where this actually ends up failing in practice with the current implementation.

personally Im more in favour of pushing then pulling from the TSS, plus I mean that would break convention here, and it is like we right now always have a flow that the chain pushes to the TSS but randomly we set up one long running listener, I know the ocw code is a little more verbose but I feel pretty strongly about keeping only push methods

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -236,6 +285,30 @@ pub async fn validate_new_reshare(
...
+/// We only care that this happens after a reshare so we just check that message isn't stale
+pub async fn validate_rotate_network_key(
+    api: &OnlineClient<EntropyConfig>,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+) -> Result<(), ValidatorErr> {
+    let latest_block_number = rpc
+        .chain_get_header(None)
+        .await?
+        .ok_or_else(|| ValidatorErr::OptionUnwrapError("Failed to get block number".to_string()))?
+        .number;
```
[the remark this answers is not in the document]
JesseAbram: no , I think that would fail as this needs to match when the OCW is fired and that isn't on finalized but on block creation

PR #1019, opened 2024-08-21
Suggestions for `new_session` benches

HCastano:
While reviewing #1016 I saw some things that could've been improved. I ended up just
doing them myself, so Jesse lemme know what you think here.


JesseAbram (approved):
looks good does that mean we are cool with the register weight unchecked call?

HCastano:
@JesseAbram I wouldn't have done all this work just to continue blocking the PR, so yes 😉

PR #1021, opened 2024-08-21
Clean up registration related tests in `pallet-registry`

HCastano:
This cleans up some of the tests in the Registry pallet related to registration. The idea
was to make sure that any old tests were still valid for the new registration flow, or
where discarded.


pallets/registry/src/tests.rs
```diff
@@ -761,68 +772,6 @@ fn it_fails_on_non_matching_verifying_keys() {
...
-
-        assert_noop!(
-            Registry::register(RuntimeOrigin::signed(1), 2, programs_info),
-            Error::<Test>::NoProgramSet
-        );
-    });
-}
-
-#[test]
-fn it_fails_empty_program_list() {
```
ameba23, line 818: Do we not want to replace this with a similar test for `register_on_chain`?  Did you do that somewhere and i missed it?

pallets/registry/src/tests.rs
```diff
@@ -77,6 +77,109 @@ fn it_tests_get_validators_info() {
...
+                bytecode: empty_program.clone(),
+                configuration_schema: empty_program.clone(),
+                auxiliary_data_schema: empty_program.clone(),
+                oracle_data_pointer: empty_program.clone(),
+                deployer: 1,
+                ref_counter: 0,
+            },
+        );
+
+        assert_ok!(Registry::register(
```
ameba23, line 117: Why still test the old flow here?  Isn't it about to be removed?

ameba23 (approved):
:star: Great that change program instance and change modification account now work with the new registration flow.

Approving pending getting the TSS tests to pass.  It looks like the problem is that they use the pre-registered accounts which are currently only registered with the old flow.  I guess we need to add them to the `RegisteredOnChain` map, giving them each a derivation path, and update their verifying keys in the `entropy-shared` constants.

I'm not sure if it is an issue that one of them is eve, and eve is also the pre-registered network parent key.  So probably we shouldn't change eve's verification key. :laughing: 

The other option would be to get rid of the pre-registered accounts and actually register in those tests. Since they are maybe not so helpful now that we don't need to do a per-user DKG anymore.

pallets/registry/src/tests.rs
HCastano: I had initially removed it in this PR but reverted it since I figured it would be better to remove the whole confirmation flow in a different PR (which would include the removal on the TSS side as well).

pallets/registry/src/tests.rs
HCastano: This was [added in a previous PR](https://github.com/entropyxyz/entropy-core/blob/1375a5bb0afc9d7f0338046c60995e718f97ebcd/pallets/registry/src/tests.rs#L325)

HCastano:
> The other option would be to get rid of the pre-registered accounts and actually register in those tests. Since they are maybe not so helpful now that we don't need to do a per-user DKG anymore.

@ameba23 yeah this is probably what I'm going to end up doing, but in the PR where I actually remove the old registration flow. It's a bit annoying to populate them in genesis anyways.

I'm going to `#[ignore]` the failing test for now since this is going to be addressed either today or tomorrow

PR #1022, opened 2024-08-21
Remove `prune_registration` extrinsic

HCastano:
With the new registration flow users are registered in a single block. This means that
the prune registration extrinsic is not needed anymore since registration are never in a
pending state.


ameba23 (approved):
Great :star:

JesseAbram (approved):
looks good but probably definitely should also pull_metadata

PR #1024, opened 2024-08-22
Check build-time measurement (MRTD) values when verifying TDX quotes

ameba23:
Related issue: https://github.com/entropyxyz/entropy-core/issues/982

This adds a constant containing an array of acceptable build-time measurement values (MRTD) which are checked on verifying a TDX quote, in order to ensure that the attestee is running one of our currently-supported releases of `entropy-tss`.

Currently however these are just some mock values.

Things to consider:
- Somewhere we probably want a mapping of version numbers to these values.
- We want to support verifying mock quotes for testing on non-tdx hardware.  So either we need a separate copy of this array if a feature flag indicating non-production environment is present, or we need a way of passing the desired version we are mocking when creating a mock quote, so that it passes this check.
- There is a bit of a chicken-and-egg problem with getting the value for the current release, as we need the built release to get MRTD value, but we need to include the MRTD value in the release.


crates/shared/src/constants.rs
```diff
@@ -73,3 +73,8 @@ pub const SIGNER_THRESHOLD: u8 = 2;
 
 /// For testing to line up chain mock data and reshare_test
 pub const TEST_RESHARE_BLOCK_NUMBER: u32 = 5;
+
+/// Accepted values of the TDX build-time measurement register - from the currently-supported
+/// releases of entropy-tss
+/// TODO #982 These are currently just mock values
+pub const ACCEPTED_MRTD_VALUES: [[u8; 48]; 2] = [[0; 48], [1; 48]];
```
JesseAbram, line 80: so I may not fully understand here so this is more of a discussion comment, but if we put this in the parameters  pallet and behind a DAO operation seems like a better flow

JesseAbram (commented):
I think the placement of the const needs to be discussed first

crates/shared/src/constants.rs
HCastano: I agree, this should be in the Parameters pallet. 

This would also seem to solve the chicken and egg problem you describe, since we can create a release of `entropy-tss` and then update the MRTD values on-chain.

ameba23: Ok, i can put it in the parameters pallet - if yous are both happy that it is secure.  My idea was to have it hard coded into the chain runtime so it could only be changed with a chain runtime update. But if changing parameters requires a similar level of authority as a runtime update would, then lets put it there.

JesseAbram: yup if you look at the other function calls there, they use ```T::UpdateOrigin::ensure_origin(origin)?;``` which allows us to tap into the chains built in DAO, which is a lot to type, happy to dive deeper on a call or at the retreat

pallets/parameters/src/lib.rs
```diff
@@ -13,7 +13,7 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! # Programs Parameters
+//! # Parameters Pallet
```
ameba23, line 16: Since this isn't just things related to programs

ameba23:
@JesseAbram @HCastano i am not sure why the chainspec pipeline is failing when building the development config. The failing check checks that the vector of MRTD values from the genesis config is not empty.  But i have added a value to for it in the development chainspec - so it should not be empty.  Any ideas?

pallets/parameters/src/lib.rs
```diff
@@ -67,13 +68,16 @@ pub mod module {
         type WeightInfo: WeightInfo;
     }
 
+    pub type MrtdValues = Vec<BoundedVec<u8, ConstU32<48>>>;
```
ameba23, line 71: I **really** don't like using a `BoundedVec` here instead of `[u8; 48]`. But the problem is this need to be something which implements `serde::Deserialize` - and arrays of more than 32 elements do not. There are some helpful crates which solve this problem but i'd rather not add another dependency.

pallets/parameters/src/lib.rs
```diff
@@ -205,5 +222,18 @@ pub mod module {
             Self::deposit_event(Event::SignerInfoChanged { signer_info });
             Ok(())
         }
+
+        #[pallet::call_index(3)]
+        // #[pallet::weight( <T as Config>::WeightInfo::change_accepted_mrtd_values())]
+        #[pallet::weight({100})]
```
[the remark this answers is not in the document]
HCastano: Oh I guess you haven't actually ran these benches yet

HCastano (approved):
Benches need to be run, but otherwise looks fine.

The CI pipeline is failing because the testnet chainspec doesn't have `accepted_mrtd_values` set. One thing we could do here is change that to be an `Option` and then set that to `None` for the testnet config. But depends on how you want to initialize stuff.

pallets/parameters/src/lib.rs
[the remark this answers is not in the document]
ameba23: No this is something else i wanted to ask about but forgot. 

This is what happens when i run the benchmarks:
```
turnip ~/r/s/e/entropy-core (peg/mrtd-values●)$ ./scripts/single_benchmark.sh
2024-08-29 07:53:13 [0] 💸 generated 3 npos voters, 3 from validators and 0 nominators
2024-08-29 07:53:13 [0] 💸 generated 3 npos targets
2024-08-29 07:53:16 Starting benchmark: pallet_parameters::change_request_limit
2024-08-29 07:53:16 Starting benchmark: pallet_parameters::max_instructions_per_programs
2024-08-29 07:53:16 Starting benchmark: pallet_parameters::change_signers_info
2024-08-29 07:53:16 panicked at /home/turnip/radish/src/entropy/entropy-core/pallets/parameters/src/benchmarking.rs:61:5:
Expected Ok(_). Got Err(
    Module(
        ModuleError {
            index: 56,
            error: [
                3,
                0,
                0,
                0,
            ],
            message: Some(
                "SignerDiffTooLarge",
            ),
        },
    ),
)
Error: Input("Error executing and verifying runtime benchmark: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\nerror while executing at wasm backtrace:\n    0: 0x5ae3e3 - entropy_runtime.wasm!rust_begin_unwind\n    1: 0x13ba5 - entropy_runtime.wasm!core::panicking::panic_fmt::h0ffbf025faab9f3f\n    2: 0x43fea9 - entropy_runtime.wasm!core::ops::function::FnOnce::call_once{{vtable.shim}}::h3b2b2a4ef75db544\n    3: 0x42f992 - entropy_runtime.wasm!pallet_parameters::benchmarking::<impl frame_benchmarking::utils::Benchmarking for pallet_parameters::module::Pallet<T>>::run_benchmark::hb69b39372f065e43\n    4: 0x59c04 - entropy_runtime.wasm!<entropy_runtime::Runtime as frame_benchmarking::utils::runtime_decl_for_benchmark::BenchmarkV1<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u32,sp_runtime::traits::BlakeTwo256>,sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic<sp_runtime::multiaddress::MultiAddress<<<sp_runtime::MultiSignature as sp_runtime::traits::Verify>::Signer as sp_runtime::traits::IdentifyAccount>::AccountId,u32>,entropy_runtime::RuntimeCall,sp_runtime::MultiSignature,(frame_system::extensions::check_spec_version::CheckSpecVersion<entropy_runtime::Runtime>,frame_system::extensions::check_tx_version::CheckTxVersion<entropy_runtime::Runtime>,frame_system::extensions::check_genesis::CheckGenesis<entropy_runtime::Runtime>,frame_system::extensions::check_mortality::CheckMortality<entropy_runtime::Runtime>,frame_system::extensions::check_nonce::CheckNonce<entropy_runtime::Runtime>,frame_system::extensions::check_weight::CheckWeight<entropy_runtime::Runtime>,pallet_transaction_payment::ChargeTransactionPayment<entropy_runtime::Runtime>,pallet_registry::pallet::ValidateConfirmRegistered<entropy_runtime::Runtime>)>>>>::dispatch_benchmark::h38bdcb9cc0d739fa\n    5: 0x287fa9 - entropy_runtime.wasm!Benchmark_dispatch_benchmark")
```

The assertion which fails is from `change_signers_info` - which i don't think should have changed: 
https://github.com/entropyxyz/entropy-core/blob/f58d2b8d27b714bf359a3b5ac8d6db953448027a/pallets/parameters/src/benchmarking.rs#L61

ameba23:
> The CI pipeline is failing because the testnet chainspec doesn't have `ParametersConfig.accepted_mrtd_values` set. One thing we could do here is change that to be an `Option` and then set that to `None` for the testnet config. But depends on how you want to initialize stuff.

Ah great good catch i missed that. :pray:

pallets/parameters/src/lib.rs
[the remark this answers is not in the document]
HCastano: Fixed the original problem in #1035.

ameba23:
I have updated the weights - but only for the one extrinsic i have added here, as @HCastano suggested. So once https://github.com/entropyxyz/entropy-core/pull/1035 is merged we should maybe run the benchmarks again (or just wait until we run them when we do a release).

PR #1025, opened 2024-08-22
Remove `confirm_registered` extrinsic

HCastano:
This is another extrinsic which we don't need anymore as a part of the new registration
flow.


crates/threshold-signature-server/src/user/api.rs
```diff
@@ -437,8 +437,8 @@ pub async fn is_registering(
     Ok(register_info.is_some())
 }
 
-/// Confirms that a address has finished registering on chain.
-pub async fn confirm_registered(
+/// Confirms that the network wide distributed key generation process has taken place.
+pub async fn confirm_jump_start(
     api: &OnlineClient<EntropyConfig>,
     rpc: &LegacyRpcMethods<EntropyConfig>,
     who: SubxtAccountId32,
```
ameba23, line 444: Not sure if we still need this argument, or the check below since this should only ever be called with the network parent key but no harm in leaving it in.

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -681,6 +679,7 @@ async fn test_program_with_config() {
     clean_tests();
 }
 
+#[ignore]
```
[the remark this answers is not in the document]
ameba23: omg i can't wait to see this test get deleted it has been a pita

do we not already have tests which use the new registration flow?

pallets/registry/src/tests.rs
```diff
@@ -739,200 +629,12 @@ fn it_fails_on_non_matching_verifying_keys() {
...
-
-        // not registered or registering
-        assert_eq!(Registry::registering(1), None);
-        assert_eq!(Registry::registered(expected_verifying_key.clone()), None);
-    })
-}
+        assert_ok!(Registry::register(RuntimeOrigin::signed(1), 2, programs_info.clone(),));
 
-#[test]
-fn it_provides_free_txs_confirm_done() {
```
ameba23, line 775: Im guessing since we changed how we do 'free' transactions we don't need the equivalent of these tests for confirming jump start completed

ameba23 (approved):
:+1:

pallets/registry/src/tests.rs
HCastano: We're not doing free transactions for the jump start confirmation. If we do go down that route we may want to bring these back, but no point in keeping them around right now

crates/threshold-signature-server/src/user/api.rs
HCastano: Yeah I think this can get cleaned up, but it'll be more clear how when I rip out the old registration flow

PR #1026, opened 2024-08-23
Update TSS tests to use new registration flow

HCastano:
In #1021 I `#[ignore]`'d a few tests that dependend on the old registration flow. In this
PR I bring (most of) those tests back and update them to use the new registration flow.

The main thing I wanted to achieve here was to not lose any test coverage switching to
the new flow.

This was a pain to go through due to the slow run time of the tests (anywhere from 1-4
mins) and the jumbled nature of some tests (e.g many things being tested as a time).

To help with this I split out some test sections into their own tests. There's definitely
still a lot of work to be done in cleaning up some of the tests (like
`test_sign_tx_no_chain`, now `test_signature_requests_fail_on_different_conditions`), but
that's something to address in later PRs.

After this PR I think we'll be good to remove the old registration flow completely.


crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -159,24 +159,34 @@ async fn test_get_signer_does_not_throw_err() {
...
+    // Here we need to use `--chain=integration-tests` and force authoring otherwise we won't be
+    // able to get our chain in the right state to be jump started.
+    let force_authoring = true;
+    let substrate_context = test_node_process_testing_state(force_authoring).await;
+    let entropy_api = get_api(&substrate_context.ws_url).await.unwrap();
+    let rpc = get_rpc(&substrate_context.ws_url).await.unwrap();
+
+    // We first need to jump start the network and grab the resulting network wide verifying key
+    // for later
+    jump_start_network(&entropy_api, &rpc).await;
```
HCastano, line 183: Since this pattern is so common we may want to put it in its own function, but maybe future work

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -605,253 +658,248 @@ async fn test_sign_tx_no_chain_fail() {
...
     }
+
     clean_tests();
 }
 
 #[ignore]
 #[tokio::test]
 #[serial]
-async fn test_program_with_config() {
+async fn test_fails_to_sign_if_non_signing_group_participants_are_used() {
```
HCastano, line 678: The diff is a bit confusing here, but this test was split out from the old `test_sign_tx_no_chain` test. 

@ameba23 do you think this is still needed in its current form? If so, can you maybe take a look at it for me, it's failing and tbh I'm not sure how to fix it 🙏

crates/threshold-signature-server/tests/sign.rs
HCastano, line 1: This functionality is covered in `register_and_sign.rs`

ameba23 (approved):
Great. Much clearer naming of the tests and comments.

There are still a few unused variable warnings in tests. 

As you mentioned, we could probably make this more concise if we had a helper which stored a program and registered a user.

Also, almost every time we call `get_sign_tx_data`, we change the verifying key in the signature request afterwards - so maybe this function should take the verifying key as an argument.

Maybe we can remove the pre-registered users from the chainspec in a follow-up, since we are no longer using them in these tests.

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -185,32 +195,55 @@ async fn test_sign_tx_no_chain() {
...
+    // The account we registered does have a program pointer, so this should succeed
     update_programs(
         &entropy_api,
         &rpc,
-        DAVE_VERIFYING_KEY,
-        &one.pair(),
+        verifying_key.as_slice().try_into().unwrap(),
+        &two.pair(),
         OtherBoundedVec(vec![
             OtherProgramInstance { program_pointer: program_hash, program_config: vec![] },
```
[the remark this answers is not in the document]
HCastano: Yeah I noticed that too. A bit strange and I would like to look into it at some point since its probably unnecessary

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -185,32 +195,55 @@ async fn test_sign_tx_no_chain() {
...
+            .await;
 
     for res in test_no_program {
         assert_eq!(res.unwrap().text().await.unwrap(), "No program pointer defined for account");
     }
+
+    // Test: We check that an account with a program succeeds in submiting a signature request
+
+    // The account we registered does have a program pointer, so this should succeed
     update_programs(
```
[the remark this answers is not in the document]
HCastano: You're right, I've moved it.

This goes to show my point - the subtest setup within each test is jumbed 😄

HCastano:
> Great. Much clearer naming of the tests and comments.

🙏 

> There are still a few unused variable warnings in tests.

Yep fixed - I basically just pushed as soon as all the tests passed lol 

> As you mentioned, we could probably make this more concise if we had a helper which stored a program and registered a user.
> 
> Also, almost every time we call get_sign_tx_data, we change the verifying key in the signature request afterwards - so maybe this function should take the verifying key as an argument.
> 
> Maybe we can remove the pre-registered users from the chainspec in a follow-up, since we are no longer using them in these tests.

Agreed with all of these. Let's leave them for follow ups though.

PR #1028, opened 2024-08-26
Fix test for rejecting a TSS node who is not part of the signing committe

ameba23:
This is based off https://github.com/entropyxyz/entropy-core/pull/1026 and fixes the test which checks that a TSS server who is not a member of the expected signing committee is unable to participate in the signing protocol.

This test was actually pretty broken before - and i am to blame.  It was sending the signature request to one validator, and then attempting a bogus connection to another validator.  Which correctly failed because they were not expecting a signing session at all.  But what we want to test here is that when they are expecting a signing session, they will only allow members of the signing committee to connect.

### DIFF
--- a/crates/threshold-signature-server/src/signing_client/protocol_transport.rs
+++ b/crates/threshold-signature-server/src/signing_client/protocol_transport.rs
@@ -170,8 +170,7 @@ async fn handle_initial_incoming_ws_message(
     };
 
     {
-        // Check that the given public key matches the public key we got in the
-        // UserTransactionRequest or register transaction
+        // Check that the given public key is of the ones we are expecting for this protocol session
         let mut listeners = app_state
             .listener_state
             .listeners
@@ -186,8 +185,7 @@ async fn handle_initial_incoming_ws_message(
             // Make the signing process fail, since one of the commitee has misbehaved
             listeners.remove(&msg.session_id);
             return Err(SubscribeErr::Decryption(
-                "Public key does not match that given in UserTransactionRequest or register \
-                 transaction"
+                "Public key does not match any of those expected for this protocol session"
                     .to_string(),
             ));
         }
--- a/crates/threshold-signature-server/src/user/tests.rs
+++ b/crates/threshold-signature-server/src/user/tests.rs
@@ -672,7 +672,6 @@ async fn test_request_limit_are_updated_during_signing() {
     clean_tests();
 }
 
-#[ignore]
 #[tokio::test]
 #[serial]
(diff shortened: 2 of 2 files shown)


crates/threshold-signature-server/src/signing_client/protocol_transport.rs
```diff
@@ -186,8 +185,7 @@ async fn handle_initial_incoming_ws_message(
             // Make the signing process fail, since one of the commitee has misbehaved
             listeners.remove(&msg.session_id);
             return Err(SubscribeErr::Decryption(
-                "Public key does not match that given in UserTransactionRequest or register \
-                 transaction"
+                "Public key does not match any of those expected for this protocol session"
```
ameba23, line 188: I changed this error message and the comment above because they mentioned registration which no longer involves a protocol session.

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -780,20 +779,16 @@ async fn test_fails_to_sign_if_non_signing_group_participants_are_used() {
...
 
-    // thread 'user::tests::test_fails_to_sign_if_non_signing_group_participants_are_used' panicked at crates/threshold-signature-server/src/user/tests.rs:848:9:
-    // assertion `left == right` failed
-    //   left: "{\"Err\":\"Subscribe message rejected: NoListener(\\\"no listener\\\")\"}"
-    //  right: "{\"Err\":\"Timed out waiting for remote party\"}"
     for res in test_user_bad_connection_res {
         assert_eq!(
             res.unwrap().text().await.unwrap(),
-            "{\"Err\":\"Timed out waiting for remote party\"}"
+            "{\"Err\":\"Oneshot timeout error: channel closed\"}"
```
ameba23, line 791: I think what error we get here is not too important - the main thing is that the connection attempt fails in the assertion below.

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -780,20 +779,16 @@ async fn test_fails_to_sign_if_non_signing_group_participants_are_used() {
     signature_request.signature_verifying_key = verifying_key.to_vec();
 
     let test_user_bad_connection_res = submit_transaction_requests(
-        vec![validator_ips_and_keys[1].clone()],
+        vec![validator_ips_and_keys[0].clone()],
```
ameba23, line 782: This was the bit that was wrong before. We should be sending the signature request to the same TS server that we attempt to make a connection to.

HCastano: 🙏

Issue #981, opened 2024-08-05
Key reshare test should demonstrate signing a message following a reshare


ameba23:
For context see https://github.com/entropyxyz/entropy-core/issues/976 and related PR: https://github.com/entropyxyz/entropy-core/pull/977

When testing key reshare, we should check that its possible to sign a message with the resulting keyshares. We currently cannot do this because its not possible to sign a message with the network parent key, and signing with a derived key is not yet implemented.

This issue is a reminder to add signing a message to the reshare test once the new signing flow is implemented.

Issue #1010, opened 2024-08-15
Full end to end test


JesseAbram:
Not immediate but there can be cases where if a refresh is done in properly it can break signing, we should have a full e2e test where we do a dkg a signing a  reshare and another signing to make sure eveything is all good

ameba23:
To do this we would need at least 4 TS servers in our test setup right?  Because with 3 we never actually do a reshare - we can only mock it.

PR #1029, opened 2024-08-26
Add test environment with four TS servers

ameba23:
For some tests it would be useful to have four TS servers - in particular the reshare tests, so that we can test adding and removing members of the signing committee during a reshare.

My personal motivation is that i am thinking to request an attestation from the new member of the signing committee before doing a reshare - and ideally i would like to test the new session handler as part of this.

This PR makes the integration test chainspec have 4 TS servers, and leaves the 'dev' chainspec with only 3 as before. Since the extra node, Dave, is not in the initial signing committee and has no initial keyshares, most of our exisiting tests which use the integration test chainspec should continue to work as they are.

This also extends the reshare test to demonstrate siging a message after a reshare. This closes https://github.com/entropyxyz/entropy-core/issues/981 and partially https://github.com/entropyxyz/entropy-core/issues/1010 - i say partially because although we test signing after a reshare, we mock the initial jumpstart DKG part - so its not really testing the complete process.

Im willing to close this if people think that testing with 4 TS servers is going to make our tests too heavy in terms of resources.


pallets/staking/src/lib.rs
```diff
@@ -274,14 +274,10 @@ pub mod pallet {
...
             if self.mock_signer_rotate.0 {
-                self.mock_signer_rotate
-                    .clone()
-                    .1
-                    .push(self.mock_signer_rotate.clone().2[0].clone());
-                NextSigners::<T>::put(NextSignerInfo {
-                    next_signers: self.mock_signer_rotate.clone().1,
-                    confirmations: vec![],
-                });
+                let next_signers = &mut self.mock_signer_rotate.1.clone();
```
ameba23, line 277: This hasn't changed how we interpret mock_signer_rotate - but it wasn't working for me as it was. I think it was mutating a clone of `self.mock_signer_rotate` but then using a different clone for next signers.

crates/testing-utils/src/helpers.rs
```diff
@@ -0,0 +1,54 @@
...
+    spawn_testing_validators,
+    substrate_context::{test_context_stationary, test_node_process_testing_state},
+    ChainSpecType,
+};
+use entropy_protocol::PartyId;
+use subxt::{backend::legacy::LegacyRpcMethods, OnlineClient};
+
+/// A helper for setting up tests which starts both a set of TS servers and a chain node and returns
+/// the chain API as well as IP addresses and PartyId of the started validators
+pub async fn spawn_tss_nodes_and_start_chain(
```
ameba23, line 27: This isn't strictly related to this PR, but since adding a way to specify which chainspec you want to use, i also added this helper which could slightly reduce boilerplate in our tests.

So this would go at the beginning of most of our tests. I haven't actually called this anywhere yet though - i wanted to run it by yous first.

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -43,41 +54,45 @@ use entropy_testing_utils::{
...
 use parity_scale_codec::Encode;
 use serial_test::serial;
+use sp_core::Pair;
 use sp_keyring::AccountKeyring;
+use subxt::utils::AccountId32;
+use synedrion::k256::ecdsa::VerifyingKey;
 
 #[tokio::test]
 #[serial]
 async fn test_reshare() {
```
ameba23, line 64: Here we try out the 4 node setup with a reshare moving the signers from alice, bob and charlie to bob, charlie and dave.

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -119,10 +118,28 @@ pub async fn create_clients(
     (app, kv_store)
 }
 
-/// Spawn 3 TSS nodes with pre-stored keyshares
-pub async fn spawn_testing_validators(add_parent_key: bool) -> (Vec<String>, Vec<PartyId>) {
+/// A way to specify whether the test environment uses the Development chainspec, which has 3 TSS
+/// nodes, or the Integration test chainspec which has 4 TSS nodes
```
HCastano, line 122: nit: I would add the info how many nodes per config on the variant doc comments instead of the enum comments

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -14,19 +14,27 @@
...
     },
 };
+use entropy_client::{
+    chain_api::{
+        entropy::runtime_types::bounded_collections::bounded_vec::BoundedVec,
+        entropy::runtime_types::pallet_registry::pallet::ProgramInstance, get_api, get_rpc,
+    },
+    Hasher,
+};
+use entropy_client::{register, sign, store_program};
```
HCastano, line 37: I'm personally a fan of the `use entropy_client as test _client` import since in the test it's clear that these functions are coming from the client and not part of this crate

crates/testing-utils/src/helpers.rs
HCastano: I like this for sure 👍

HCastano (approved):
Tbh not a fan of spinning up four nodes as part of each test, but it's one of those things that if we need to do it to get sufficient test coverage then it is what is it 🤷 

Would be curious about what kind of system resources we're consuming and if this is really a concern though.

ameba23:
> Tbh not a fan of spinning up four nodes as part of each test, but it's one of those things that if we need to do it to get sufficient test coverage then it is what is it 🤷

I hear you. Having choice between development or integration test setup is not so helpful, because if i understand correctly, we need the integration test setup's force authoring thingy in order to jumpstart the network - which is pretty much all of our tests - meaning we always use four nodes.

It is actually possible to use the integration test chainspec and only 3 TSS nodes.  I tried it and for registering and signing, no-one complains that dave is not there. Its only for a reshare that dave is needed.  But it still kinda feels wrong to have dave in the chainspec and not be running.

So one option to solve this might be to add the force authoring stuff to the development chainspec, and then use that for all the tests which don't involve a reshare. 

> Would be curious about what kind of system resources we're consuming and if this is really a concern though.

Tbh i think in comparison to the chain node it is relatively little.  Especially as we don't actually run the TSS servers in separate OS processes - its just one `entropy-tss` process with axum serving on different ports and using different files for the kvdb.

PR #1031, opened 2024-08-26
No unbonding when signer or next signer

### LINKED ISSUE #942: Handle unbonding and slashing while a signer
A signer is a special role that holds the network parent key. We should not allow unbonding of validators while a signer, they will get removed eventually, a little shitty for users but I am cool with this. As well when we get to slashing would have to force a rotation of the signers to get the offending party out

JesseAbram:
Related #942

This is a little controversial as it hurts UX a bit but I'm ok with that. I did wanna bring up that this breaks convention of staking and maybe if we rename this pallet to staking and mock all the calls in that pallet we can use all the frontends that are built around staking, a little more pallet code but a hella a lot less GUI code. 

Also @HCastano wanna double check all my entries here, I read up and I believe there are only three ways to unbond from a validator 

- chill
- unbond
- withdraw_unbonded

but second eyes on that would be nice.

Closes #942.


ameba23 (approved):
+1 This seems like something we need. 

One question i have is how do we force people to use these wrappers and not call the underlying `pallet-staking` extrinsics?  Are they somehow not exposed by our runtime?

Also, if i understand right, validators can anyway only leave at a session change, which is currently when we do a reshare.  So if they call unbond (or chill etc) during a session, could we mark them as being the one who wants to leave the signing committee at the next reshare - or block them from being chosen as a next signer.  That way validators can leave whenever they want.

The argument against this would be that it would make it more complicated to change the frequency of reshares, which i think we will want to do, and there could be complications or security issues if several members of the signing comittee want to unbond in the same session.

JesseAbram:
> +1 This seems like something we need.
> 
> One question i have is how do we force people to use these wrappers and not call the underlying `pallet-staking` extrinsics? Are they somehow not exposed by our runtime?
> 

That is handled in the base call filter here https://github.com/entropyxyz/entropy-core/pull/1031/files#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecR320-R322


> Also, if i understand right, validators can anyway only leave at a session change, which is currently when we do a reshare. So if they call unbond (or chill etc) during a session, could we mark them as being the one who wants to leave the signing committee at the next reshare - or block them from being chosen as a next signer. That way validators can leave whenever they want.
> 
> The argument against this would be that it would make it more complicated to change the frequency of reshares, which i think we will want to do, and there could be complications or security issues if several members of the signing comittee want to unbond in the same session.

Yes I had this same thought, however this works if one wants to, but if more than one wants to it gets messy, and it isn't like we are locking them in forever, it isn't that long ( a day or two) I think this is worse UX but way simpler, that being said, id say lets do this for now and we can potentially look at and discuss having them pushed out if they want to unbond earlier

pallets/staking/src/lib.rs
```diff
@@ -397,9 +399,74 @@ pub mod pallet {
...
+
+            pallet_staking::Pallet::<T>::unbond(origin, value)?;
+
+            Ok(Some(<T as Config>::WeightInfo::unbond(signers.len() as u32)).into())
+        }
+
+        /// Wraps's substrate chill but checks to make targeted validator sure not signer or next signer
+        #[pallet::call_index(3)]
+        #[pallet::weight(<T as Config>::WeightInfo::chill(MAX_SIGNERS as u32))]
+        pub fn chill(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
HCastano, line 439: Call it fate, but I was [listening to this](https://soundcloud.com/alexunger/all-i-wanna-do) while reviewing 😆

pallets/staking/src/lib.rs
```diff
@@ -397,9 +399,74 @@ pub mod pallet {
...
+            let controller = ensure_signed(origin.clone())?;
+            let ledger =
+                pallet_staking::Pallet::<T>::ledger(StakingAccount::Controller(controller.clone()))
+                    .map_err(|_| Error::<T>::NoThresholdKey)?;
```
HCastano, line 443: This map contains more than just validator info, it also has nominator info so we may want to change the error here to reflect that

pallets/staking/src/lib.rs
```diff
@@ -397,9 +399,74 @@ pub mod pallet {
...
+            let next_signers = Self::next_signers();
+            if next_signers.is_some() {
+                ensure!(
+                    !next_signers
+                        .ok_or(Error::<T>::ReshareNotInProgress)?
+                        .next_signers
+                        .contains(&validator_id),
+                    Error::<T>::NoUnbondingWhenNextSigner
+                );
+            }
```
HCastano, line 460: Since we have the `is_some()` check + an `ok_or()` we might be able to get away with an `if let` or some other combinator here

pallets/staking/src/lib.rs
```diff
@@ -397,9 +399,74 @@ pub mod pallet {
...
```
HCastano, line 429: Can probably split this out into its own validation function which is used for the two or three other extrinsics

pallets/staking/src/tests.rs
HCastano, line 1: Here we should add some tests for the nominator flow

pallets/staking/src/benchmarking.rs
```diff
@@ -124,11 +136,76 @@ benchmarks! {
     assert_last_event::<T>(Event::<T>::ThresholdAccountChanged(bonder, server_info).into());
   }
 
+  unbond {
```
HCastano, line 139: Can you take the body of this (and all the other benches added in this PR) and manually RustFmt it using the [Rust Playground](https://play.rust-lang.org/)? This isn't done by the `fmt` script since it's since a decl macro.

pallets/staking/src/benchmarking.rs
```diff
@@ -124,11 +136,76 @@ benchmarks! {
...
+    let c in 0 .. MAX_SIGNERS as u32;
```
HCastano, line 140:
```suggestion
    let s in 0 .. MAX_SIGNERS as u32;
```

HCastano (changes requested):
Okay, so while the direction of this is correct, the implementation breaks the nominator flow since they can also `chill()`, `unbond()`, etc.

What we need to do here is check if an account is a validator or nominator and act differently depending on the case. See [`chill_stash()`](https://github.com/paritytech/polkadot-sdk/blob/97fa922c85976523f755acdd104fdc77ed63dae9/substrate/frame/staking/src/pallet/impls.rs#L396) from the Staking pallet for example.

HCastano:
> I did wanna bring up that this breaks convention of staking and maybe if we rename this pallet to staking and mock all the calls in that pallet we can use all the frontends that are built around staking, a little more pallet code but a hella a lot less GUI code.

Oh also, can you elaborate on this? Not sure how renaming and mocking ties into using staking frontends

JesseAbram:
> > I did wanna bring up that this breaks convention of staking and maybe if we rename this pallet to staking and mock all the calls in that pallet we can use all the frontends that are built around staking, a little more pallet code but a hella a lot less GUI code.
> 
> Oh also, can you elaborate on this? Not sure how renaming and mocking ties into using staking frontends

mocking may not be the write word, wrapping is probably better. Pretty much all staking GUIs would call staking.chill or staking.function. Our code blocks that for the specific calls (chill, unbond........etc) so we can add some extra logic, it would be stakingExtention.function, if we rename staking_extention to staking and staking to idk frame_staking, and wrap all the rest of the calls in that pallet in staking extention, (and keep the same interface) then the GUIs for staking like polkadot js frontend or any other staking wallets should just work, instead of us having to roll our own

JesseAbram:
> Okay, so while the direction of this is correct, the implementation breaks the nominator flow since they can also `chill()`, `unbond()`, etc.
> 
> What we need to do here is check if an account is a validator or nominator and act differently depending on the case. See [`chill_stash()`](https://github.com/paritytech/polkadot-sdk/blob/97fa922c85976523f755acdd104fdc77ed63dae9/substrate/frame/staking/src/pallet/impls.rs#L396) from the Staking pallet for example.

I think I get it, i think I broke the nominator flow, just to be sure we are on the same page, I still wanna apply the same logic to both however for nominator I need to get who they are nominating and for a validator same flow just get the related stash to the controller

HCastano:
> mocking may not be the write word, wrapping is probably better. Pretty much all staking GUIs would call staking.chill or staking.function. Our code blocks that for the specific calls (chill, unbond........etc) so we can add some extra logic, it would be stakingExtention.function, if we rename staking_extention to staking and staking to idk frame_staking, and wrap all the rest of the calls in that pallet in staking extention, (and keep the same interface) then the GUIs for staking like polkadot js frontend or any other staking wallets should just work, instead of us having to roll our own

Umm, not sure if it'll be that simple. UIs will probably want to do validation on their
end anyways, and we're emitting different events than the Staking pallet (so the UI might
say something succeeded, but in reality it failed). Something worth following up on though.

> I think I get it, i think I broke the nominator flow, just to be sure we are on the same page, I still wanna apply the same logic to both however for nominator I need to get who they are nominating and for a validator same flow just get the related stash to the controller

For the nominator flow we need to allow a nominator to call `chill()`, `unbond()`, etc.
That's not possible right now since the nominator account would fail the checks that the
`controller` is a signer.

So here we have a couple of options:

1. Skip some of these checks in that case that an an account is a nominator **and** not a validator.
2. If a controller is a nominator, check to see who they're delegaging to. If that person
   is a signer then we don't allow them to perform an action.

I can see (2) being pretty annoying for nominators, but this would help prevent the case
where a validator only puts up a bond of say, `1 UNIT` themselves and then uses a
different nomination account to get reach the required stake.

JesseAbram:
> > mocking may not be the write word, wrapping is probably better. Pretty much all staking GUIs would call staking.chill or staking.function. Our code blocks that for the specific calls (chill, unbond........etc) so we can add some extra logic, it would be stakingExtention.function, if we rename staking_extention to staking and staking to idk frame_staking, and wrap all the rest of the calls in that pallet in staking extention, (and keep the same interface) then the GUIs for staking like polkadot js frontend or any other staking wallets should just work, instead of us having to roll our own
> 
> Umm, not sure if it'll be that simple. UIs will probably want to do validation on their end anyways, and we're emitting different events than the Staking pallet (so the UI might say something succeeded, but in reality it failed). Something worth following up on though.
> 
> > I think I get it, i think I broke the nominator flow, just to be sure we are on the same page, I still wanna apply the same logic to both however for nominator I need to get who they are nominating and for a validator same flow just get the related stash to the controller
> 
> For the nominator flow we need to allow a nominator to call `chill()`, `unbond()`, etc. That's not possible right now since the nominator account would fail the checks that the `controller` is a signer.

on reading into it, it actually wouldn't but the check won't apply to a nominator which also needs to be fixed. pretty much stash would not apply to who the nominator is validating but their own stash and would fail, however this info is kept in the Nominators storage struct
> 
> So here we have a couple of options:
> 
> 1. Skip some of these checks in that case that an an account is a nominator **and** not a validator.
> 2. If a controller is a nominator, check to see who they're delegaging to. If that person
>    is a signer then we don't allow them to perform an action.
> 
> I can see (2) being pretty annoying for nominators, but this would help prevent the case where a validator only puts up a bond of say, `1 UNIT` themselves and then uses a different nomination account to get reach the required stake.

yes 2 is where I was landing mentally too, you don't need to bond from your validator and it is needed

pallets/staking/src/mock.rs
```diff
@@ -397,7 +397,7 @@ impl pallet_parameters::Config for Test {
 pub fn new_test_ext() -> sp_io::TestExternalities {
     let mut t = system::GenesisConfig::<Test>::default().build_storage().unwrap();
     let pallet_balances = pallet_balances::GenesisConfig::<Test> {
-        balances: vec![(1, 100), (2, 100), (3, 100), (4, 100)],
+        balances: vec![(1, 100), (2, 100), (3, 100), (4, 100), (7, 100), (8, 100), (101, 200)],
```
HCastano, line 400:
```suggestion
        balances: vec![(1, 100), (2, 100), (3, 100), (4, 100), (7, 100), (8, 100), (9, 100)],
```

For consistency? Or maybe can you explain why you had it as `(101, 200)`?

pallets/staking/src/lib.rs
```diff
@@ -533,6 +573,47 @@ pub mod pallet {
             Ok(ledger.stash)
         }
 
+        /// Ensures that the current validator is not a signer or a next signer
+        pub fn ensure_not_signer_or_next_signer(
```
HCastano, line 577: This should also mention that it's doing a nominator check

HCastano (commented):
I still need to look through the benches, but it does look like the nominator flow is working better now

pallets/staking/src/mock.rs
JesseAbram: nah I just needed a number I knew was not used, 9 probably works

pallets/staking/src/benchmarking.rs
```diff
@@ -124,11 +136,73 @@ benchmarks! {
     assert_last_event::<T>(Event::<T>::ThresholdAccountChanged(bonder, server_info).into());
   }
 
+  unbond {
```
HCastano, line 139: This comment is gonna apply to all three of these benches now: we may need to add another parameter, say `let n in 0 .. MaxNominations` ([1]([MaxNominationsOf](https://docs.rs/pallet-staking/37.0.0/pallet_staking/type.MaxNominationsOf.html)) since the `ensure_not_signer_or_next_signer` helper reads the `Nominator` map from storage each time

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ At the moment this project **does not** adhere to
   cleaned up. A lot of storage entries, events, and extrinsics were removed from the `Registry`
   pallet. The genesis build config was also removed. Additionally, the `new/user/` HTTP endpoint in
   the TSS was removed since it was no longer necessary.
+- in [#1031](https://github.com/entropyxyz/entropy-core/pull/1031), more Staking calls were blocked to go through staking_extention. This makes sure no funds can be unbonded from a validator if they are currently in the signing comittee. This was applied to `unbond`, `chill`, and `withdraw_unbonded`
```
HCastano, line 21:
```suggestion
- In [#1031](https://github.com/entropyxyz/entropy-core/pull/1031), more Staking calls were blocked
  to go through the `staking_extention` pallet. This makes sure no funds can be unbonded from a
  validator if they are currently in the signing comittee. This was applied to `unbond`, `chill`,
  and `withdraw_unbonded`
```

Wrapping this to 100 characters

HCastano (commented):
Main thing here is that I think we need to account for nominators in benches now.

Once that and the other review comments are addressed we should be good to go.

HCastano (approved):

PR #1035, opened 2024-08-29
Fix `change_signers_info` benchmark

HCastano:
In https://github.com/entropyxyz/entropy-core/pull/978/ a couple of extra checks were added to
ensure that parameters related to signing could only change so quickly. This ended up breaking one
of the benchmarks since the benchmark setup used a difference that was larger than the allowed
amount.

I've fixed that by reading the current values from storage (set at Genesis) and updating those by
the allowed amount.

Maybe worth revisiting https://github.com/entropyxyz/entropy-core/pull/916...


ameba23 (approved):
Great

PR #1036, opened 2024-08-30
Add test helper function which stores a program and registers a user 

ameba23:
Now that we no longer have pre-registered accounts, in order to test signing we need to first store a program and register a user.  To reduce boilerplate in our siging tests, this adds a test helper function which stores a default program and registers a user using that program.

I've also made a change to the `get_sign_tx_data` test helper function.  It now takes the verifying key as an argument, rather than defaulting to `DAVE_VERIFYING_KEY`, and sets the block number to the current block number rather than defaulting to zero.

One unrelated thing i noticed:  We have a test called `signature_request_with_derived_account_works` which looks to me like it doesn't test anything extra than our other signing tests - it is left over from when we had two different signing flows. Do yous think we can get rid of it? 
https://github.com/entropyxyz/entropy-core/blob/624b37d3bad1acdee74f724c7a2cf2a3e2ab0159/crates/threshold-signature-server/src/user/tests.rs#L320


crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -291,3 +291,43 @@ pub async fn jump_start_network_with_signer(
...
 }
+
+/// Helper to store a program and register a user. Returns the verify key and program hash.
+#[cfg(test)]
+pub async fn store_program_and_register(
+    api: &OnlineClient<EntropyConfig>,
+    rpc: &LegacyRpcMethods<EntropyConfig>,
+    user: &sr25519::Pair,
+    deployer: &sr25519::Pair,
+) -> ([u8; 33], sp_core::H256) {
```
HCastano, line 302: We should have a typedef for the verifying key somewhere, no?

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -291,3 +291,43 @@ pub async fn jump_start_network_with_signer(
...
+    use entropy_client::{
+        self as test_client,
+        chain_api::entropy::runtime_types::pallet_registry::pallet::ProgramInstance,
+    };
+    use entropy_testing_utils::constants::TEST_PROGRAM_WASM_BYTECODE;
+    use sp_core::Pair;
+
+    let program_hash = test_client::store_program(
+        &api,
+        &rpc,
```
HCastano, line 312:
```suggestion
        api,
        rpc,
```

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -291,3 +291,43 @@ pub async fn jump_start_network_with_signer(
...
+        vec![],
+        vec![],
+        vec![],
+    )
+    .await
+    .unwrap();
+
+    let (verifying_key, _registered_info) = test_client::register(
+        &api,
+        &rpc,
```
HCastano, line 324:
```suggestion
        api,
        rpc,
```

HCastano (approved):
LGTM.

For your question around `signature_request_with_derived_account_works`, the point of that one versus the other tests is that it's a happy path only test.

The ones ones afair test different failure conditions, so I didn't want to mix them up.

JesseAbram (approved):

PR #1041, opened 2024-09-07
In production, generate TDX quote using configfs-tsm

ameba23:
This adds TDX quote generation using configfs-tsm, a linux filesystem interface to generate TDX quotes in production.

If we want to run the TS server on non-production hardware, we have to use either test mode or enable the `unsafe` feature flag, or attestation will fail.

This uses the [`configfs-tsm`](https://github.com/entropyxyz/configfs-tsm) crate which is only a few lines and has no dependencies.  There is [still some work to be done](https://github.com/entropyxyz/configfs-tsm/issues/1) on it but the api used here should stay the same. I am considering moving that code into the [`tdx-quote`](https://github.com/entropyxyz/tdx-quote) crate, to avoid needing to manage yet another crate.


crates/threshold-signature-server/Cargo.toml
```diff
@@ -72,6 +72,7 @@ sha2             ="0.10.8"
 hkdf             ="0.12.4"
 project-root     ={ version="0.2.2", optional=true }
 tdx-quote        ={ git="https://github.com/entropyxyz/tdx-quote", optional=true, features=["mock"] }
+configfs-tsm     ={ git="https://github.com/entropyxyz/configfs-tsm" }
```
HCastano, line 75: I would pin this to a specific commit hash or tag

HCastano (approved):
Looks fine. I took a quick look through the crate and it does look simple, so maybe worth rolling into `tdx-quote` until we feel like it needs to be its own thing

JesseAbram (commented):
looks good but can we make a new flag and put all the tdx stuff behind it?

ameba23:
> looks good but can we make a new flag and put all the tdx stuff behind it?

Personally, i think the default build should be a production build, and if you want unsafe testing features, such as making mock quotes, you should have to specify that, which is currently the case.

I'm willing to do it the other way around - meaning default build makes mock quotes and production build needs a `production` feature to be enabled.  But i want a thumbs up from both of you on that. We also need to think about our release scripts - probably we want a production release and a testnet version release.

HCastano:
@ameba23 
> Personally, i think the default build should be a production build, and if you want unsafe testing features, such as making mock quotes, you should have to specify that, which is currently the case.
>
> I'm willing to do it the other way around - meaning default build makes mock quotes and production build needs a production feature to be enabled. But i want a thumbs up from both of you on that. We also need to think about our release scripts - probably we want a production release and a testnet version release.

Until we get a good testing and release pipeline with TDX hardware we should have the default be non-TDX.

ameba23:
> @ameba23
> 
> > Personally, i think the default build should be a production build, and if you want unsafe testing features, such as making mock quotes, you should have to specify that, which is currently the case.
> > I'm willing to do it the other way around - meaning default build makes mock quotes and production build needs a production feature to be enabled. But i want a thumbs up from both of you on that. We also need to think about our release scripts - probably we want a production release and a testnet version release.
> 
> Until we get a good testing and release pipeline with TDX hardware we should have the default be non-TDX.

Ok i will make a `production` feature flag.  But will wait until my other open PR is done to avoid conflicts.

ameba23:
I have added a `production` flag, and now if that flag is not present mock quotes will always be used.  CI is passing so i am gonna go ahead and hit merge. Lemme know if you(s) are not happy with how that flag is set up.

PR #1045, opened 2024-09-11
Add a programs version

### LINKED ISSUE #688: Add a version number to program Info
Like how we did it with registered accounts this will help with backwards compatibility

JesseAbram:
Adds a version number to programs 

Closes #688 

Paired with https://github.com/entropyxyz/programs/pull/91


crates/shared/src/constants.rs
```diff
@@ -76,3 +76,6 @@ pub const SIGNER_THRESHOLD: u8 = 2;
 
 /// For testing to line up chain mock data and reshare_test
 pub const TEST_RESHARE_BLOCK_NUMBER: u32 = 5;
+
+/// Program version number, must be incremented if version number changes
+pub const PROGRAM_VERSION_NUMBER: u8 = 0;
```
ameba23, line 81: We don't yet have a check for this when evaluating the program, but when we do, i think it might make sense for this constant to be exported by `entropy-programs-runtime` so that it will always match the runtime version(s) which are actually used.

The argument not to would be if we need this constant in the program's pallet, so that we can refuse to accept programs with an unsupported version number.

ameba23 (approved):
+1 for everything here.  

As i mentioned in the programs PR, the only thing i am unsure about would be to maybe make the version number actually be the major/minor version number of `entropy-programs-runtime`.  So the current version would be `0.10` rather than `0`.

crates/shared/src/constants.rs
JesseAbram: I struggled with this decision a little, I came to this for a few reasons, one was it is a PITA the way we are pulling stuff into the templates to make a change and see it reflected. Also we would also have any changes need to be reflected in core before it is usable, and this keeps our updates for releases to core

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ At the moment this project **does not** adhere to
   cleaned up. A lot of storage entries, events, and extrinsics were removed from the `Registry`
   pallet. The genesis build config was also removed. Additionally, the `new/user/` HTTP endpoint in
   the TSS was removed since it was no longer necessary.
+- In [#1045](https://github.com/entropyxyz/entropy-core/pull/1045), `ProgramsInfo` now takes `version_number` to maintain backwards compatibilite if programs runtime is updated
```
HCastano, line 21: Can you wrap this to 90 characters?

HCastano, line 21:
```suggestion
- In [#1045](https://github.com/entropyxyz/entropy-core/pull/1045), `ProgramsInfo` now takes `version_number` to maintain backwards compatibility if programs runtime is updated 
```

RELEASE_CHECKLIST.md
```diff
@@ -14,6 +14,7 @@ tagged as the final release.
 - [ ] If an existing call/extrinsic has changed (new pallet index, new call index, parameter changes,
   etc.), bump `transaction_version` and bump `spec_version`
 - [ ] If you're confused about what to bump, read [this](https://paritytech.github.io/polkadot-sdk/master/sp_version/struct.RuntimeVersion.html)
+- [ ] If the programs runtime has been updated increment ```PROGRAM_VERSION_NUMBER``` in ```shared``` crate
```
HCastano, line 17:
```suggestion
- [ ] If the programs runtime has been updated increment `PROGRAM_VERSION_NUMBER` in `shared` crate
```

crates/client/src/client.rs
```diff
@@ -196,6 +196,7 @@ pub async fn sign(
 }
 
 /// Store a program on chain and return it's hash
+#[allow(clippy::too_many_arguments)]
```
HCastano, line 199: Maybe time to couple the input data

crates/test-cli/src/lib.rs
```diff
@@ -30,6 +30,7 @@ use entropy_client::{
         VERIFYING_KEY_LENGTH,
     },
 };
+pub use entropy_shared::PROGRAM_VERSION_NUMBER;
```
HCastano, line 33: Can you place this on top of the `use` imports? It makes it more clear that this is being re-exported

crates/test-cli/src/lib.rs
```diff
@@ -196,9 +204,23 @@ pub async fn run_command(
...
             let mut programs_info = vec![];
 
-            for program in programs {
+            for (i, program) in programs.into_iter().enumerate() {
+                let program_version_number =
+                    if let Some(ref program_version_numbers) = program_version_numbers {
+                        program_version_numbers[i]
+                    } else {
+                        0u8
+                    };
```
HCastano, line 213: There's probably a way to change this to use `or_default`, since for a `u8` that's `0` anyways

crates/test-cli/src/lib.rs
```diff
@@ -276,6 +299,11 @@ pub async fn run_command(
                 )?,
             };
 
+            let program_version_number = match program_version_number_option {
+                Some(program_version_number) => program_version_number,
+                None => program_version_number.expect("No Version number passed"),
```
HCastano, line 304: I thought we just defaulted to `0` in this case, why fail?

crates/test-cli/src/lib.rs
```diff
@@ -315,9 +349,24 @@ pub async fn run_command(
...
             let mut programs_info = Vec::new();
-            for program in programs {
+
+            for (i, program) in programs.into_iter().enumerate() {
+                let program_version_number =
+                    if let Some(ref program_version_numbers) = program_version_numbers {
+                        program_version_numbers[i]
+                    } else {
+                        0u8
+                    };
```
HCastano, line 359: Same thing here, can probably be tidied

pallets/programs/src/benchmarking.rs
```diff
@@ -58,15 +60,16 @@ benchmarks! {
     let value = CurrencyOf::<T>::minimum_balance().saturating_mul(1_000_000_000u32.into());
     let _ = CurrencyOf::<T>::make_free_balance_be(&deployer, value);
 
-  }: _(RawOrigin::Signed(deployer.clone()), program.clone(), configuration_schema.clone(), auxiliary_data_schema.clone(), oracle_data_pointer.clone())
+  }: _(RawOrigin::Signed(deployer.clone()), program.clone(), configuration_schema.clone(), auxiliary_data_schema.clone(), oracle_data_pointer.clone(), version_number)
```
HCastano, line 63:
```suggestion
  }: _(
      RawOrigin::Signed(deployer.clone()),
      program.clone(),
      configuration_schema.clone(),
      auxiliary_data_schema.clone(),
      oracle_data_pointer.clone(),
      version_number,
  )
```

HCastano (approved):
I think there's a better way to manage versions probably leaning into using the `programs-runtime` a bit more, but this is fine for now.

crates/shared/src/constants.rs
JesseAbram: https://github.com/entropyxyz/entropy-core/issues/1056 opened this

PR #1048, opened 2024-09-12
Fix hardcoded threshold amount

### LINKED ISSUE #898: Setting t and n for threshold signing
Currently on the `Migrate-to-threshold` branch n is set to the number of validators and t is a fixed proportion (75% rounding down) of n.  This will means if the number of validators changes, signing wont work.

We need:

- To have a way of setting t and n (eg: using the parameters pallet).
- If n is independent of the number of validators (which i guess it must be), we need a way to select which validators to use for DKG / signing.

See also https://github.com/entropyxyz/entropy-core/issues/859

JesseAbram:
Closes #898

### DIFF
--- a/crates/client/src/user.rs
+++ b/crates/client/src/user.rs
@@ -50,7 +50,7 @@ pub async fn get_signers_from_chain(
         .ok_or_else(|| SubgroupGetError::ChainFetch("Get all validators error"))?;
 
     let key_info_query = entropy::storage().parameters().signers_info();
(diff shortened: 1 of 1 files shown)


HCastano (approved):

ameba23:
@JesseAbram we should probably also change it here: 
https://github.com/entropyxyz/entropy-core/blob/815c35676e22c454de086032ea6c6ef9f0a1bd7c/crates/threshold-signature-server/src/helpers/user.rs#L93

Issue #1032, opened 2024-08-27
Run TSS tests in parallel


HCastano:
We currently run all, or at least most, of the TSS tests serially. This is done to avoid
having different tests write conflicting state to the same DBs, and to ensure each test
is using a "clean" environment.

After some of the refactoring work in #1026 where a few tests were split out, the run
time of the testing CI step went from ~45 mins to ~60 mins. This is way too long to give
developers feedback about their PR, and it also consumes more minutes on our runners (and
thus incurs extra costs).

I'm proposing that we look into running these tests in parallel by setting up an
independent network for each test. To do this we'd need to ensure that non-conflicting
ports are used for each test, as well as different database locations (for both the
Substrate nodes and TSS servers).

While we might be able to do this with the test infrastrcuture we already have, I think
it's worthwhile to look into using a combination of [zombienet](https://github.com/paritytech/zombienet) and [cargo-nextest](https://nexte.st/).

HCastano:
Related issue: https://github.com/entropyxyz/entropy-core/issues/487#issuecomment-1804126321

PR #1049, opened 2024-09-12
Run TSS tests in parallel

HCastano:
As mentioned in #1032, the runtime of our tests is now pushing an hour. This PR changes
how our tests are setup in order to allow tests to run in parallel instead of serially.

For example, now the following command runs in parallel:

```
cargo test -p entropy-tss --release -- signature_request_with_derived_account_works test_signature_requests_fail_on_different_conditions --nocapture

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 32 filtered out; finished in 139.16s
```

vs. running this on `master`

```
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 30 filtered out; finished in 240.54s
```

I'll add more details here once it's out of a draft state.

### DIFF
--- a/.github/workflows/build-and-run-node-test.yaml
+++ b/.github/workflows/build-and-run-node-test.yaml
@@ -33,7 +33,7 @@ jobs:
       - name: Run `cargo build && cargo test`
         run: |
           pushd node
-          cargo build -p entropy-tss --all-targets --release -j $(nproc)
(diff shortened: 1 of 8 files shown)


HCastano:
This ended up having issues when I was running it on the CI that I wasn't able to reproduce locally.

While there are some things that I'd maybe like to take out of this into other PRs, I this isn't high priority enough right now to continue.

Because of that I'm gonna close this.

PR #1051, opened 2024-09-13
Handle Provisioning Certification Keys (PCKs)

ameba23:
As part of the attestation feature https://github.com/entropyxyz/entropy-core/issues/982 we need to handle PCK certificates when a validator joins. These contain the public key used to sign attestations, and are signed by Intel.

Rather than storing the whole certificate, my current proposal is to verify it once with Intel's public key, and then only store the public key with which we need to verify quotes.

However it could be argued we should keep the whole certificate including the signature - i am open to discuss this.

This PR handles these public keys and uses them in quote verification, but it does not handle extracting them from a PCK certificate when a validator joins, or verifying that the certificate is signed by Intel. This will come in another PR.

PCKs are stored as a field in the staking extension pallet's `ServerInfo` struct.  This might not seem like the best home for them - i could have made a mapping of TSS account IDs to PCKs in the attestation pallet.  But i deliberately did it like this because i want it to be impossible that an active validator does not have an associated PCK.

When generating mock quotes, the PCK keypair is derived from the TSS account ID. This saves us having to generate and store them somewhere, and makes it easy to compute the keypairs for the test accounts.


pallets/staking/src/lib.rs
```diff
@@ -115,6 +115,7 @@ pub mod pallet {
         pub tss_account: AccountId,
         pub x25519_public_key: X25519PublicKey,
         pub endpoint: TssServerURL,
+        pub provisioning_certification_key: VerifyingKey,
```
ameba23, line 118: I don't like that this is a `BoundedVec` - it has caused all sorts of problems.  I would have preferred to use a `[u8; 33]`. But i tried and i had difficulties with serializing / encoding.  Just using `Vec<u8>` i find too permissive, so in the end i went with `BoundedVec`.

JesseAbram: like all bounded vec does is just put a limit on a vec, we can do that manually, it is just everytime we touch it we would need to make sure we apply a size check

crates/shared/src/types.rs
```diff
@@ -109,6 +109,10 @@ pub enum HashingAlgorithm {
 /// A compressed, serialized [synedrion::ecdsa::VerifyingKey<k256::Secp256k1>]
 pub type EncodedVerifyingKey = [u8; VERIFICATION_KEY_LENGTH as usize];
 
+#[cfg(not(feature = "wasm"))]
+pub type BoundedVecEncodedVerifyingKey =
```
JesseAbram, line 113: Im not sure I see this anywhere else

JesseAbram (approved):

pallets/attestation/src/lib.rs
```diff
@@ -175,8 +177,18 @@ pub mod pallet {
             let accepted_mrtd_values = pallet_parameters::Pallet::<T>::accepted_mrtd_values();
             ensure!(accepted_mrtd_values.contains(&mrtd_value), Error::<T>::BadMrtdValue);
 
-            // TODO #982 Check that the attestation public key matches that from PCK certificate
-            let _attestation_key = quote.attestation_key;
+            // Check that the attestation public key is signed with the PCK
+            let provisioning_certification_key = decode_verifying_key(
+                &server_info
+                    .provisioning_certification_key
```
HCastano, line 183: More a note to myself than anything, but this is gonna have to get abstracted out similar to the X25519 key in #1063

pallets/staking/Cargo.toml
```diff
@@ -17,6 +17,7 @@ scale-info ={ version="2.11", default-features=false, features=["derive"] }
 log        ={ version="0.4.22", default-features=false }
 serde      ={ version="1.0.210", default-features=false }
 rand_chacha={ version="0.3", default-features=false }
+p256       ={ version="0.13.2", default-features=false, features=["ecdsa", "alloc"] }
```
HCastano, line 20: Is this import used?

crates/testing-utils/src/helpers.rs
```diff
@@ -52,3 +54,18 @@ pub async fn spawn_tss_nodes_and_start_chain(
...
+    tdx_quote::VerifyingKey::from(pck)
+}
+
+/// For each test TSS account, display the encoded mock PCK
+pub fn print_test_pck_verifying_keys() {
+    for tss_account in crate::constants::TSS_ACCOUNTS.iter() {
+        let pck = derive_mock_pck_verifying_key(tss_account);
+        println!("{:?}", encode_verifying_key(&pck));
+    }
+}
```
HCastano, line 71: 👍

crates/shared/src/types.rs
HCastano: It's used in the `provisioning_certification_key` constants and chainspecs

HCastano (approved):
I don't totally understand your comments around using Intel's public key once to avoid storing the whole cert, but this might be a bigger conversation around the quote/attestation flow from my end

pallets/staking/Cargo.toml
ameba23: No - good catch

ameba23:
> I don't totally understand your comments around using Intel's public key once to avoid storing the whole cert, but this might be a bigger conversation around the quote/attestation flow from my end

@HCastano i think we should chat about this as i have some code for handling the certificates when a validator initially joins (or changes their endpoint later), which should probably link in with what you are doing in 1063.

Issue #1052, opened 2024-09-13
Selecting initial validators for jumpstart DKG


ameba23:
In trying to write an end-to-end test https://github.com/entropyxyz/entropy-core/issues/1010 i found there is an issue with how we select who performs the initial jumpstart DKG.

Currently the entire validator set is selected.  Which works great as long as there are n validators. If there are more than n, we need a way to select n to be the initial signer committee.

This issue is currently not picked up by our tests, because the jumpstart test uses the development chainspec, with 3 TSS nodes, and n = 3. If we try to run this test using the integration test chainspec with 4 TSS nodes, the test fails.

In order to do a full test with both a jump start and a reshare, we need 4 TSS nodes.

This means we need a way to select n initial nodes from the validator set.  We could do this deterministically using block number. This means that whoever submits the jump start extrinsic can choose who these validators are by waiting until a particular block.

Other options would be: 
- To use the substrate randomness API - but we then need to store the chosen validators so that the TS servers can check it. 
- To automatically begin the jumpstart as soon as the nth validator joins.

ameba23:
I made a draft PR where we select validators using block number - im not sure if it is the best solution but it at least serves to demonstrate what i am talking about.

PR #1053, opened 2024-09-13
Select validators for jumpstart DKG

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1052

This selects n validators for initial jumpstart DKG when handling the jumpstart extrinsic, using on-chain randomness.

I've also removed the signature request account field from some types relating to DKG, since DKG is now only used for network jumpstart.


crates/threshold-signature-server/src/user/api.rs
```diff
@@ -300,6 +307,55 @@ pub async fn generate_network_key(
...
+        let validators_query = entropy::storage().session().validators();
+        query_chain(api, rpc, validators_query, None).await?.ok_or_else(|| {
+            UserErr::OptionUnwrapError("Cannot get validators info from session pallet".to_string())
+        })?
+    };
+
+    // Select n validators
+    let mut rng = rand::rngs::StdRng::seed_from_u64(block_number.into());
+    let selected_validators: Vec<_> =
+        all_validators.choose_multiple(&mut rng, n.into()).cloned().collect();
```
ameba23, line 339: This is inspired by how @JesseAbram was selecting signers in the relay PR.  Here it probably is not much better than using modulo, but i think it makes the code a bit neater.

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -282,15 +282,18 @@ pub async fn jump_start_network_with_signer(
...
+        vec![ValidatorName::Alice, ValidatorName::Bob, ValidatorName::Charlie, ValidatorName::Dave];
     for validator_name in validators_names {
         let mnemonic = development_mnemonic(&Some(validator_name));
         let (tss_signer, _static_secret) =
             get_signer_and_x25519_secret_from_mnemonic(&mnemonic.to_string()).unwrap();
         let jump_start_confirm_request =
             entropy::tx().registry().confirm_jump_start(BoundedVec(EVE_VERIFYING_KEY.to_vec()));
 
-        submit_transaction(api, rpc, &tss_signer, &jump_start_confirm_request, None).await.unwrap();
+        // Ignore the error as one confirmation will fail
```
ameba23, line 294: Here we don't know which validators are going to be picked, but we want all those who are picked to send confirmations.  Since nothing bad happens if the wrong one confirms, we just have them all send confirmations here.

pallets/registry/src/lib.rs
```diff
@@ -246,14 +268,20 @@ pub mod pallet {
...
             let validator_stash =
                 pallet_staking_extension::Pallet::<T>::threshold_to_stash(&ts_server_account)
                     .ok_or(Error::<T>::NoThresholdKey)?;
-            let validators = pallet_session::Pallet::<T>::validators();
-            ensure!(validators.contains(&validator_stash), Error::<T>::NotValidator);
+
+            // Check that the confirmation is coming from one of the selected validators
+            let (_block_number, selected_validators) =
+                JumpstartDkg::<T>::iter().last().ok_or(Error::<T>::JumpStartNotInProgress)?;
+            let selected_validators: Vec<_> =
```
HCastano, line 279: Are you sure this is consistently giving you the entry with the latest block? From the docs it says that iterating on a map doesn't guarantee any order

pallets/registry/src/tests.rs
```diff
@@ -301,7 +300,23 @@ fn it_jumps_the_network() {
         assert_ok!(Registry::jump_start_network(RuntimeOrigin::signed(1)));
         assert_eq!(
             Registry::jumpstart_dkg(0),
-            vec![NETWORK_PARENT_KEY.encode()],
+            vec![
```
HCastano, line 303: Do you mind making a note that these come from the mock genesis config?

pallets/registry/src/benchmarking.rs
```diff
@@ -102,6 +103,15 @@ benchmarks! {
...
       });
 
+    // Add the jump start record
+    let block_number = <frame_system::Pallet<T>>::block_number();
+    <JumpstartDkg<T>>::set(block_number, vec![
+        ValidatorInfo {
+            x25519_public_key: [0; 32],
+            ip_address: vec![20],
+            tss_account: accounts[1].encode(),
+        }
```
HCastano, line 113: Since the size of this list can vary depending on the number of validators, we should probably take this into account during the benches

HCastano (commented):
Only thing I'm worried about here is the order when reading from a storage map, but otherwise looks good

pallets/registry/src/benchmarking.rs
ameba23: Ok - so the thing to do would be add a few more validators to this vec. Without wanting to get into discussion of how big n will be in production, I would just make it whatever n is set to by default in parameters for now.

pallets/registry/src/lib.rs
ameba23: From the docs of the `iter()` function from `IterableStorageMap` which is a trait that `StorageMap` implements:

'Enumerate all elements in the map in lexicographical order of the encoded key'
https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/trait.IterableStorageMap.html#tymethod.iter

So the question is whether lexicographically sorting scale encoded block numbers is the same as sorting them numerically. I think block numbers become a U256, but not sure about the encoding of a U256.  I could try to write a test to figure it out.

ameba23: @HCastano so in the end - good catch!

I wrote a test and iterating over block numbers in a `StorageMap` gives them in what appears to be arbitrary order despite them being sorted after being encoded, so no idea what encoding is being used. So i've changed this to use a comparison on the decoded key. I've left the test in, not sure if we really need it.

pallets/registry/src/benchmarking.rs
HCastano: It makes sense in the context of the benches, but I think the benches are actually parameterized wrong.

@JesseAbram doesn't look like we're using `c` in these benches, only the `MAX_SIGNERS`. Maybe something to follow up on?

HCastano (approved):
Thanks for the fixes!

pallets/registry/src/benchmarking.rs
```diff
@@ -117,12 +126,22 @@ benchmarks! {
     let threshold_account: T::AccountId = whitelisted_caller();
     let expected_verifying_key = BoundedVec::default();
 
-    // add validators and a registering user
+    // add validators
     for i in 0..MAX_SIGNERS {
         let validators = add_non_syncing_validators::<T>(MAX_SIGNERS as u32, 0);
```
ameba23, line 131: Not relevant to this PR, but while we are on the topic of whether stuff in this benchmark makes sense:
I am wondering why this line is inside the for loop and not before it.
Why do we need to call `add_non_syncing_validators` 15 times?

Issue #1069, opened 2024-09-26
Unignore integration tests


JesseAbram:
In #1061 I ignored the integration tests. It was after a convo with @ameba23 where it came down to fixing them to not need unsafe or test helpers was needed but was too much for the current PR (and not immediately needed) 

anyways @ameba23 had fixing that on his todo list anyways, so I ignored it until then

PR #1070, opened 2024-09-26
Unignore register and sign integration test, and do a non-mock jumpstart

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1069


JesseAbram (approved):
looks good, hmmmm I wonder if this would also fix #1071 .......I mean it would, but is it worth it time wise

Issue #1071, opened 2024-09-26
Handle test key generation better


JesseAbram:
In #1050 we were having trouble with key managment, since there is now randomness and 4 potential signers and 3 are selected. This means we can't create alice, bob, charlie keys anymore and assume those three will be singing. In 1050 we......I......mmm me and @ameba23 made it alice, bob, and dave, which holds true due to test randomness with the chain, however, it needs to be more flexible.

PR #1073, opened 2024-09-27
Pregenerate keyshares sets for all possible initial signer comittees

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1071

Since keyshares are tied to particular party IDs, we cannot give alice's keyshare to bob and expect things to work.

This is an issue now that the initial signer set is chosen randomly - it makes it hard to use the pre-generated keyshares which are already assigned to partiular party IDs (which are TSS account IDs).

One option would be to generate the keyshares at the point when which we know who the signers are.  But this would slow down our tests quite a bit (although since we use a centralized generation it would still be faster than doing a DKG).

So my proposed solution is to pre-generate a set of keyshares for each possible initial signer set.

Since we have 4 nodes, and n=3, we can refer to each set by the name of the node who is **not** in the signer set (which is the one who will act as the relayer node).

So set 'alice' consists of ['bob', 'charlie', 'dave'] and set 'bob' consists of ['alice', 'charlie', dave'], etc.


crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -74,6 +74,12 @@ pub enum ValidatorName {
     Eve,
 }
 
+impl std::fmt::Display for ValidatorName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", format!("{:?}", self).to_lowercase())
```
ameba23, line 79: This is used for putting the validator names in filepaths - we just take the validator name as lower case

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -307,14 +297,21 @@ pub async fn jump_start_network_with_signer(
...
         }
     }
+    if let Some(non_signer) = non_signer {
+        for validator_name in validators_names {
+            if non_signer != validator_name {
+                put_keyshares_in_db(non_signer, validator_name).await;
+            }
+        }
+    } else {
+        panic!("No non signer");
```
ameba23, line 312: I guess since we panic here, it does not make sense to return an option - as the option will always be `Some`. The alternative would be to not panic and just not put any keyshares. :shrug:

JesseAbram: can we just ignore it and return none?

JesseAbram (changes requested):
looks good just definitely needs a good readme with how to run and why etc etc

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -307,14 +297,21 @@ pub async fn jump_start_network_with_signer(
...
         }
     }
+    if let Some(non_signer) = non_signer {
+        for validator_name in validators_names {
+            if non_signer != validator_name {
+                put_keyshares_in_db(non_signer, validator_name).await;
+            }
+        }
+    } else {
+        tracing::error!("No non signer - not storing pre-generated keyshares");
```
HCastano, line 312:
```suggestion
        tracing::error!("Missing non-signer - not storing pre-generated keyshares");
```

scripts/create-test-keyshares/src/main.rs
```diff
@@ -17,60 +17,78 @@
...
-            bob_pair.signer().clone(),
-            charlie_pair.signer().clone(),
-        )
-        .await;
-        let test_keyshres_serialized =
+    for (_keypair, name) in keypairs_and_names.iter() {
+        let (keypairs_this_time, names_this_time): (Vec<sr25519::Pair>, Vec<ValidatorName>) =
+            keypairs_and_names.iter().filter(|(_, n)| n != name).cloned().unzip();
+
+        // We cant do .try_into().unwrap() here becasue sr25519::Pair doesn't implement Debug
```
HCastano, line 56: We can map the error to something that does, e.g `.map_err(|_| ())`

HCastano (approved):
Something I'd want to look into in the future is not having 50 build artifacts as part of the repo - so maybe generating these once either at build-time or test-build-time. 

I'm thinking that its a similar thing to a lockfile. These can be generated so they don't necessarily need to be version controlled in all cases.

scripts/create-test-keyshares/src/main.rs
ameba23: Like it, thanks!

ameba23:
@JesseAbram i can't merge this until you re-review

JesseAbram (approved):

PR #1074, opened 2024-09-27
Run multiple test validator

JesseAbram:
Reduces testing time significantly as blocks all come in, in 6 seconds instead of being skipped. 

Can push this farther by adding this to the other substrate spin up methods (there are few and it may not have as added an effect)

### DIFF
--- a/crates/shared/src/constants.rs
+++ b/crates/shared/src/constants.rs
@@ -75,7 +75,7 @@ pub const MAX_SIGNERS: u8 = 15;
 pub const SIGNER_THRESHOLD: u8 = 2;
 
 /// For testing to line up chain mock data and reshare_test
-pub const TEST_RESHARE_BLOCK_NUMBER: u32 = 5;
+pub const TEST_RESHARE_BLOCK_NUMBER: u32 = 7;
 
 /// Program version number, must be incremented if version number changes
 pub const PROGRAM_VERSION_NUMBER: u8 = 0;
--- a/crates/testing-utils/src/substrate_context.rs
+++ b/crates/testing-utils/src/substrate_context.rs
@@ -18,8 +18,6 @@ use subxt::{config::substrate::SubstrateExtrinsicParams, OnlineClient};
 
 use super::node_proc::TestNodeProcess;
 use crate::chain_api::*;
-use std::time::{Duration, self};
-use std::thread;
 /// Verifies that the Entropy node binary exists.
 ///
 /// If a path is provided using the `ENTROPY_NODE` environment variable, that will take priority.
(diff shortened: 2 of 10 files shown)


crates/testing-utils/src/substrate_context.rs
```diff
@@ -62,48 +61,96 @@ pub async fn test_node_process_with(
...
     force_authoring: bool,
-) -> TestNodeProcess<EntropyConfig> {
-    test_node(AccountKeyring::Alice, "--chain=integration-tests".to_string(), force_authoring).await
+) -> Vec<TestNodeProcess<EntropyConfig>> {
+    let alice_bootnode = Some(
+        "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWM7EoKJfwgzAR1nAVmYRuuFq2f3GpJPLrdfhQaRsKjn38"
+            .to_string(),
+    );
+    // reduses message from chain to same TSS cleaning up a lot of logging
+    let fuck_off_tss_ip = Some("127.0.0.1:4010".to_string());
```
JesseAbram, line 120: please let me keep this variable name

ameba23: Why not make this be the actual TS server endpoint of the associated TS servers?  Wouldn't this greatly improve our tests as we would no longer need to mock the offchain worker messages?

crates/shared/src/constants.rs
```diff
@@ -75,7 +75,7 @@ pub const MAX_SIGNERS: u8 = 15;
 pub const SIGNER_THRESHOLD: u8 = 2;
 
 /// For testing to line up chain mock data and reshare_test
-pub const TEST_RESHARE_BLOCK_NUMBER: u32 = 5;
+pub const TEST_RESHARE_BLOCK_NUMBER: u32 = 7;
```
ameba23, line 78: Why did this need to change? I am wondering if there is maybe still something not quite right with the how this block number is handled.

ameba23 (approved):
Great that this speed our tests up considerably - although i was expecting that it was only adding one extra chain node to get the 20min speedup, not three.

I would really like it if we could use this to no longer need to mock the propagation pallet messages to bob, charlie and dave's TS servers, as i think that is a weak area in our test setup.

crates/testing-utils/src/substrate_context.rs
JesseAbram: so one thing we do in the tests is mock these calls then get the returns and check them, if we sent it straight to the TSS from the chain, then we could not check the return errors and we would have to get hella creative with the tests and re write them to make sure the goals were accomplished. I do not think this makes the test better but if we agree it does then it would be a hella test re write and outside the scope of this PR

PR #1077, opened 2024-09-30
Add integration test with DKG, reshare, register, sign

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1010

I had to bump `entropy_shared::constants::TEST_RESHARE_BLOCK_NUMBER` from 5 to 7 to get this to work when signing a message before the reshare - because the added time to sign a message meant that the block number was already greater than 5 by the time the reshare happened. @JesseAbram is anyway bumping that constant to 7 in another PR.

Personally i think having this test means we can remove [`integration_test_register_and_sign`](https://github.com/entropyxyz/entropy-core/blob/master/crates/threshold-signature-server/tests/register_and_sign.rs), since everything tested there is also tested here.  But happy to leave it in if others think its useful.


JesseAbram:
> Personally i think having this test means we can remove [`integration_test_register_and_sign`](https://github.com/entropyxyz/entropy-core/blob/master/crates/threshold-signature-server/tests/register_and_sign.rs), since everything tested there is also tested here. But happy to leave it in if others think its useful.

nah let's get rid of it, if we already test it then ya let's shrink our testing time

JesseAbram (approved):

ameba23:
The reshare test is failing - which is unrelated to this PR. And in the last commit the jumpstart test failed, also unrelated to this PR.  I am gonna re-run CI.  But if this keeps happening i'll make an issue.

HCastano (approved):
Nice nice.

Ya the CI failure is unrelated to your PR, Jesse is working on a fix now

ameba23:
This is still failing after the latest fixes, but the reshare test passes.  The two differences between this test and the reshare test are that this one does a DKG jumpstart, whereas the reshare test uses the pre-generated keyshares, and this one signs a message before doing the reshare, which takes some time.

crates/threshold-signature-server/tests/register_sign_reshare_sign.rs
```diff
@@ -0,0 +1,220 @@
...
+
+    let block_number = TEST_RESHARE_BLOCK_NUMBER;
+    let onchain_reshare_request =
+        OcwMessageReshare { new_signer: new_signer.0.to_vec(), block_number };
+
+    run_to_block(&rpc, block_number + 1).await;
+    // Send the OCW message to all TS servers who don't have a chain node
+    let client = reqwest::Client::new();
+    let response_results = join_all(
+        [3002, 3003, 3004]
```
JesseAbram, line 172: @ameba23 hardcoding these values does not seem right as one will probably fail as only 3 validators have the key

JesseAbram: looking into it more an increase of TEST_RESHARE_BLOCK_NUMBER should fix it, and this logically makes sense but since we get the info above probably should unhardcode this

Issue #1072, opened 2024-09-27
Signer changing TSS account 


JesseAbram:
Currently there is a btree map for tss_account as the key when signing in synedrion. There is also a way for a TSS to change their TSS [account](https://github.com/entropyxyz/entropy-core/blob/master/pallets/staking/src/lib.rs#L373) . If they are a signer and do this, this will break signing with them. This can be fixed if everyone updates their TSS key for them. Either a pre lookup to get their tss key and check and change them in the key share before signing if this is possible @fjarri , if so that is cheap and can happen wtv. If not then maybe a reshare triggered with a new TSS...... or the same one but if they do it too much then maybe slash @HCastano

ameba23:
The problem with changing the account id in the keyshare, is that all other keyshares in the set have a list of the other party ids - so we would have to change it in all the keyshares.

The other problem is that if a validator want to change their TSS account ID because they are moving the physical location of where their TS server is run, they would anyway have no way of bringing the keyshare with them.

I think the simplest solution would be to not allow chainging TSS account ID whilst being a signer.

Another option would be that changing TSS account ID triggers a reshare, replacing the keyshare associated with the old TS account with the new one.

JesseAbram:
yes both those are options, however, why can't before sign I do a lookup of all signers on the chain, get their TSS accounts, check it against the TSS accounts in the keyshare, and replace the ones that are not there with the new ones?

PR #1078, opened 2024-09-30
Block tss chain when signer

JesseAbram:
Related #1072 

Fixes 1072 but maybe not the best UX, open to leave the issue open and changes it, but it closes the vulnerability and was hella fast to implement


ameba23 (approved):
I think this is a good move.  Maybe we need to do the same for changing http endpoint.

JesseAbram:
> I think this is a good move. Maybe we need to do the same for changing http endpoint.

I think I see your point here with tdx, but there may be cases where you can, for example port forwarding, but it brings up a good point, redesigning these flows with TDX in ming

PR #1079, opened 2024-09-30
Fix master build

JesseAbram:
Fixes master build by not querying potentially missed event then repinging until the event block had been pruned, this instead just goes to non pruning state of the chain

### DIFF
--- a/crates/threshold-signature-server/src/helpers/tests.rs
+++ b/crates/threshold-signature-server/src/helpers/tests.rs
@@ -21,7 +21,10 @@
 use crate::{
     app,
     chain_api::{
-        entropy::{self, runtime_types::bounded_collections::bounded_vec::BoundedVec},
+        entropy::{
+            self, runtime_types::bounded_collections::bounded_vec::BoundedVec,
+            runtime_types::pallet_staking_extension::pallet::JumpStartStatus,
+        },
         EntropyConfig,
     },
     get_signer,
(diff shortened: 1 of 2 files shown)


crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -389,23 +392,31 @@ pub async fn do_jump_start(
...
-    for _ in 0..75 {
-        std::thread::sleep(std::time::Duration::from_millis(1000));
-        let block_hash = rpc.chain_get_block_hash(None).await.unwrap();
-        let events = EventsClient::new(api.clone()).at(block_hash.unwrap()).await.unwrap();
-        let jump_start_event = events.find::<entropy::registry::events::FinishedNetworkJumpStart>();
-        if let Some(_event) = jump_start_event.flatten().next() {
-            got_jumpstart_event = true;
-            break;
-        };
+    assert_eq!(format!("{:?}", jump_start_status), format!("{:?}", JumpStartStatus::Done));
```
ameba23, line 416: This will always be true, but no harm in leaving it in

ameba23 (approved):
:1st_place_medal: Great that this works and CI is passing again.

I don't totally understand why it works. I am guessing it must be that sometimes the jumpstart finished event was happening whilst we were waiting for the http responses, and so we were missing it.

```run_to_block(&rpc, block_number + 9).await;```
This feels like something which could break again easily when something changes either with the protocol or the hardware we run the tests on.

I feel like we keep getting problems in tests relating to mocking the OCW requests (eg the `test_store_share` test).  I know you disagree but i think the way to go is to only mock these when you are specifically testing the behavior of that http endpoint.

JesseAbram:
> 🥇 Great that this works and CI is passing again.
> 
> I don't totally understand why it works. I am guessing it must be that sometimes the jumpstart finished event was happening whilst we were waiting for the http responses, and so we were missing it.
> 
> `run_to_block(&rpc, block_number + 9).await;` This feels like something which could break again easily when something changes either with the protocol or the hardware we run the tests on.
> 
this should be fine this was because we.....I upped the refresh sent block from 5 to 7, so this created a race condition, now it should be good. 

> I feel like we keep getting problems in tests relating to mocking the OCW requests (eg the `test_store_share` test). I know you disagree but i think the way to go is to only mock these when you are specifically testing the behavior of that http endpoint.

mmmm fair we can bring it up on our next core meeting,

PR #1083, opened 2024-10-01
Avoid panic by checking that we have a non-signing validator before selecting one

ameba23:
Ahead of us making a release i wanted to try out our current setup with docker-compose. I should have known it was not going to work with the relaying because we have 3 nodes and all are signers - so there is no-one to be a relayer for our signature requests.

This is what happens when we try to sign:
```
~/r/s/e/e/c/test-cli (master●●)[130] $ cargo run --release -- sign -m //Alice 0380b55a941d514baf9e5c4b96239cda40c1854f83919262029bdfec340aed4c13 'heysdlafkjsdlfkj'
    Finished `release` profile [optimized] target(s) in 0.51s
     Running `/home/turnip/radish/src/entropy/entropy-core/target/release/entropy-test-cli sign -m //Alice 0380b55a941d514baf9e5c4b96239cda40c1854f83919262029bdfec340aed4c13 heysdlafkjsdlfkj`
User account for current call: 5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY
thread 'main' panicked at /home/turnip/.cargo/registry/src/index.crates.io-6f17d22bba15001f/rand-0.8.5/src/rng.rs:134:9:
cannot sample empty range
```

`entropy-client::sign` panics, rather than returning an error.  This fixes that.

### DIFF
--- a/crates/client/src/errors.rs
+++ b/crates/client/src/errors.rs
@@ -96,8 +96,6 @@ pub enum ClientError {
     TryFromSlice(#[from] std::array::TryFromSliceError),
     #[error("User is not registered on-chain")]
     NotRegistered,
-    #[error("No synced validators")]
-    NoSyncedValidators,
     #[error("Cannot confirm program was created")]
     CannotConfirmProgramCreated,
     #[error("Cannot confirm program was removed")]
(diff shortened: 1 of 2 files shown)


crates/client/src/errors.rs
```diff
@@ -96,8 +96,6 @@ pub enum ClientError {
     TryFromSlice(#[from] std::array::TryFromSliceError),
     #[error("User is not registered on-chain")]
     NotRegistered,
-    #[error("No synced validators")]
-    NoSyncedValidators,
```
ameba23, line 100: Unrelated to this PR but i think we don't use this anymore

HCastano (approved):

PR #1086, opened 2024-10-01
Add fourth node to `devnet-local` configuration

ameba23:
It is no longer possible to sign a message with our docker-compose setup as signing requires 4 nodes - 3 signers and a relayer node.

This adds a fourth node, 'dave', to to the docker-compose setup and makes the `devnet-local` chainspec have four nodes.

This is mostly copied from https://github.com/entropyxyz/entropy-core/pull/940 which added a third node.


ameba23:
I can confirm i have tried this an successfully jumpstarted, registered and signed a message using the test-cli

JesseAbram (approved):

node/cli/src/chain_spec/dev.rs
```diff
@@ -86,46 +93,20 @@ pub fn devnet_local_docker_three_node_initial_tss_servers(
...
-
     let dave = (
         crate::chain_spec::tss_account_id::DAVE.clone(),
         crate::chain_spec::tss_x25519_public_key::DAVE,
-        "dave-tss-server:3003".to_string(),
+        "dave-tss-server:3004".to_string(),
         provisioning_certification_key::DAVE.clone(),
     );
 
-    let eve = (
```
HCastano, line 115: So I'm pretty sure `eve` was used here instead of `charlie` because the SDK was using the Charlie account for testing.

I like Charlie over Eve here, but it may cause problems for @entropyxyz/js-devs.

HCastano (approved):
Cool cool 👍 

I added a fourth node to the `--dev` config as well.

node/cli/src/chain_spec/dev.rs
ameba23: Hmm, ok i wasn't aware of that.  The reason i changed it to charlie is because we use Alice, Bob, Charlie, Dave as the 4 nodes in our tests.

HCastano: I made a note of the change in the `CHANGELOG`

PR #1092, opened 2024-10-02
Minor fixes to entropy-client needed for updating the status page

ameba23:
Related PR: https://github.com/entropyxyz/entropy-network-status-page/pull/21

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -47,10 +47,12 @@ use base64::prelude::{Engine, BASE64_STANDARD};
 use entropy_protocol::RecoverableSignature;
 use entropy_shared::HashingAlgorithm;
 use futures::stream::StreamExt;
-use sp_core::{sr25519, Pair};
+use sp_core::{
+    sr25519::{self, Signature},
+    Pair,
+};
 use subxt::{
     backend::legacy::LegacyRpcMethods,
-    ext::sp_core::sr25519::Signature,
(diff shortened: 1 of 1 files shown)


crates/client/src/client.rs
```diff
@@ -47,10 +47,12 @@ use base64::prelude::{Engine, BASE64_STANDARD};
...
 use entropy_shared::HashingAlgorithm;
 use futures::stream::StreamExt;
-use sp_core::{sr25519, Pair};
+use sp_core::{
+    sr25519::{self, Signature},
+    Pair,
+};
 use subxt::{
     backend::legacy::LegacyRpcMethods,
-    ext::sp_core::sr25519::Signature,
```
ameba23, line 53: This is only exposed with the `substrate-compat` feature which isn't necessarily enabled - so rather use this from sp_core.

crates/client/src/client.rs
```diff
@@ -327,7 +329,7 @@ pub async fn put_register_request_on_chain(
     let registered_event = registered_events
         .find::<entropy::registry::events::AccountRegistered>()
         .flatten()
-        .find_map(|event| (event.0 == signature_request_keypair.public().into()).then_some(event))
+        .find_map(|event| (event.0 == signature_request_keypair.public().0.into()).then_some(event))
```
ameba23, line 332: For some reason the traits needed to convert this directly are not available with the features enabled when we use this crate for the status page.

HCastano (approved):

PR #1095, opened 2024-10-02
Remove bad clippy ignores

JesseAbram:
I think the title says it all


ameba23 (approved):
omg i had been wondering why sometimes unused imports were getting ignored, but i never thought it could be something so simple.  Great bit of tidying.

PR #1097, opened 2024-10-03
Check current signers still validators

JesseAbram:
Related #1114


crates/protocol/src/execute_protocol.rs
```diff
@@ -334,6 +334,7 @@ pub async fn execute_reshare(
     chans: Channels,
     threshold_pair: &sr25519::Pair,
     inputs: KeyResharingInputs<KeyParams, PartyId>,
+    verifiers: &BTreeSet<PartyId>,
```
ameba23, line 337: im not sure we need this argument as it looks like whenever we call this function `verifyiers` is the same as `inputs.new_holders`

ameba23 (approved):
:star: this is great.

Is it possible to further improve this by having the parties who are leaving the signing set also participate in the reshare, but not receive new shares? So that we don't have to keep t parties the same. I can't remember whether synderion lets you do this.

pallets/staking/src/lib.rs
```diff
@@ -716,15 +719,39 @@ pub mod pallet {
...
+                    }
+                }
+            }
 
-            if current_signers_length <= signers_info.total_signers as usize {
+            while current_signers.len() < signers_info.total_signers as usize {
                 let mut randomness = Self::get_randomness();
                 // grab a current signer to initiate value
-                let mut next_signer_up = &current_signers[0].clone();
+                let mut next_signer_up = &validators[0].clone();
```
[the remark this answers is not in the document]
JesseAbram: just a place holder for rust not complaining, gets changed lower down (not initialized compile error)

JesseAbram:
> ⭐ this is great.
> 
> Is it possible to further improve this by having the parties who are leaving the signing set also participate in the reshare, but not receive new shares? So that we don't have to keep t parties the same. I can't remember whether synderion lets you do this.

not currently opened an issue, a limitation with synedrion see #1114

crates/protocol/src/execute_protocol.rs
JesseAbram: yes originally I changed it to allow for old holders to be more and not in new holders however ran into the #1114 issue, didn't change it back because this is still more right imo and adds no overhead, and hopefully will be needed when the issue is resolved

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -273,8 +275,15 @@ pub async fn validate_new_reshare(
         .await?
         .ok_or_else(|| ValidatorErr::ChainFetch("Not Currently in a reshare"))?;
 
-    if reshare_data.new_signer != chain_data.new_signer
-        || chain_data.block_number != reshare_data.block_number
+    let mut hasher_chain_data = Blake2s256::new();
```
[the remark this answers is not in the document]
JesseAbram: yes you are right a holdover of a change I revereted from #1114 but will change back

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -169,7 +170,8 @@ pub async fn new_reshare(
     .await?;
 
     let (new_key_share, aux_info) =
-        execute_reshare(session_id.clone(), channels, signer.signer(), inputs, None).await?;
+        execute_reshare(session_id.clone(), channels, signer.signer(), inputs, &new_holders, None)
```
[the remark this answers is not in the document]
JesseAbram: explained lower down

pallets/staking/src/lib.rs
```diff
@@ -716,15 +719,39 @@ pub mod pallet {
...
+                        remove_indexs.push(i);
+                    }
+                }
+                if remove_indexs.is_empty() {
+                    current_signers.remove(0);
+                } else {
+                    remove_index_len = remove_indexs.len();
+                    let remove_indexs_reversed: Vec<_> = remove_indexs.iter().rev().collect();
+                    // TODO: Only remove up to threhsold https://github.com/entropyxyz/entropy-core/issues/1114
+                    let truncated = if remove_index_len > signers_info.threshold as usize {
```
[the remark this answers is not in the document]
JesseAbram: yes, writing a better test for regression now

pallets/staking/src/lib.rs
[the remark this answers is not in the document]
ameba23: it looks to me like it only gets changed lower down if `validators[0]` is currently a  member of the signer set

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -113,13 +113,15 @@ async fn test_reshare() {
     let new_signer = all_validators.iter().find(|v| !signer_stash_accounts.contains(v)).unwrap();
 
     let block_number = TEST_RESHARE_BLOCK_NUMBER;
-    let onchain_reshare_request =
-        OcwMessageReshare { new_signer: new_signer.0.to_vec(), block_number };
+    let onchain_reshare_request = OcwMessageReshare {
+        new_signers: vec![new_signer.0.to_vec()],
+        block_number: block_number - 1,
```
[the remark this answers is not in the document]
JesseAbram: because this needs to match the onchain info or it will not pass the validate check

pallets/staking/src/lib.rs
```diff
@@ -695,15 +698,36 @@ pub mod pallet {
...
+                } else {
+                    remove_index_len = remove_indexs.len();
+                    let remove_indexs_reversed: Vec<_> = remove_indexs.iter().rev().collect();
+                    let truncated = remove_indexs_reversed
+                        [..(signers_info.total_signers as usize - signers_info.threshold as usize)]
+                        .to_vec();
+                    for remove_index in truncated {
+                        current_signers.remove(*remove_index);
+                    }
+                }
```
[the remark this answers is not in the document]
JesseAbram: commented it up more, for what it is worth the truncated should eventually be removed

crates/shared/src/types.rs
```diff
@@ -55,8 +55,8 @@ pub struct OcwMessageDkg {
 #[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 #[derive(Clone, Encode, Decode, Debug, Eq, PartialEq, TypeInfo)]
 pub struct OcwMessageReshare {
-    // Stash address of new signer
-    pub new_signer: Vec<u8>,
+    // Stash addresses of new signers
+    pub new_signers: Vec<Vec<u8>>,
```
dvdplm, line 59: A few lines above, in `ValidatorInfo`, `Vec`s are provided with the full path, `codec::alloc::vec::Vec<u8>`. Should we be consistent? The short form here seems better.

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -116,14 +116,15 @@ pub async fn new_reshare(
...
+    // also acts as verifiers as is everyone in the party
+    let new_holders: BTreeSet<PartyId> =
         validators_info.iter().cloned().map(|x| PartyId::new(x.tss_account)).collect();
-
-    let pruned_old_holders =
-        prune_old_holders(&api, &rpc, data.new_signer, validators_info.clone()).await?;
-
+    // old holders -> next_signers - new_signers (will be at least t)
+    let old_holders =
+        &prune_old_holders(&api, &rpc, data.new_signers, validators_info.clone()).await?;
```
dvdplm, line 125: I can't find where the `(will be at least t)` invariant is checked; doesn't seem to happen in `prune_old_holders` (also, but unrelated to this PR, that method seems to be cloning wildly but not sure that's needed).

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -273,8 +274,8 @@ pub async fn validate_new_reshare(
         .await?
         .ok_or_else(|| ValidatorErr::ChainFetch("Not Currently in a reshare"))?;
 
-    if reshare_data.new_signer != chain_data.new_signer
-        || chain_data.block_number != reshare_data.block_number
+    if chain_data.new_signers != reshare_data.new_signers
+        || chain_data.block_number != reshare_data.block_number.saturating_sub(1)
```
dvdplm, line 278: Comparing the two `new_signers` collections is probably more expensive than the block number check, so it's faster to flip them around.
```suggestion
    if chain_data.block_number != reshare_data.block_number.saturating_sub(1)
        || chain_data.new_signers != reshare_data.new_signers
```

crates/threshold-signature-server/tests/register_sign_reshare_sign.rs
```diff
@@ -153,19 +153,16 @@ async fn do_reshare(api: &OnlineClient<EntropyConfig>, rpc: &LegacyRpcMethods<En
...
-    let new_signer = all_validators.iter().find(|v| !signer_stash_accounts.contains(v)).unwrap();
+    let reshare_data_query = entropy::storage().staking_extension().reshare_data();
+    let reshare_data = query_chain(&api, &rpc, reshare_data_query, None).await.unwrap().unwrap();
 
     let block_number = TEST_RESHARE_BLOCK_NUMBER;
-    let onchain_reshare_request =
-        OcwMessageReshare { new_signer: new_signer.0.to_vec(), block_number };
+    let onchain_reshare_request = OcwMessageReshare {
+        new_signers: reshare_data.new_signers.into_iter().map(|s| s.to_vec()).collect(),
+        block_number: block_number - 1,
```
dvdplm, line 162: Does this blow up on block 0? Maybe `checked_sub`?

JesseAbram: the block number is hard coded one line above to a number way over 0

crates/threshold-signature-server/src/validator/api.rs
JesseAbram: ya this is a little hard to follow flow this comes from the chain itself here https://github.com/entropyxyz/entropy-core/pull/1097/files#diff-9be73ee9078236e99d5b726b1a2d9acd38446cedbe770d2f23d39ef51b1763ceR720

PR #1098, opened 2024-10-03
Release `v0.3.0-rc.1`

HCastano:
This is the release candidate for the `0.3.0` release of Core.

This release has a large portion of the non-TDX specific changes outlined in the
[Tofino spec](https://hackmd.io/@entropyxyz/H1cyIpSgR). More TDX specific features will come in `0.4.0`.

Notably, this includes changes like:
- Fully on-chain registration
- t-of-N signing instead of N-of-N signing
- The addition of the relayer role
- The removal of signing subgroups
- The removal of permissioned access mode

For more details take a look through the `CHANGELOG`.


runtime/src/weights/frame_benchmarking.rs
```diff
@@ -30,8 +30,8 @@
 // --wasm-execution=compiled
 // --pallet=frame_benchmarking
 // --extrinsic=*
-// --steps=50
-// --repeat=20
+// --steps=25
+// --repeat=10
```
HCastano, line 34: I ran the benchmarks with half the steps and repeat values to save time. Spot checking the bench results it doesn't look like there's much of a difference since the last release.

JesseAbram (approved):

PR #1106, opened 2024-10-08
Single line for release checklist command

vitropy:
The backslash, when treated literally, will cause the `entropy-test-cli` to fail to load its subcommand. In Markdown, this should either have been written as:

```
the command \
    with a new line
```

or just remove the slash and leave the single-line code formatting. I propose this simpler, latter option.


HCastano (approved):
I prefer the split line approach, but in this case since it's not too long a line it's fine.

Can you also update the signature command below to match?

Thanks!

PR #1123, opened 2024-10-21
Add quote guards to `ServerInfo` related extrinsics

HCastano:
This is one of the follow-ups to #1109, in which we will enforce that validators who want
to change certain info about their validator will need to pass an attestation check
first.

For the `change_endpoint()` extrinsic we assume that that the underlying hardware and
Account ID stay the same, so a quote from the existing TSS Account ID is expected.

For the `change_threshold_accounts()` extrinsic we assume that the underlying hardware
has changed, so a quote from a new TSS account ID is expected.


crates/client/src/tests.rs
```diff
@@ -57,13 +89,66 @@ async fn test_change_threshold_accounts() {
...
+
+        let input_data = entropy_shared::QuoteInputData::new(
+            tss_public_key,
+            *x25519_public_key.as_bytes(),
+            nonce,
+            block_number,
+        );
+
+        let mut pck_seeder = StdRng::from_seed(tss_public_key.0);
+        let pck = tdx_quote::SigningKey::random(&mut pck_seeder);
```
HCastano, line 140: @ameba23 sorry since this test is a bit of a mess right now, but do you have any idea why the PCK verification might be failing? Should I be using a different input than then `tss_public_key` for the seeder?

I've commented out the line in the Attestation crate where this fails.

ameba23: Sorry i somehow missed this.

Yes the idea was to use tss account id as the seed.

 I guess with `change_endpoint` its pretty likely that the hardware is changing so PCK would also change (unless the hardware is moving to a different internet connection).  Not sure about `change_threshold_account`.

HCastano: With `change_endpoint` we don't necessarily need to be changing the hardware. Maybe we're just putting our existing node behind a FQDN, or behind a load balancer.

With `change_threshold_account` if we make the change to generate the TSS account ID in the enclave then this would very likely indicate that the hardware has also changed. 

That said, what if in the second case we just spin up a second instance of `entropy-tss` on the same hardware. This would result in a different `account_id` and `x25519_public_key`, but would it have a different PCK? I'm not familiar enough to know.

JesseAbram (approved):

crates/client/src/tests.rs
ameba23: > That said, what if in the second case we just spin up a second instance of entropy-tss on the same hardware. This would result in a different account_id and x25519_public_key, but would it have a different PCK? I'm not familiar enough to know.

In my understanding PCK would be the same but the 'attestation key' should change.  A quote contains two signatures: the quote data is signed with the attestation key, and the attestation public key is signed with the PCK. I chose not to store the attestation public key on chain, because we can already see from the quote itself that it has been endorsed by the PCK - so the PCK is the thing we are interested in.

But this does mean that there can be several TSS nodes associated with the same PCK.  I can't think of a reason why this would be a problem but worth being aware of.

crates/test-cli/src/lib.rs
```diff
@@ -145,6 +145,11 @@ enum CliCommand {
     ChangeEndpoint {
         /// New endpoint to change to (ex. "127.0.0.1:3001")
         new_endpoint: String,
+        /// The Intel TDX quote used to prove that this TSS is running on TDX hardware.
+        ///
+        /// The quote format is specified in:
+        /// https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_TDX_DCAP_Quoting_Library_API.pdf
+        quote: String,
```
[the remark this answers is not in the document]
HCastano: Ya, I just kept the generic for now since it's too early to tell what this will look like.

Made https://github.com/entropyxyz/entropy-core/issues/1154 though.

crates/client/src/tests.rs
```diff
@@ -36,7 +42,33 @@ async fn test_change_endpoint() {
     let api = get_api(&substrate_context.node_proc.ws_url).await.unwrap();
     let rpc = get_rpc(&substrate_context.node_proc.ws_url).await.unwrap();
 
-    let result = change_endpoint(&api, &rpc, one.into(), "new_endpoint".to_string()).await.unwrap();
+    // By using this `Alice` account we can skip the `request_attestation` step since this is
```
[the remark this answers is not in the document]
HCastano: Ya can't remove it yet. It would be nice to do so if we could streamline the quote generation for tests/benches a bit more in the future

crates/test-cli/src/lib.rs
```diff
@@ -155,6 +160,13 @@ enum CliCommand {
         new_tss_account: String,
         /// New x25519 public key
         new_x25519_public_key: String,
+        /// The new Provisioning Certification Key (PCK) certificate chain to be used for the TSS.
+        new_pck_certificate_chain: Vec<String>,
```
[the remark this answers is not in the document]
HCastano: Made https://github.com/entropyxyz/entropy-core/issues/1154

Issue #1127, opened 2024-10-22
Proposal: Do not use persistant storage in `entropy-tss`


ameba23:
Currently we use persistent storage (the on-disk key value database) to store the following things in `entropy-tss`:
 - The network keyshare, if we have one.
 - Our secret keys for TSS account / x25519 encryption.
 - Block numbers and other data for various security checks

There are two problems with this:
 1. Although our kvdb offers encryption with a password, we currently don't have a way of encrypting it such that the operator of the host machine cannot access it.
 2. Even if we solved (1), since the secret keys used in authentication persist re-starting the VM running the `entropy-tss` process, 
   it is possible that the operator restarts with a modified version of the binary but keeps the (encrypted) authentication keys intact.

For these reasons i think we should consider removing the on-disk kvdb, and have secret data be lost if the VM process is killed.
That is, every time the `entropy-tss` binary is launched, fresh TSS account / x25519 keys are generated. These should be submitted to the chain together with a request for attestation.

My proposal of how to do this, would be that the `AppState` struct would look like this:

```rust
#[derive(Clone)]
pub struct AppState {
    listener_state: ListenerState,
    pub configuration: Configuration,
    tss_signer: PairSigner<EntropyConfig, sr25519::Pair>,
    x25519_secret_key: x25519_dalek::StaticSecret,
    keyshare: Arc<RwLock<Option<KeyShareWithAuxInfo>>>
    new_user_last_block_number: Arc<RwLock<Option<u32>>>,
    ...other block numbers etc used in various checks (could also be in a separate struct)
}
```

JesseAbram:
as for all this makes sense, for the proposal tho, 

We can just create one hashmap instead of an individual slot for each item, would keep the same structure as our kvdb

Issue #1015, opened 2024-08-19
TSS account mnemonics and TDX attestation


ameba23:
This related to the TDX attestation feature: https://github.com/entropyxyz/entropy-core/issues/982

In protocol sessions, TS servers are authenticated using their TSS account IDs and x25519 keys, both of which are derived from a mnemonic which is passed in by the user when starting the server. This allows the user to back them up as they wish.

In a TEE context, we want to be sure that the server we are authenticating is running in a trusted domain. We can do this by including these public keys in the attestation quote, which is signed by a hardware-protected keypair.

However, since these keypairs were generated outside of the trusted domain and passed in by the user, the attestation doesn't prove that anyone authenticating with these keys must be in the trusted domain.  We really need them to be generated inside the trusted domain and never leave.

So i propose that they should not be passed in but generated on initial startup. This will make it impossible to back them up externally.  But i don't think they should be backed up. Rather, if access to the trusted domain is lost, a new threshold server with new credentials should be started. The staking extension pallet has an extrinsic which allows these account details to be changed.

The alternative would be to do an attestation at the beginning of every protocol session, including some fresh data such as the session id. But this has the disadvantage that it can only be verified by the other participating parties - who could decide to collude. Authenticating with keys which are linked to an on-chain attestation seems more secure to me.

JesseAbram:
I strongly agree

PR #1128, opened 2024-10-22
Generate mnemonic internally in `entropy-tss`

ameba23:
Addresses https://github.com/entropyxyz/entropy-core/issues/1015 by not allowing mnemonics to be passed in (except in development / testing) via the command line or an environment variable. Instead they are always randomly generated internally.

However i think this does not completely resolve the issue - see https://github.com/entropyxyz/entropy-core/issues/1127

### DIFF
--- a/crates/threshold-signature-server/src/helpers/launch.rs
+++ b/crates/threshold-signature-server/src/helpers/launch.rs
@@ -217,37 +217,6 @@ pub struct StartupArgs {
     /// Returns the AccountID and Diffie-Hellman Public Keys associated with this server.
     #[arg(long = "setup-only")]
     pub setup_only: bool,
-
-    /// The BIP-39 mnemonic (i.e seed phrase) to use for deriving the Threshold Signature Server
-    /// SR25519 account ID and the X25519 public key.
-    ///
-    /// The SR25519 account is responsible for signing and submitting extrinsics to the Entropy
-    /// network.
-    ///
-    /// The X25519 public key is used for encrypting/decrypting messages to other threshold
-    /// servers.
-    ///
-    /// Note that this may keep a mnemonic in your shell history. If you would like to avoid this
-    /// use one of the alternative options, or tools like the 1Password CLI.
-    ///
-    /// **Alternatives**: There are two other ways to supply the mnemonic to the TSS: the
-    /// `--mnemonic-file` flag and the `THRESHOLD_SERVER_MNEMONIC` environment variable.
-    ///
-    /// **Warning**: Passing this flag will overwrite any existing mnemonic! If you would like to
-    /// use an existing mnemonic omit this flag when running the process.
-    #[arg(long = "mnemonic")]
-    pub mnemonic: Option<bip39::Mnemonic>,
-
-    /// The path to a file containing the BIP-39 mnemonic (i.e seed phrase) to use for deriving the
-    /// Threshold Signature Server SR25519 account ID and the X25519 public key.
-    ///
-    /// **Alternatives**: There are two other ways to supply the mnemonic to the TSS: the
-    /// `--mnemonic` flag and the `THRESHOLD_SERVER_MNEMONIC` environment variable.
-    ///
-    /// **Warning**: Passing this flag will overwrite any existing mnemonic! If you would like to
-    /// use an existing mnemonic omit this flag when running the process.
-    #[arg(long = "mnemonic-file", conflicts_with = "mnemonic")]
-    pub mnemonic_file: Option<PathBuf>,
 }
 
 pub async fn has_mnemonic(kv: &KvManager) -> bool {
--- a/crates/threshold-signature-server/src/main.rs
+++ b/crates/threshold-signature-server/src/main.rs
@@ -20,8 +20,8 @@ use clap::Parser;
 use entropy_tss::{
     app,
     launch::{
-        development_mnemonic, load_kv_store, setup_latest_block_number, setup_mnemonic, setup_only,
-        Configuration, StartupArgs, ValidatorName,
+        development_mnemonic, has_mnemonic, load_kv_store, setup_latest_block_number,
+        setup_mnemonic, setup_only, Configuration, StartupArgs, ValidatorName,
     },
     AppState,
 };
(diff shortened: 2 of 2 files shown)


ameba23:
Im pretty sure failing CI is unrelated to this PR - its the good ol' reshare test

JesseAbram:
> If I understand correctly, this will not just "have an impact on devops flow," it will break the possibility of deterministically bringing up a specific TSS server with a prior knowledge of a given private key. This is not really feasible from a devops perpsective _at all_ because we will not be able to replace a broken/failed/otherwise unusable TSS instance without replacing the TSS instance _with a new key_. If that key or account needs to be communicated to any other part of the system, we can not make use of this change, period.
> 
> If this understand is correct, then this will simply break automation, full stop. Please either correct my understanding or do not merge this.

yes it will but also using TDX would break this flow anyways, as well we can automate around this, by having the last step of a spin up process run a script to fund the TSS and inform the chain what the TSS is. Giving access to the kvdb or the TSS accounts outside of TDX removes the security of running TDX

ameba23:
So maybe this is not useful without the other part where the TSS informs the chain of its public keys. With things as they currently are this would need to happen before the `validate` extrinsic gets submitted by the node operator, as the node operator needs to know these public keys for that extrinsic.

The obvious place to do this would be the attestation pallet's `request_attestation` extrinsic, as this needs to anyway be called by the TSS on launch and before `validate`, as `validate` needs to include a quote.  The problem is it doesn't give us the x25519 public key, and i think @HCastano would object to us putting it in there as he's quite keen on not mixing up the logic of the attestation pallet and staking extension pallet.  Also, currently the node operator has no way of getting the quote from the TSS node which they will need for the validate extrinsic (which is a separate issue i guess).

So my proposal would be that the TS server has a GET route which returns its account ID and x25519 public keys as JSON.  Not sure if we could just put this in the existing `/healthz` route or we need a new one.

@entropyxyz/system-reliability-engineers - i feel like it is tricky to design TS launch/attestation process accommodating for both TSS nodes which are hard-coded at genesis as well as dynamically joining ones.  Ideally i would like that we remove TS server details from the testnet chainspec and have them all join after launching the chain.

ameba23:
I have added a GET route for retrieving the account ID and x25519 public key.

ameba23:
@vitropy i am re-requesting a review as this is currently blocked by requested changes. I know this is going to break deployment scripts, but if the plan is to have the next release run on TDX hardware, then the deployment scripts are anyway going to need some changes.

HCastano:
This has been open for a while, we should really try and resolve it soon.

At the moment, having deterministic accounts makes the following easier:
1. Pre-funding TSS servers as appropriate (currently being done via the chainspec
   `balances` but could be done "manually" ahead of time too)
1. Associating TSS accounts to particular a particular node/stash account (done through
   the chainspec's `thresholdServers` field atm)
1. Possibly associate them at the infrastructure level (e.g what account should go on
   what machine and who shold be allowed to talk to whom), but correct me if I'm wrong
   here

For (1), I'm thinking that we can work around this by exploring free transactions again.
There's quite a lot to think about here in order to do this safely (i.e not opening up
DoS vectors for the chain) and it's not as simple as slapping `Pays::No` to some
extrinsics. I still need to do more reading, but maybe having a `SignedExtension` to
faciliate this would be the way.

There's only three extrinsics I found that are called by the TSS (`confirm_jump_start`,
`confirm_key_reshare`, and `request_attestation`) which would make the scope of this
pretty small.

For (2), as Peg suggested we can move away from doing this in the chainspec and instead
having the node operator do this after genesis. From an automation point of view this is
doable but a bit annoying.

The flow would be:
- spin up chain nodes
- spin up TSS nodes
- query TSS info via `GET /info`
- have TSS request an attestation on chain (`pallet_attestation::request_attestation)`
- have operator register their TSS on chain (`pallet_staking_extention::validate`)

Before commenting on (3) I'd need to know a bit more about how the infrastructure is
being spun up atm and what assumptions are being used.

cc @entropyxyz/core-developers @entropyxyz/system-reliability-engineers

ameba23:
@vitropy i can't merge this until you approve :pray:

ameba23:
Related to this PR: https://github.com/entropyxyz/entropy-core/issues/1203

ameba23:
> If I understand correctly, this route is simply providing the same information that the TSS server spits out in its `.entropy/account_id` and `.entropy/public_key` files, no? I would rather continue to use the filesystem as a communication mechanism for this information, not an HTTP request, if it's all the same to you. 

When `entropy-tss` is running in a confidential virtual machine, you wont have ssh access to it.  So the only way to read those files from the host would be to read them out of the virtual machine image file.  Currently you can do this, but ideally it will be encrypted and there will be no way for the host operator to access it (although we haven't figured out how to do that bit yet).
 
> > have operator register their TSS on chain (`pallet_staking_extention::validate`)

In production, the network is going to start small and grow.  When a new validator joins, they need to submit an extrinsic (a transaction to the chain), which contains the details (account ID, http endpoint, etc) of their TSS server.  `pallet_staking_extention::validate` is that extrinsic.  Currently with our testnet setup we don't need to use it because all nodes on the network are pre-configured at genesis.

CHANGELOG.md
```diff
@@ -15,6 +15,18 @@ At the moment this project **does not** adhere to
...
   `AttestationHandler` config type was added to the Staking Extension pallet. The `KeyProvider` and
   `AttestationQueue` config types were removed from the Attestation pallet.
+- In [#1128](https://github.com/entropyxyz/entropy-core/pull/1128) mnemonics can no longer be passed
+  in via a command line argument, file, or environment variable. Instead they are randomly generated
+  internally.
+- In [#1179](https://github.com/entropyxyz/entropy-core/pull/1179) the format of TDX Quote input data has
+  been changed.
+- In [#1147](https://github.com/entropyxyz/entropy-core/pull/1147) a field is added to the
+  chainspec: `jump_started_signers` which allows the chain to be started in a pre-jumpstarted state
+  for testing. If this is not desired it should be set to `None`.
```
HCastano, line 25: This stuff needs to be cleaned up after merging `master` since this PR was open before the last release

CHANGELOG.md
```diff
@@ -15,6 +15,18 @@ At the moment this project **does not** adhere to
...
+
+### Added
+- [#1128](https://github.com/entropyxyz/entropy-core/pull/1128) adds an `/info` route to `entropy-tss`
```
HCastano, line 28:
```suggestion
- In [#1128](https://github.com/entropyxyz/entropy-core/pull/1128) an `/info` route was added to `entropy-tss`
```

CHANGELOG.md
```diff
@@ -15,6 +15,18 @@ At the moment this project **does not** adhere to
   structure, and the `NodeInfoChanged` event were removed from the Staking Extension pallet. The
   `AttestationHandler` config type was added to the Staking Extension pallet. The `KeyProvider` and
   `AttestationQueue` config types were removed from the Attestation pallet.
+- In [#1128](https://github.com/entropyxyz/entropy-core/pull/1128) mnemonics can no longer be passed
```
HCastano, line 18: I'd mention the `--mnemonic` flag(s) here specifically since that would be the "user facing" breaking change

HCastano (approved):
Danke

PR #1134, opened 2024-10-23
Remove declare synced

JesseAbram:
This is a holdover of subgroups, validators are immediately synced as there is only one parent key

### DIFF
--- a/pallets/registry/src/benchmarking.rs
+++ b/pallets/registry/src/benchmarking.rs
@@ -25,8 +25,8 @@ use frame_system::{EventRecord, RawOrigin};
 use pallet_programs::{ProgramInfo, Programs};
 use pallet_session::Validators;
 use pallet_staking_extension::{
-    benchmarking::create_validators, IsValidatorSynced, JumpStartDetails, JumpStartProgress,
-    JumpStartStatus, ServerInfo, ThresholdServers, ThresholdToStash,
+    benchmarking::create_validators, JumpStartDetails, JumpStartProgress, JumpStartStatus,
+    ServerInfo, ThresholdServers, ThresholdToStash,
 };
 use sp_runtime::traits::Hash;
 use sp_std::{vec, vec::Vec};
@@ -48,7 +48,6 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
 
 pub fn add_non_syncing_validators<T: Config>(
     validator_amount: u32,
-    syncing_validators: u32,
 ) -> Vec<<T as pallet_session::Config>::ValidatorId> {
     let validators = create_validators::<T>(validator_amount, SEED);
     let account = account::<T::AccountId>("ts_account", 1, SEED);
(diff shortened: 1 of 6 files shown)


pallets/staking/src/benchmarking.rs
```diff
@@ -272,7 +272,7 @@ benchmarks! {
     let block_number = 1;
     let nonce = NULL_ARR;
     let x25519_public_key = NULL_ARR;
-    let endpoint = b"http://localhost:3001".to_vec();
+    let endpoint = vec![];
```
ameba23, line 275: Why did this change?

ameba23 (approved):
:100:

pallets/staking/src/lib.rs
```diff
@@ -559,18 +546,6 @@ pub mod pallet {
...
-        #[pallet::weight(<T as Config>::WeightInfo::declare_synced())]
-        pub fn declare_synced(origin: OriginFor<T>, synced: bool) -> DispatchResult {
-            let who = ensure_signed(origin.clone())?;
-            let stash = Self::threshold_to_stash(who).ok_or(Error::<T>::NoThresholdKey)?;
-            IsValidatorSynced::<T>::insert(&stash, synced);
-            Self::deposit_event(Event::ValidatorSyncStatus(stash, synced));
-            Ok(())
-        }
-
         #[pallet::call_index(7)]
```
HCastano, line 549: We might as well bite the bullet here and make it a `transaction_version` bump by changing this to `6`

pallets/registry/src/benchmarking.rs
```diff
@@ -48,7 +48,6 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
 
 pub fn add_non_syncing_validators<T: Config>(
```
HCastano, line 49: Can we rename this then to just `add_validators`?

HCastano (approved):
There's also code in the TSS that needs to be moved for this (e.g `--no-sync`). I'm okay if you want to do that here assuming it's not too much code, or in a follow up PR.

This also needs a `CHANGELOG` entry.

pallets/staking/src/benchmarking.rs
JesseAbram: so there are two ways to run benches

compile then run them (like in our scripts) this is the more closer to production way 
or the easier way run them with cargo test --features=runtime-benchmarks

the first uses the runtime config sets 

the second uses the mocks 

So this would pass the first not the second, now it passes both

HCastano: @JesseAbram I thought I bumped the max length to account for this, no? If not, can you bump it to like 25 bytes, it helps makes the test more readable.

JesseAbram: mmm pretty sure that would effect other tests

JesseAbram: also wtv, changed

HCastano: Ah sorry, I did change this buy in #1123 which is still a draft PR 😅

Issue #641, opened 2024-02-16
Speed up protocol execution by executing some of the sub-tasks concurrently


ameba23:
On a hunt for TODOs without related issues ahead of the audit, i noticed there is a bunch of them related to running parts of the protocol execution loop concurrently which @fjarri added.

https://github.com/entropyxyz/entropy-core/blob/ea9d0c5b92a03afff7faf0fadafb4f5853b9fa60/crates/protocol/src/execute_protocol.rs#L83

Some of them relate to sending out protocol messages concurrently.  In practice im not sure how much difference this will make due to the channels we have between the protocol execution task and the tasks which handle websocket connections with the other parties.  Like maybe they are already being sent out concurrently.  

But in two places, we call [`synedrion::sessions::Session::process_message()`](https://docs.rs/synedrion/latest/synedrion/sessions/struct.Session.html#method.process_message) for each peer in serial. 

I think its worth experimenting with getting this to happen concurrently against some benchmarks.

If people think this is something worth prioritizing, i could image to have a play around with this.

See related signing protocol benchmark discussion:  https://github.com/entropyxyz/entropy-core/issues/622

ameba23:
Closing as completed - will be handled by `manul`

PR #1136, opened 2024-10-28
Concurrently process messages in protocol execution loop

ameba23:
Related issue: https://github.com/entropyxyz/entropy-core/issues/641

Now that some fixes in synedrion mean that `Session` is `sync`, and we only need read access to `Session` to process / create messages, message processing / creation from different peers during a round should be able to happen concurrently without needing a mutex. Things are a bit more complicated than last time i tried to do this because we have to handle messages from future sub-protocol for the DKG flow (which is now composed of 3 sub-protocols).

This PR adds concurrency both for processing incoming messages (calls to `session.process_message`) and outgoing messages (calls to `session.make_message`).

Since most of the time goes into processing incoming messages (according to my benchmarks from a while back), i was expecting this to speed things up significantly with a large number of parties. Unfortunately, it does not make too much difference (on my laptop, the signing protocol with 8 parties went from 19.7s to 17.9s).  I think the problem is that i am simulating all parties on one machine, so i don't really benefit from the concurrency - cpu cores are spread across all parties so the total time taken doesn't change much. I hope that in production this wont be the case, because the other parties will be running on other machines.

Trying this on our TDX box, which has 32 cores, the signing protocol with 16 parties takes 12.94s on this branch and 15.25s on master.  I was hoping for a bigger difference.

### DIFF
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -73,7 +73,11 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult + 'static>(
     mut chans: Channels,
     session: Session<Res, sr25519::Signature, PairWrapper, PartyId>,
     session_id_hash: [u8; 32],
-) -> Result<(Res::Success, Channels), GenericProtocolError<Res>> {
+) -> Result<(Res::Success, Channels), GenericProtocolError<Res>>
+where
+    <Res as synedrion::ProtocolResult>::ProvableError: std::marker::Send,
+    <Res as synedrion::ProtocolResult>::CorrectnessProof: std::marker::Send,
+{
     let session_id = synedrion::SessionId::from_seed(&session_id_hash);
     let tx = &chans.0;
     let rx = &mut chans.1;
@@ -106,7 +110,7 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult + 'static>(
         }
 
         // Channel for receiving results of processing messages
-        let (process_tx, mut process_rx) = mpsc::unbounded_channel();
+        let (process_tx, mut process_rx) = mpsc::channel(1024);
         let current_round = session.current_round();
         let session_arc = Arc::new(session);
 
@@ -126,19 +130,16 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult + 'static>(
 
                     if let ProtocolMessagePayload::MessageBundle(payload) = message.payload.clone() {
                         if payload.session_id() == &session_id {
-                            let preprocessed = {
-                                // let session = session_arc.read().unwrap();
-                                // Perform quick checks before proceeding with the verification.
-                                session_arc.preprocess_message(&mut accum, &message.from, *payload)?
-                            };
+                            // Perform quick checks before proceeding with the verification.
+                            let preprocessed =
+                                session_arc.preprocess_message(&mut accum, &message.from, *payload)?;
 
                             if let Some(preprocessed) = preprocessed {
                                 let session_clone = session_arc.clone();
                                 let tx_clone = process_tx.clone();
                                 tokio::spawn(async move {
-                                    // let session = session_clone.read().unwrap();
-                                    let result = session_clone.process_message(&mut OsRng, preprocessed).unwrap();
-                                    tx_clone.send(result).unwrap();
+                                    let result = session_clone.process_message(&mut OsRng, preprocessed);
+                                    tx_clone.send(result).await.unwrap();
                                 });
                             }
                         } else {
@@ -154,7 +155,7 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult + 'static>(
                 // Result from processing a message
                 maybe_result = process_rx.recv() => {
                     if let Some(result) = maybe_result {
-                        accum.add_processed_message(result)??;
+                        accum.add_processed_message(result?)??;
                     }
                 }
             }
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -95,58 +95,41 @@ where
 
         // Send outgoing messages
         let destinations = session_arc.message_destinations();
-        if !destinations.is_empty() {
-            let mut handles = Vec::new();
-            for destination in destinations.iter() {
-                let session_arc = session_arc.clone();
-                let tx = tx.clone();
-                let my_id = my_id.clone();
-                let destination = destination.clone();
-                let join_handle = tokio::spawn(async move {
-                    let result = match session_arc.make_message(&mut OsRng, &destination) {
-                        Ok((message, artifact)) => {
-                            match tx.send(ProtocolMessage::new(&my_id, &destination, message)) {
-                                Ok(_) => Ok(artifact),
-                                Err(err) => {
-                                    let err: GenericProtocolError<Res> = err.into();
-                                    Err(err)
-                                },
-                            }
-                        },
-                        Err(err) => Err(err.into()),
-                    };
-                    result
-                });
-                handles.push(join_handle)
-            }
+        let join_handles = destinations.iter().map(|destination| {
+            let session_arc = session_arc.clone();
+            let tx = tx.clone();
+            let my_id = my_id.clone();
+            let destination = destination.clone();
+            tokio::spawn(async move {
+                session_arc
+                    .make_message(&mut OsRng, &destination)
+                    .map(|(message, artifact)| {
+                        tx.send(ProtocolMessage::new(&my_id, &destination, message))
+                            .map(|_| artifact)
+                            .map_err(|err| {
+                                let err: GenericProtocolError<Res> = err.into();
+                                err
+                            })
+                    })
+                    .map_err(|err| {
+                        let err: GenericProtocolError<Res> = err.into();
+                        err
+                    })
+            })
+        });
 
-            for result in try_join_all(handles).await? {
-                accum.add_artifact(result?)?;
-            }
+        for result in try_join_all(join_handles).await? {
+            accum.add_artifact(result??)?;
         }
 
         // Process cached messages
-        if !cached_messages.is_empty() {
-            let cached_messages_len = cached_messages.len();
-            let (process_tx, mut process_rx) = mpsc::channel(cached_messages_len);
-            for preprocessed in cached_messages {
-                let session_arc = session_arc.clone();
-                let process_tx = process_tx.clone();
-                tokio::spawn(async move {
-                    let result = session_arc.process_message(&mut OsRng, preprocessed);
-                    if process_tx.send(result).await.is_err() {
-                        tracing::error!(
-                            "Protocol finished before cached message processing result sent"
-                        );
-                    }
-                });
-            }
+        let join_handles = cached_messages.into_iter().map(|preprocessed| {
+            let session_arc = session_arc.clone();
+            tokio::spawn(async move { session_arc.process_message(&mut OsRng, preprocessed) })
+        });
 
-            for _ in 0..cached_messages_len {
-                if let Some(result) = process_rx.recv().await {
-                    accum.add_processed_message(result?)??;
-                }
-            }
+        for result in try_join_all(join_handles).await? {
+            accum.add_processed_message(result?)??;
         }
 
         // Receive and process incoming messages
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -18,7 +18,7 @@
 use num::bigint::BigUint;
 use rand_core::{CryptoRngCore, OsRng};
 use sp_core::{sr25519, Pair};
-use std::collections::VecDeque;
+use std::sync::{Arc, Mutex};
 use subxt::utils::AccountId32;
 use synedrion::{
     ecdsa::VerifyingKey,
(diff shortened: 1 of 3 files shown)


crates/protocol/src/execute_protocol.rs
```diff
@@ -85,64 +89,118 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
...
+            // Channel for receiving message artifacts
+            let (artifact_tx, mut artifact_rx) = mpsc::channel(destinations.len());
+            for destination in destinations.iter() {
+                let session_arc = session_arc.clone();
+                let tx = tx.clone();
+                let my_id = my_id.clone();
+                let artifact_tx = artifact_tx.clone();
+                let destination = destination.clone();
+                tokio::spawn(async move {
+                    let result = match session_arc.make_message(&mut OsRng, &destination) {
```
ameba23, line 107: These match expressions are a bit of a mess. If anyone knows a neater way to map these let me know.

crates/protocol/src/execute_protocol.rs
```diff
@@ -85,64 +89,118 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
...
+                        accum.add_processed_message(result?)??;
+                    }
+                }
             }
         }
 
-        match session.finalize_round(&mut OsRng, accum)? {
+        // Get session back out of Arc
+        let session_inner =
+            Arc::try_unwrap(session_arc).map_err(|_| GenericProtocolError::ArcUnwrapError)?;
```
ameba23, line 202: Here we assume there are no other references to the session at this point, which is always the case in the happy path as the protocol only finished when all messages are processed.  

@fjarri Im not sure if there are some syendrion error cases where `session.can_finalize()` will return true before all received messages have been processed.  If that is the case, instead of returning an error here we should wait until all other references to `session` are dropped.

The alternative would be to put it in a `Arc<RwLock<>>` for the entire protocol (not just one round).

crates/protocol/tests/protocol.rs
```diff
@@ -33,37 +33,40 @@ use helpers::{server, ProtocolOutput};
 
 use std::collections::BTreeSet;
 
+/// The maximum number of worker threads that tokio should use
+const MAX_THREADS: usize = 16;
```
ameba23, line 37: I was setting the number of tokio worker threads to `num_cpus`, but i had issue running with 32 worker threads on our 32 core box. Tokio documentation says 'it is advised to keep this value on the smaller side'.

ameba23:
I had to re-run CI twice to get this to pass - the failures don't look to me like they are related to this PR.  I will make an issue.

crates/protocol/src/execute_protocol.rs
```diff
@@ -85,64 +89,118 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
...
+                                    let err: GenericProtocolError<Res> = err.into();
+                                    Err(err)
+                                },
+                            }
+                        },
+                        Err(err) => Err(err.into()),
+                    };
+                    if artifact_tx.send(result).await.is_err() {
+                        tracing::error!("Protocol finished before outgoing message artifact sent");
+                    }
```
HCastano, line 121:
```suggestion
                tokio::spawn(async move {
                    let result = session_arc
                        .make_message(&mut OsRng, &destination)
                        .map(|(message, artifact)| {
                            tx.send(ProtocolMessage::new(&my_id, &destination, message))
                                .map(|_| artifact)
                                .map_err(|err| {
                                    let err: GenericProtocolError<Res> = err.into();
                                    err
                                })
                        })
                        .map_err(|err| {
                            let err: GenericProtocolError<Res> = err.into();
                            err
                        });

                    if artifact_tx.send(result).await.is_err() {
                        tracing::error!("Protocol finished before outgoing message artifact sent");
                    }
                });
```

Got this to work, but not sure how much better it is tbh. Also had to made a small change below

HCastano (approved):
Kinda sucks that the performance numbers aren't better, but maybe the tests on a single machine aren't representative.

I'd wait until Bogdan reviews to merge since I'm not the best with async Rust.

crates/protocol/src/execute_protocol.rs
```diff
@@ -69,11 +69,15 @@ impl RandomizedPrehashSigner<sr25519::Signature> for PairWrapper {
     }
 }
 
-pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
+pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult + 'static>(
```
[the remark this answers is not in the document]
ameba23: This is from the error handling using the generic `GenericProtocolError<Res>` in a spawned task.  Probably its possible to use some other error type in the spawned task and converted to a generic error after its been passed back out to the host task.  Hopefully then we wouldn't need the static lifetime.

crates/protocol/src/execute_protocol.rs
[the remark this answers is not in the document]
ameba23: Agree we should put a better error message.  But ideally i would only leave this in if i was pretty sure this was never going to happen.  We only get to this point in the code when the loop has broken because `can_finalize` is true - which means the protocol round is finished.  If there are still some references to `Session` lying around in spawned tasks at that point then this will fail. But that would also mean some messages have not yet been processed, so the round should not be finished.

Since `try_unwrap` actually returns the `Arc` back in the error variant of the `Result` if it fails, it would be possible to loop around waiting until those references get dropped because the tasks finish.  But i think its only worth doing that if it is actually possible to get into such a state.

crates/protocol/src/execute_protocol.rs
ameba23: thank you!

fjarri:
> I think the problem is that i am simulating all parties on one machine, so i don't really benefit from the concurrency - cpu cores are spread across all parties so the total time taken doesn't change much.

I can't come up with an easy way to do this. `manul` has a testing helper that takes a list of sessions and simulates execution in a single thread; we could extend that to optionally add parallel message processing, so that you could run the protocol with and without it in a benchmark and see a more informative difference. But that has to wait until https://github.com/entropyxyz/synedrion/pull/156 is in.

crates/protocol/src/execute_protocol.rs
fjarri: Just spitballing here, maybe `Result::and_then()` will be cleaner?

crates/protocol/src/execute_protocol.rs
fjarri: Not in the current `synedrion` The current `manul` version can exit prematurely if there are so many invalid messages already registered that it will be impossible to finalize (and for the current protocols it means if *at least one* invalid message is registered). Also it will be possible to finalize early in the future with the support of nodes dropping out during the protocol (https://github.com/entropyxyz/manul/issues/11). 

The way I envision this API to be used is that as soon as you get the definitive answer from `can_finalize()` (`true` now, or `CanFinalize::Yes` or `Never` in `manul`), you stop processing messages, wait for all the existing processing tasks to finish (there are no cancel points in them), and finalize the round (or terminate the session if the round can never be finalized, in `manul`). 

Now for the waiting part, I know how I would organize it in Python, but I am not exactly sure for `tokio` - haven't played with async much in Rust. I just assumed that if Python can do it, it would be possible somehow in Rust too, I hope I was correct :)

ameba23:
> I can't come up with an easy way to do this. `manul` has a testing helper that takes a list of sessions and simulates execution in a single thread; we could extend that to optionally add parallel message processing, so that you could run the protocol with and without it in a benchmark and see a more informative difference. But that has to wait until [entropyxyz/synedrion#156](https://github.com/entropyxyz/synedrion/pull/156) is in.

Good, that sounds promising.  I was wondering if there may be able to do it by using a deterministic RNG and running the protocol twice.  On the second time all but one parties replay their responses from the first time, so only one party is actually doing any computation, and we could benchmark it.

crates/protocol/src/execute_protocol.rs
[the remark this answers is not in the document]
ameba23: I've tried to remove the need for this static and can't come up with something.

Having read up about it, i am with @JesseAbram on this - think we should not worry.

The answer here:
https://stackoverflow.com/questions/66510485/when-is-a-static-lifetime-not-appropriate

says that `'static` means **can** live forever, not **will** live forever.  We can see from this code that the references in the spawned tasks are only going to be around as long as the messages are still being processed.

crates/protocol/src/execute_protocol.rs
```diff
@@ -69,11 +70,15 @@ impl RandomizedPrehashSigner<sr25519::Signature> for PairWrapper {
...
     mut chans: Channels,
     session: Session<Res, sr25519::Signature, PairWrapper, PartyId>,
     session_id_hash: [u8; 32],
-) -> Result<(Res::Success, Channels), GenericProtocolError<Res>> {
+) -> Result<(Res::Success, Channels), GenericProtocolError<Res>>
+where
+    <Res as synedrion::ProtocolResult>::ProvableError: std::marker::Send,
+    <Res as synedrion::ProtocolResult>::CorrectnessProof: std::marker::Send,
```
dvdplm, line 80: Do you really need the full path here? Isn't `Send` enough?

crates/protocol/src/execute_protocol.rs
```diff
@@ -85,64 +90,96 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
...
-            let processed = session.process_message(&mut OsRng, preprocessed)?;
+        // Process cached messages
+        let join_handles = cached_messages.into_iter().map(|preprocessed| {
+            let session_arc = session_arc.clone();
+            tokio::spawn(async move { session_arc.process_message(&mut OsRng, preprocessed) })
+        });
 
-            // This will happen in a host task.
-            accum.add_processed_message(processed)??;
+        for result in try_join_all(join_handles).await? {
```
dvdplm, line 131: This will abort all the `process_message` tasks and return immediately. Is this what we want here? It might be exactly what we want, but just checking it's what we expect.

crates/protocol/src/execute_protocol.rs
```diff
@@ -85,64 +90,96 @@ pub async fn execute_protocol_generic<Res: synedrion::ProtocolResult>(
...
+                            let preprocessed =
+                                session_arc.preprocess_message(&mut accum, &message.from, *payload)?;
+
+                            if let Some(preprocessed) = preprocessed {
+                                let session_arc = session_arc.clone();
+                                let tx = process_tx.clone();
+                                tokio::spawn(async move {
+                                    let result = session_arc.process_message(&mut OsRng, preprocessed);
+                                    if tx.send(result).await.is_err() {
+                                        tracing::error!("Protocol finished before message processing result sent");
```
dvdplm, line 157: Hmm, isn't this a `SendError`, i.e. the rx side of the channel closed or the buffer is full or some such shenanigans? Like, the log message seems to assume that cause is that the protocol finished prematurely but are we positive that is the case?

crates/protocol/src/execute_protocol.rs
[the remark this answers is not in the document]
dvdplm: +1 for an error type describing the problem, `ArcUnwrapError`.

crates/protocol/tests/protocol.rs
dvdplm: What kind of problems did you run into? And perhaps why not pick a multiple of whatever `num_cpu` reports? Like half or a quarter?

dvdplm (commented):
I am curious about the benchmarks numbers you're seeing and if you did any profiling (by eyeballing or using some tool) on this? Where is the time spent?

crates/protocol/tests/protocol.rs
ameba23: The problem was the test just hangs and only finishes when the test harness timeout runs out. I guess we could pick half of `num_cpus` if num_cpus is greater than some limit.  Since in CI i think we only have 2 cpus, so we definitely want both.  Bear in mind this is only for tests / benchmarks.  I think in production we use the tokio default which if ive understood right is actually the number of cpus: https://docs.rs/tokio/latest/tokio/attr.main.html#multi-threaded-runtime

crates/protocol/src/execute_protocol.rs
ameba23: You mean if there is an error in one of the tasks, all will be aborted because of the `?` on `try_join_all`?  If so, yes i think this is what we want.  If we fail to process any message we assume we cannot recover and abort the protocol session.

crates/protocol/src/execute_protocol.rs
ameba23: I am pretty sure that it is the case.

If the channel is full there will be no error, it will `await` until there is space for a message.

That only leaves that the receiver is closed due to either `close()` being called (which we never do) or the the receiver being dropped.

crates/protocol/src/execute_protocol.rs
ameba23: Yes you are right we can just do `Send`

ameba23:
> I am curious about the benchmarks numbers you're seeing and if you did any profiling (by eyeballing or using some tool) on this? Where is the time spent?

@dvdplm It was back in April i did this so things might have changed, and i didn't have any special technique other than adding a lot of log lines with `std::time::Instant::elapsed()`. I was using `criterion` for benchmarking, but abandoned it because of strange problems with things appearing to work on my laptop but not on @HCastano 's. See 
https://github.com/entropyxyz/entropy-core/pull/738#discussion_r1569098163

I'm pretty sure i came to the conclusion that incoming message processing is the biggest time hog for the signing protocol.  I didn't pay much attention to the other protocols because the signing protocol is the only one where we keep the user waiting.

It would be great to look at this in more detail. Probably it makes more sense to look at synedrion's benchmarking setup, but i wanted to also do it here so that we can try to optimise the execution loop like in this PR, and also to take into account the overhead of websockets and encryption.

Thanks for taking the time to review this btw.

crates/protocol/src/execute_protocol.rs
dvdplm: Less because of the `?` and more because of how `try_join_all` works: abort as soon as any of the tasks abort.

PR #1140, opened 2024-11-01
Protocol message versioning

ameba23:
This adds a version number to the `SubscribeMessage` sent at the beginning of each protocol session, and checks that it matches the current version (or a list of supported backward compatible versions which currently only contains the current version).

We could also add a version number to `ProtocolMessage`, but i decided not to - i think check the version once at the beginning of each protocol session is enough - i don't think we need to check individual messages during a session as well.

### DIFF
--- a/crates/protocol/src/protocol_transport/subscribe_message.rs
+++ b/crates/protocol/src/protocol_transport/subscribe_message.rs
@@ -13,10 +13,12 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use crate::SessionId;
+use crate::{SessionId, PROTOCOL_MESSAGE_VERSION, SUPPORTED_PROTOCOL_MESSAGE_VERSIONS};
 use serde::{Deserialize, Serialize};
 use sp_core::{sr25519, Pair};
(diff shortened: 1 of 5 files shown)


crates/protocol/src/protocol_transport/subscribe_message.rs
```diff
@@ -47,4 +55,49 @@ impl SubscribeMessage {
...
+
+    pub fn check_supported(&self) -> Result<(), ProtocolVersionMismatchError> {
+        if self.version > PROTOCOL_MESSAGE_VERSION {
+            Err(ProtocolVersionMismatchError::VersionTooNew(PROTOCOL_MESSAGE_VERSION))
+        } else if !SUPPORTED_PROTOCOL_MESSAGE_VERSIONS.contains(&self.version) {
+            Err(ProtocolVersionMismatchError::VersionTooOld(
+                *SUPPORTED_PROTOCOL_MESSAGE_VERSIONS
+                    .iter()
+                    .min()
+                    .expect("At least one protocol message version must be supported"),
```
ameba23, line 67: I think `expect` is ok here because it will only fail if `SUPPORTED_PROTOCOL_MESSAGE_VERSIONS`, which is a `const`, is empty.

JesseAbram (approved):

Issue #1119, opened 2024-10-18
Jumpstart failing in CI runs


HCastano:
Looks like the jumpstart is failing periodically for some of our tests. For example, see [this CI run](https://github.com/entropyxyz/entropy-core/actions/runs/11408861423/job/31747928369?pr=1109).

This is probably because these tests rely on timeouts. It would be nice to get this behaviour to be more consistent.

frankiebee:
jump start is failing consistently in sdk [ci](https://github.com/entropyxyz/entropy-core/issues/1112) because of timeout and we have a 6 minute timeout

frankiebee:
we seem to have got it passing now all of the sudden no explanation as to why when we added are exit call (are tests wernt exiting when jump start fails) but thats unrelated 🤷

mixmix:
i want to know if it's intermittent. Jump start failing unpredictably will massively hamper our ability to work (+ approve rc's)

HCastano:
@frankiebee @mixmix while I'm not ruling out that this could be a problem with the jump start itself, for this specific issue it's probably more related to the Rust test success/failure criteria.

> jump start is failing consistently in sdk [ci](https://github.com/entropyxyz/sdk/actions/runs/11355042770/job/31750437090) because of timeout and we have a 6 minute timeout

I would try and shorten the retry time there. You shouldn't need to wait more than maybe a minute for the jumpstart to happen.

JesseAbram:
hmmm interesting, will investigate more next week

frankiebee:
we are retrying every 10 blocks watching logs it seems when jump start succeeds locally it does this within a 3 block period

frankiebee:
unless theirs a reason we should try for 50? i'm seeing that when it gets retried at 50 blocks it works

JesseAbram:
50 blocks is the default time we allow before a retry can happen

frankiebee:
Got it. 
note: sdk time out is now based on blocks and i'm [retrying every 50 up to 200 ](https://github.com/entropyxyz/sdk/pull/421/files#diff-f3b6dc936840aaa286b87afbfa57e827e8c68f402c5e74f21e755ef20a8898e8R94-R104)from the first try
note: what it looks like from sdk tests 3 out of 8 trys to jumpstart suceed the first go arouznd the other 5 have to retry.
[...]

PR #1147, opened 2024-11-05
Allow offchain worker requests to all TSS nodes in `entropy-tss` test environment 

ameba23:
This PR sets the TSS node endpoints associated with all chain nodes in our test setup for `entropy-tss`.

It means that http requests coming from the propagation pallet will be made to all four TSS nodes.  Previously, only Alice got a request from the chain and the other 3 TSS server's had similar requests made by an http client in the test code itself.

This had the advantage that we could test the handling of bad inputs given to these endpoints.  But a lot of care had to be taken to make the mock requests appear at the right time, and we have had problems with the tests occasionally failing, eg: https://github.com/entropyxyz/entropy-core/issues/1119

I am hoping that we can make the tests more reliable and more closely emulate what will happen in production by doing this.

This PR also makes a change to the genesis config making it possible to start in a pre-jumpstarted state to test signing.  For an explainantion see the (merged) sub-PR which introduces that change: https://github.com/entropyxyz/entropy-core/pull/1162

A couple of related changes:
- There is now only one set of pre-generated keyshares, as we know beforehand who the initial signer set will be as it is part of the chainspec.
- I've used a different helper function when setting up a lot of our tests: `testing_utils::helpers::spawn_tss_nodes_and_start_chain`. This has the advantage that it sets up both the chain and TSS nodes in one function, making a bit less boilerplate code in the tests.  But it has a gotcha, which is that because it lives in `testing-utils` and not `entropy-tss`, `cfg(test)` is false meaning the unsafe api does not get loaded. So it can only be used in the tests which don't need the unsafe API.  I could move it into `entropy-tss`, but then we would not be able to use it in integration tests.
- We had a test which did a DKG, then signs a message, then a reshare, then signs another message. The problem with this is that if doing a real DKG, the signer set is not known at genesis.  Before, we had chosen the `next_signers` for the reshare to match what signers get chosen, but there is no guarantee that this will continue to work if the initial chain state changes.  So now we only test resharing following a mock DKG where we know the initial signers beforehand.


ameba23:
I am a bit stuck with this.
 
DKG is working fine but the reshare protocol has problems.

To be specific its the aux gen protocol when run from within the reshare protocol:

The DKG protocol is composed of the key init, reshare and aux gen protocol - so we know all of these will work with this setup.

Reshare is composed of reshare and aux gen.  It seems from looking at logs that the reshare runs, but only two of the three parties finalize successfully, then during aux gen a connection gets dropped.

I tried switching things around so the aux gen protocol gets run before the reshare, and it seems that aux gen is the problem as a connection gets dropped and it seems only one peer runs the protocol loop. 

There are also issues with mocking the DKG using pre-generated keyshares which we do in the signing tests, as now when the jump start extrinsic is submitted, it starts a real DKG, and things get messy when the mock DKG confirmations are also submitted.  But i think this should be not to hard to fix.

ameba23:
I had a hunch that the problems i am seeing are related to problems introduced by my recent PR: https://github.com/entropyxyz/entropy-core/pull/1136

I have now tried rolling back to before that commit and re-applying these changes and running the tests, but i am seeing exactly the same behavior as on this branch.  So thats not the problem.

Here is the relevant part of the test logging output:
```
  2024-11-11T09:45:06.271344Z  INFO entropy_tss::signing_client::protocol_transport: Got ws connection, with message: SubscribeMessage { session_id: Reshare { verifying_key: [2, 24, 213, 93, 160, 174, 33, 208, 3, 29, 227, 227, 175, 53, 23, 229, 3, 220, 55, 68, 234, 118, 51, 129, 176, 175, 185, 68, 247, 93, 250, 222, 233], block_number: 21 }, public_key: 2cbc68e8bf0fbc1c28c282d1263fc9d29267dc12a1044fb730e8b65abc37524c (5D5Mw6Wb...), signature: 8259ea5bbf1308232f884f7eaf85d494c664e318ef21f1ee26b5575c0fa542789c2ca902e78fb74447016c549ff35b4cc67353360041b295da14ab0e00348781 }
    at crates/threshold-signature-server/src/signing_client/protocol_transport.rs:159

  2024-11-11T09:45:06.271898Z  INFO entropy_tss::signing_client::protocol_transport: Got ws connection, with message: SubscribeMessage { session_id: Reshare { verifying_key: [2, 24, 213, 93, 160, 174, 33, 208, 3, 29, 227, 227, 175, 53, 23, 229, 3, 220, 55, 68, 234, 118, 51, 129, 176, 175, 185, 68, 247, 93, 250, 222, 233], block_number: 21 }, public_key: 946140d3d5ddb980c74ffa1bb64353b5523d2d77cdf3dc617fd63de9d3b66338 (5FRFqLyd...), signature: 9aa6c7204d2135cadb94878b4117617a9336c4becd2ceb04903d570a7ad6fa38d223dec1ca25f4279a49191296a8cfeb386117be6b33e912e55fc1e22aaba288 }
    at crates/threshold-signature-server/src/signing_client/protocol_transport.rs:159

  2024-11-11T09:45:06.272088Z  INFO entropy_tss::signing_client::protocol_transport: Got ws connection, with message: SubscribeMessage { session_id: Reshare { verifying_key: [2, 24, 213, 93, 160, 174, 33, 208, 3, 29, 227, 227, 175, 53, 23, 229, 3, 220, 55, 68, 234, 118, 51, 129, 176, 175, 185, 68, 247, 93, 250, 222, 233], block_number: 21 }, public_key: 946140d3d5ddb980c74ffa1bb64353b5523d2d77cdf3dc617fd63de9d3b66338 (5FRFqLyd...), signature: b4a22cbf3746335a7a757824c63cbeb01a31a3fa428d4a860c47ae790257a118bdabe88fe31e4ad8a854d8d49b6770c7fd192a7a88813753dc116a16bd726b82 }
    at crates/threshold-signature-server/src/signing_client/protocol_transport.rs:159

  2024-11-11T09:45:06.272212Z  INFO entropy_protocol::execute_protocol: Executing reshare
    at crates/protocol/src/execute_protocol.rs:343
    in entropy_tss::validator::api::new_reshare
    in entropy_tss::http-request with uuid: d4731f8e-5b02-4f75-8218-6a612164990a, uri: /validator/reshare, method: POST

  2024-11-11T09:45:06.272554Z  INFO entropy_protocol::execute_protocol: Executing reshare
    at crates/protocol/src/execute_protocol.rs:343
    in entropy_tss::validator::api::new_reshare
    in entropy_tss::http-request with uuid: 829e963e-68a0-4c46-b628-a42aa0c1bba4, uri: /validator/reshare, method: POST

  2024-11-11T09:45:06.272964Z  INFO entropy_protocol::execute_protocol: Executing reshare
    at crates/protocol/src/execute_protocol.rs:343
    in entropy_tss::validator::api::new_reshare
    in entropy_tss::http-request with uuid: badb679b-ba0f-4287-bc39-f65952c250f2, uri: /validator/reshare, method: POST

  2024-11-11T09:45:06.274708Z  INFO entropy_protocol::execute_protocol: Finished reshare
    at crates/protocol/src/execute_protocol.rs:361
    in entropy_tss::validator::api::new_reshare
    in entropy_tss::http-request with uuid: 829e963e-68a0-4c46-b628-a42aa0c1bba4, uri: /validator/reshare, method: POST

  2024-11-11T09:45:06.274719Z  INFO entropy_protocol::execute_protocol: Starting aux gen
    at crates/protocol/src/execute_protocol.rs:365
    in entropy_tss::validator::api::new_reshare
    in entropy_tss::http-request with uuid: 829e963e-68a0-4c46-b628-a42aa0c1bba4, uri: /validator/reshare, method: POST

  2024-11-11T09:45:10.401568Z  INFO entropy_protocol::execute_protocol: Finished reshare
    at crates/protocol/src/execute_protocol.rs:361
    in entropy_tss::validator::api::new_reshare
    in entropy_tss::http-request with uuid: d4731f8e-5b02-4f75-8218-6a612164990a, uri: /validator/reshare, method: POST

  2024-11-11T09:45:10.401590Z  INFO entropy_protocol::execute_protocol: Starting aux gen
    at crates/protocol/src/execute_protocol.rs:365
    in entropy_tss::validator::api::new_reshare
    in entropy_tss::http-request with uuid: d4731f8e-5b02-4f75-8218-6a612164990a, uri: /validator/reshare, method: POST

  2024-11-11T09:45:17.100457Z  WARN entropy_tss::signing_client::api: Websocket connection closed unexpectedly MessageAfterProtocolFinish
    at crates/threshold-signature-server/src/signing_client/api.rs:140
```

ameba23:
Had a look at this problem with @HCastano thismorning.  He made a suggestion to try running the DKG with an additional reshare immediately after the DKG finishes.  To see if it is something to do with the test conditions for the reshare test rather than the reshare protocol itself.

I tried doing this, but got into a muddle with new holders and old holders, and was having synedrion's `no entry found for key` error.

But if i run an extra aux gen protocol at the end of the DKG protocol, it does work.

crates/testing-utils/src/helpers.rs
```diff
@@ -26,31 +26,28 @@ pub use tdx_quote::encode_verifying_key;
 
 /// A helper for setting up tests which starts both a set of TS servers and a chain node and returns
 /// the chain API as well as IP addresses and PartyId of the started validators
+///
+/// Note that since this function does not reside in entropy-tss, cfg(test) will be false when the
+/// TSS nodes are set up, meaning the unsafe API will not be enabled
```
dvdplm, line 31: Why not introduce a `testing` feature and change the cfg clause to be `#[cfg(any(feature = "testing", test))]`?

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -41,195 +35,81 @@ use entropy_client::{
...
-    let new_signers = {
-        let signer_query = entropy::storage().staking_extension().signers();
-        let signer_ids = query_chain(&api, &rpc, signer_query, None).await.unwrap().unwrap();
-        let mut signers = Vec::new();
-        for signer in signer_ids {
-            let query = entropy::storage().staking_extension().threshold_servers(signer);
-            let server_info = query_chain(&api, &rpc, query, None).await.unwrap().unwrap();
-            signers.push(server_info);
+    let mut i = 0;
+    let new_signer_ids = loop {
```
dvdplm, line 94:
```suggestion
    // Wait up to 2min for reshare to complete: check once every second if we have a new set of signers.
    let new_signer_ids = loop {
```

maybe can be helpful for readers?

dvdplm: Just out of curiosity, but doesn't `subxt` provide facilities to subscribe to changes to some storage item? Setup a callback that gets called when the event happens?

ameba23: Yep you are right, i think this is the event we need to wait for: https://github.com/entropyxyz/entropy-core/blob/fddd9e4fc41d7526fb361b976a7de96938e97dd1/pallets/staking/src/lib.rs#L374

HCastano (commented):
All looks fairly reasonable, but I need another look before approving

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -192,21 +198,24 @@ pub async fn spawn_testing_validators(
...
+    });
+    let dave_id = PartyId::new(SubxtAccountId32(
+        *get_signer(&dave_kv).await.unwrap().account_id().clone().as_ref(),
+    ));
+    ids.push(dave_id);
+
+    if chain_spec_type == ChainSpecType::IntegrationJumpStarted {
+        put_keyshares_in_db(ValidatorName::Alice, alice_kv).await;
+        put_keyshares_in_db(ValidatorName::Bob, bob_kv).await;
+        put_keyshares_in_db(ValidatorName::Charlie, charlie_kv).await;
```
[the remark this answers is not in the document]
ameba23: Yeah, theres always one party who doesn't have a keyshare

crates/testing-utils/src/helpers.rs
ameba23: That is a good idea

crates/testing-utils/src/helpers.rs
```diff
@@ -26,31 +26,28 @@ pub use tdx_quote::encode_verifying_key;
...
 pub async fn spawn_tss_nodes_and_start_chain(
     chain_spec_type: ChainSpecType,
-) -> (OnlineClient<EntropyConfig>, LegacyRpcMethods<EntropyConfig>, Vec<String>, Vec<PartyId>) {
+) -> (
+    Vec<TestNodeProcess<EntropyConfig>>,
+    OnlineClient<EntropyConfig>,
+    LegacyRpcMethods<EntropyConfig>,
+    Vec<String>,
+    Vec<PartyId>,
```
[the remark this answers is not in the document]
ameba23: Thats a good idea. But could turn into quite a big change, because once having a struct we may as well pass it to test helper functions.

Generally, we could bundle api and rpc together into a struct and use it in all client functions.

The `validator_ips` and `validator_ids` we actually very rarely use because we know what are going to be.  So i think they could be ditched from here and replaced with constants.

I would maybe do that in a followup.

pallets/staking/src/lib.rs
```diff
@@ -308,8 +311,19 @@ pub mod pallet {
                     new_signers,
                 })
             }
+
+            if let Some(jump_started_signers) = &self.jump_started_signers {
+                Signers::<T>::put(jump_started_signers.clone());
+                JumpStartProgress::<T>::put(JumpStartDetails {
+                    jump_start_status: JumpStartStatus::Done,
+                    confirmations: jump_started_signers.clone(),
+                    verifying_key: Some(BoundedVec::try_from(EVE_VERIFYING_KEY.to_vec()).unwrap()),
```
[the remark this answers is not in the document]
ameba23: This is for the case where we start the chain in a pre-jumpstarted state for tests, and use the pre-generated keyshares, for which we know the verifying key.  We should probably rename that constant to something like `PREGENERATED_NETWORK_VERIFYING_KEY`.

crates/threshold-signature-server/src/validator/tests.rs
ameba23: I didn't mean to resolve this conversation - github did it when i pressed 'commit suggestion'

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -430,3 +362,8 @@ async fn put_jumpstart_request_on_chain(
     let registering_tx = entropy::tx().registry().jump_start_network();
     submit_transaction(api, rpc, &account, &registering_tx, None).await.unwrap();
 }
+
+/// Given a ServerInfo, get the port number
+pub fn get_port(server_info: &ServerInfo<SubxtAccountId32>) -> u32 {
+    std::str::from_utf8(&server_info.endpoint).unwrap().split(":").last().unwrap().parse().unwrap()
```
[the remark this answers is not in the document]
ameba23: Agree, good call.  It turns out `SocketAddr.port()` returns a u16, which i guess is the idiomatic way to represent a port number, but our unsafe_get helper for which this is used takes a u32.  I think im just going to stick with u32 for rather than change that helper is this PR.

crates/shared/src/constants.rs
```diff
@@ -12,8 +12,8 @@ pub const DAVE_VERIFYING_KEY: EncodedVerifyingKey = [
     3, 42, 97, 187, 199, 208, 95, 166, 102, 15, 38, 146, 173, 111, 175, 123, 62, 132, 178, 237,
     150, 199, 194, 240, 153, 30, 113, 104, 57, 63, 54, 2, 65,
 ];
-// This key is associated with a constant key share generation from DETERMINISTIC_KEY_SHARE_EVE
-pub const EVE_VERIFYING_KEY: EncodedVerifyingKey = [
+// This key is associated with a constant key share generation from DETERMINISTIC_KEY_SHARE_NETWORK
+pub const PREGENERATED_NETWORK_VERIFYING_KEY: EncodedVerifyingKey = [
```
ameba23, line 16: This is just renamed for clarity

crates/threshold-signature-server/tests/jumpstart_register_sign.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright (C) 2023 Entropy Cryptography Inc.
```
ameba23, line 1: I renamed the test file which makes the diff hard to read.  But basically this test no longer includes a reshare.

ameba23:
Disappointingly, the reshare test failed on this branch thismorning with a timeout waiting for reshare.  I can't recreate the fail locally - the test always passes.  Tthere is something odd happening with `rotate_network_keyshare` being called an extra time, but that doesn't actually cause anything bad to happen.

crates/threshold-signature-server/src/validator/tests.rs
dvdplm: It was merely an idea sparked by a remote memory! Feel free to pursue (or not) in a separate PR. :)

dvdplm:
I'd just like to give kudos for the *excellent* PR description here. Allows even people unfamiliar with the code to read and understand. 🥇

JesseAbram (approved):

HCastano (approved):
Nice work 💪 

Definitely not an easy PR to get through the finish line!

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -130,25 +128,33 @@ pub async fn create_clients(
 /// A way to specify which chainspec to use in testing
 #[derive(Copy, Clone, PartialEq)]
 pub enum ChainSpecType {
-    /// The development chainspec, which has 3 TSS nodes
-    Development,
```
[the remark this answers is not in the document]
ameba23: It is never used  in tests - im not sure whether we might find some use for it in the future.  I removed it because i thought from the comment that there wasn't much point in having a 3 node setup available because signing would not be possible.  But i realise now that the comment is just outdated and that that is the chainspec we use for docker-compose etc.

Happy to put it back in if you think it would be useful.

PR #1148, opened 2024-11-05
Integration oracle test

JesseAbram:
This provides an end to end integration test of oracle data, using the example oracle-data program in the programs repo


crates/testing-utils/src/constants.rs
```diff
@@ -83,6 +83,8 @@ pub const TEST_PROGRAM_WASM_BYTECODE: &[u8] =
 pub const FAUCET_PROGRAM: &[u8] = include_bytes!("../faucet_program.wasm");
 /// `infinite_loop.wasm` from the `programs` repo.
 pub const TEST_INFINITE_LOOP_BYTECODE: &[u8] = include_bytes!("../infinite_loop.wasm");
+/// `infinite_loop.wasm` from the `programs` repo.
```
ameba23, line 86:
```suggestion
/// `oracle_example.wasm` from the `programs` repo.
```

ameba23 (approved):
:100: :mirror_ball:

I guess we could also wait to block 100 and check it fails, but since we have tests for the program itself in the programs repo, we probably don't need to do that here as well.

JesseAbram:
> 💯 🪩
> 
> I guess we could also wait to block 100 and check it fails, but since we have tests for the program itself in the programs repo, we probably don't need to do that here as well.

ya totally too long and we have the tests for it in many other ways

PR #1153, opened 2024-11-06
Update programs to accept multiple oracle data

JesseAbram:
Updates core to allow for multiple oracle inputs

pairs with https://github.com/entropyxyz/programs/pull/94


crates/client/src/client.rs
```diff
@@ -210,14 +210,14 @@ pub async fn store_program(
     program: Vec<u8>,
     configuration_interface: Vec<u8>,
     auxiliary_data_interface: Vec<u8>,
-    oracle_data_pointer: Vec<u8>,
+    oracle_data_pointer: Vec<Vec<u8>>,
```
ameba23, line 213:
```suggestion
    oracle_data_pointers: Vec<Vec<u8>>,
```

pallets/programs/src/lib.rs
```diff
@@ -263,7 +272,7 @@ pub mod pallet {
                     bytecode: new_program.clone(),
                     configuration_schema: configuration_schema.clone(),
                     auxiliary_data_schema: auxiliary_data_schema.clone(),
-                    oracle_data_pointer: oracle_data_pointer.clone(),
+                    oracle_data_pointers: oracle_data_pointers.clone(),
```
ameba23, line 275: Not directly related to this PR, but i can't see where we check that each oracle data pointer is not longer than `MaxOracleKeyLength`.

ameba23 (approved):
:1st_place_medal: :t-rex:

crates/threshold-signature-server/Cargo.toml
```diff
@@ -48,7 +48,7 @@ entropy-client={ version="0.3.0-rc.1", path="../client", default-features=false,
 ] }
 
 # Programs
-entropy-programs-runtime="0.10.0"
+entropy-programs-runtime={ git="https://github.com/entropyxyz/programs.git", branch="master" }
```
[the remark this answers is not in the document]
JesseAbram: mmm maybe release checklist? maybe an issue?

pallets/programs/src/lib.rs
JesseAbram: yes I was going to make an issue but pretty much either or is fine, since the oracle data is a bounded vec, so either we check the length or we check that it exists (in which case a length check is not needed imo, will make an issue

JesseAbram: https://github.com/entropyxyz/entropy-core/issues/1166

PR #1162, opened 2024-11-13
Add option to start in a pre-jumpstarted state to genesis config

ameba23:
This is a sub-PR of https://github.com/entropyxyz/entropy-core/pull/1147 and targets that branch.

To speed things up in our signing tests we mock the jumpstart DKG and use pre-generated keyshares.  Changes in #1147 mean that the way we were mocking the jumpstart no longer works, because sending the jumpstart extrinsic will trigger an actual jumpstart DKG.

This fixes that by offering the option to start the chain in a pre-jumpstarted state.  

The pre-generated keyshares are still used, but since we know who the initial signers are, we only need one set and can remove the extra ones we created to get around the problem of DKG signers being chosen randomly.

It should also make the signing tests slightly faster as we no longer need the jump start and jump start confirmation extrinsics to go through before we can sign.

I am also hoping it will shed light on the problems i am having resharing, as we wont have the problem of needing to time the reshare to happen after DKG.

The downside is, i couldn't figure out how to do this without adding an extra chainspec type, which complicates the code a bit.


ameba23:
I am having a strange problem following making these changes to the chainspec.

In tests, if i make any api or rpc call to the chain, even something as simple as getting current block number with rpc, i get this error:

```
thread 'user::tests::signature_request_with_derived_account_works' panicked at crates/threshold-signature-server/src/helpers/tests.rs:304:89:
called `Result::unwrap()` on an `Err` value: GenericSubstrate(Rpc(ClientError(RestartNeeded(Transport(i/o error: Connection reset by peer (os error 104)
```

I have tried building this branch with docker compose, and everything is working fine, i can register and sign messages with the test CLI.  So it seems it related to the test setup, not the chainspec itself.

Issue #1163, opened 2024-11-13
Add JSON output to select entropy-test-cli commands


cooldracula:
_Why is this issue relevant?_

The entropy-test-cli is really useful in DevOps' ansible scripts for starting up a network and running basic smoke tests so we can run through the same [release checklist](https://github.com/entropyxyz/entropy-core/blob/23946aa2dd56e78a37d7fe78d0e3c5c5a132fef8/RELEASE_CHECKLIST.md#release-branch-and-local-network-checks) as core, but on staging infrastructure.

It would be easier to automate these steps if there was JSON output for the relevant commands.  This would let us reliably check that a command succeeded and assert our smoke tests are passing.  

_What steps are required to resolve this?_

JSON Output in entropy-test-cli for the following commands :
- status
   - i.e a success status if network is reachable and active. 
- jumpstart-network
   - i.e. a success status once network is jumpstarted. Bonus would be an error status plus message if there was an issue.  
- register
  - a success status and the verifying key
- sign
  - a success status and the transaction details

_Does this change the spec? HTTP, extrinsic, or storage? Is it breaking? Clearly describe the new interface._

It does not.

ameba23:
This is underway.

Worth noting though that you can already programmatically determine whether a command was successful or not with the process exit status code.  All commands will exit with 0 on success and 1 on failure.  So for commands like jumpstart, update-programs, remove-program, etc, which do not give you any information in the happy case, this should already be enough to use in a script.

ameba23:
@cooldracula this will get closed when i merge the PR, but please re-open if something else is needed, specifically:
- If you need a published (crates.io) version of the cli
- If you need to do different things in a script depending on the error type.  We could give different exit codes.
- If you need to be able to get details of the validators / TSS nodes (this would be easy to add).
- If you need to be able to get the current jumpstart status. Separate issue, but just so you are aware - sometimes i have found that the jumpstart command errors with a timeout, but the network eventually does jumpstart.

PR #1164, opened 2024-11-13
Add json output to test CLI command output

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1163

This adds optional JSON output to the test CLI.

This is activated with the `--json` or `-j` command line argument.  Note that it must be given before the subcommand like this: `entropy-test-cli --json status` and not `entropy-test-cli status --json`.

Currently not all available information is given, for example the status command gives you the hashes of stored programs and account ids of registered accounts, but not the additional metadata you will see in the non-json output, such as which programs are associated with which account, who stored a program, how many times a program was used, etc.  These fields could easily be added so please say if they would be useful. 

Generally binary data is represented as hex.

### Example output:
```
$ entropy-test-cli --json status
{
  "accounts": [
    "029d00ff44b70613228963e45a05fe743ea92beb8ece19d48fb6c60d66386fe03b",
    "024609b84f24cdf494c356e48390927df1fbd16cf893d21c9774561b9805a006ae",
    "03cecec00e1c008a1c16c8cbb175b8de3eb10ff03fd79ed82a04e4caa3d21c159d",
    "02c1fd5c996e716f4ad6a3651861ce0efc4db3da88bbff7b84328ecb772f4eddaf",
    "0229e6718ed1d9a1568b4d087f66a00612133b4652b9c03ae9ba62c1ead33dd5d4",
    "03dd0713f53f162e4ce01b064db664682a270ac3a17bccc47ea670ac2a1037f306",
    "037363e1634cd106a78027fa4b160e6430c52ddef27d60828f2483c8e203f56e86",
    "035c0e76a179f5340303e9b219a983a494ad80d2b16d422bd34173ee0281f2a08c",
    "032fc02cb38a2dc5f5bee6fb68600aa9b05f4e18b1b196c1f31d1fde21c06d1e3d",
    "036f424bacf0252ddc470dd797cd511ca5bd5c37845f7a0bef60c8afaedf176fd4",
    "03cc90e7fbcace6b2945ba6718a7e04dc296f76a3788eab223f9e666360829f385",
    "03d53398bf3efa12edebc212959b3b2442612b74c0f0c45c209ec9a664731dc66f",
    "038ad84e7bfdea764737319123f60019365616c48b4c1ef5230fd05e22390c702c",
    "03aee03ad9862e9f31d06f7d1b4b388ad1c66152ad17f919fc16fcc75929b08db3"
  ],
  "programs": [
    "3289435006003c3c83da5582d779d810c343f177fb1fecddbaca5d326ef9d9b3",
    "3a1d45fecdee990925286ccce71f78693ff2bb27eae62adf8cfb7d3d61e142aa",
    "0000000000000000000000000000000000000000000000000000000000000000"
  ]
}

$ entropy-test-cli --json register ../testing-utils/template_barebones.wasm -m //Peg
"035c0e76a179f5340303e9b219a983a494ad80d2b16d422bd34173ee0281f2a08c"

$ entropy-test-cli --json sign "035c0e76a179f5340303e9b219a983a494ad80d2b16d422bd34173ee0281f2a08c" "this is a message"
{
  "signature": "46C2821E3FD730D0B0030CBD89C7AE9265A31B7863C207E6AC971B349D52D5C65FEF136F6D80E375769D47C07CA6F6C8EC0344C7B1B9591DCFF0EE14875D495B",
  "recovery_id": 1
}
```

For commands which finish successfully with no output, an empty JSON object is displayed: `{}`. Im not sure if it would be more appropriate to just give no ouput.

## Errors

Errors are currently **not** displayed as JSON. If a command fails, the exit status will be `1`, and the original error message will be displayed to standard error.  If needed we could wrap this in `{ "error": "error message" }`, but actually serializing the errors to JSON is probably going to be tricky as i imagine they will not all implement `Serialize`.

### DIFF
--- a/crates/test-cli/src/lib.rs
+++ b/crates/test-cli/src/lib.rs
@@ -258,7 +258,7 @@ pub async fn run_command(
             if cli.json {
                 Ok(serde_json::to_string_pretty(&verifying_key)?)
             } else {
-                Ok(format!("Verifying key: {},\n{:?}", hex::encode(verifying_key), registered_info))
+                Ok(format!("Verifying key: {},\n{:?}", verifying_key, registered_info))
             }
         },
         CliCommand::Sign { signature_verifying_key, message, auxilary_data, mnemonic_option } => {
--- a/crates/test-cli/src/lib.rs
+++ b/crates/test-cli/src/lib.rs
@@ -20,7 +20,9 @@ use colored::Colorize;
 use entropy_client::{
     chain_api::{
         entropy::runtime_types::{
-            bounded_collections::bounded_vec::BoundedVec, pallet_registry::pallet::ProgramInstance,
+            bounded_collections::bounded_vec::BoundedVec,
+            pallet_programs::pallet::ProgramInfo,
+            pallet_registry::pallet::{ProgramInstance, RegisteredInfo},
         },
         EntropyConfig,
     },
@@ -32,7 +34,7 @@ use entropy_client::{
 };
 pub use entropy_shared::PROGRAM_VERSION_NUMBER;
 use sp_core::{sr25519, Hasher, Pair};
-use sp_runtime::traits::BlakeTwo256;
+use sp_runtime::{traits::BlakeTwo256, Serialize};
 use std::{fs, path::PathBuf};
 use subxt::{
     backend::legacy::LegacyRpcMethods,
@@ -46,7 +48,7 @@ use subxt::{
     about = "CLI tool for testing Entropy",
     long_about = "This is a CLI test client.\nIt requires a running deployment of Entropy with at least two chain nodes and two TSS servers."
 )]
-struct Cli {
+pub struct Cli {
     #[clap(subcommand)]
     command: CliCommand,
     /// The chain endpoint to use.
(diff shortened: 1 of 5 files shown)


ameba23:
Since this is a breaking change it is paired with a PR in programs to update the CLI for deploying programs, which has `entropy-test-cli` as a dependency.  https://github.com/entropyxyz/programs/pull/96

crates/test-cli/src/lib.rs
```diff
@@ -186,15 +191,22 @@ enum CliCommand {
     },
 }
 
+impl Cli {
+    fn log(&self, text: String) {
```
JesseAbram, line 195: curious why this was needed and not just a println macro?

JesseAbram (approved):

crates/test-cli/src/lib.rs
ameba23: yeah its a wierd way of doing it. just so it only logs when the json option is false, to prevent the log lines being parsed as json.

cooldracula:
This looks good to me, thank you!

cooldracula (approved):

PR #1173, opened 2024-11-19
Add TSS endpoint to get TDX quote

ameba23:
Part of https://github.com/entropyxyz/entropy-core/issues/982 - see https://github.com/entropyxyz/entropy-core/issues/982#issuecomment-2456757856

This adds an endpoint to the TSS server which allows the operator to get a quote for use in the `validate`, `change_endpoint`, and `change_tss_accounts` extrinsics.

In doing this i noticed a possible problem with our design.  I have made an issue: https://github.com/entropyxyz/entropy-core/issues/1174

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -39,7 +39,9 @@ use crate::{
     },
     client::entropy::staking_extension::events::{EndpointChanged, ThresholdAccountChanged},
     substrate::{get_registered_details, submit_transaction_with_pair},
-    user::{get_all_signers_from_chain, get_validators_not_signer_for_relay, UserSignatureRequest},
+    user::{
+        self, get_all_signers_from_chain, get_validators_not_signer_for_relay, UserSignatureRequest,
+    },
     Hasher,
 };
 
@@ -426,6 +428,9 @@ async fn jumpstart_inner(
 /// some point in the near future.
 ///
 /// The returned `nonce` must be used when generating a `quote` for the chain.
+///
+/// This wraps [user::request_attestation] to convert the error to a [ClientError] consistant with
+/// other functions in this module
 #[tracing::instrument(
     skip_all,
     fields(
@@ -435,19 +440,7 @@ async fn jumpstart_inner(
 pub async fn request_attestation(
     api: &OnlineClient<EntropyConfig>,
     rpc: &LegacyRpcMethods<EntropyConfig>,
-    attestee: sr25519::Pair,
-) -> Result<Vec<u8>, ClientError> {
-    tracing::debug!("{} is requesting an attestation.", attestee.public());
-
-    let request_attestation = entropy::tx().attestation().request_attestation();
-
-    let result =
-        submit_transaction_with_pair(api, rpc, &attestee, &request_attestation, None).await?;
-    let result_event = result
-        .find_first::<entropy::attestation::events::AttestationIssued>()?
-        .ok_or(crate::errors::SubstrateError::NoEvent)?;
-
-    let nonce = result_event.0;
-
-    Ok(nonce)
+    attestee: &sr25519::Pair,
+) -> Result<[u8; 32], ClientError> {
+    Ok(user::request_attestation(api, rpc, attestee).await?)
 }
--- a/crates/test-cli/src/lib.rs
+++ b/crates/test-cli/src/lib.rs
@@ -189,6 +189,14 @@ enum CliCommand {
         #[arg(short, long)]
         mnemonic_option: Option<String>,
     },
+    /// Request a TDX quote from a TS server and write it to a file
+    GetTdxQuote {
+        /// The socket address of the TS server, eg: 127.0.0.1:3002
+        tss_endpoint: String,
+        /// The filename to write the quote to. Defaults to quote.dat
+        #[arg(short, long)]
+        output_filename: Option<String>,
+    },
 }
 
 impl Cli {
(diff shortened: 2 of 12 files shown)


crates/client/src/client.rs
```diff
@@ -435,19 +440,7 @@ async fn jumpstart_inner(
...
-    let result_event = result
-        .find_first::<entropy::attestation::events::AttestationIssued>()?
-        .ok_or(crate::errors::SubstrateError::NoEvent)?;
-
-    let nonce = result_event.0;
-
-    Ok(nonce)
+    attestee: &sr25519::Pair,
+) -> Result<[u8; 32], ClientError> {
+    Ok(user::request_attestation(api, rpc, attestee).await?)
```
ameba23, line 445: This function needed to move out of this module because it is behind a feature flag which entropy-tss does not enable.  I have left this wrapper so that there is a version of it with the same error type as the other full-client functions.

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -77,6 +78,30 @@ pub async fn attest(
...
+) -> Result<(StatusCode, Vec<u8>), AttestationErr> {
+    let (signer, x25519_secret) = get_signer_and_x25519_secret(&app_state.kv_store).await?;
+    let api = get_api(&app_state.configuration.endpoint).await?;
+    let rpc = get_rpc(&app_state.configuration.endpoint).await?;
+
+    // Request attestation to get nonce
+    let nonce = request_attestation(&api, &rpc, signer.signer()).await?;
+
+    // We also need the current block number as input
+    let block_number =
```
ameba23, line 96: This is the problem - we cannot know at what block number the node operator is going to submit this quote

crates/test-cli/src/lib.rs
```diff
@@ -189,6 +189,14 @@ enum CliCommand {
         #[arg(short, long)]
         mnemonic_option: Option<String>,
     },
+    /// Request a TDX quote from a TS server and write it to a file
+    GetTdxQuote {
+        /// The socket address of the TS server, eg: 127.0.0.1:3002
```
HCastano, line 194:
```suggestion
        /// The socket address of the TS server, e.g: `127.0.0.1:3001`
```

crates/test-cli/src/lib.rs
```diff
@@ -189,6 +189,14 @@ enum CliCommand {
...
+        tss_endpoint: String,
+        /// The filename to write the quote to. Defaults to quote.dat
+        #[arg(short, long)]
```
HCastano, line 197: Don't think we need a `short` argument here.

crates/test-cli/src/lib.rs
```diff
@@ -189,6 +189,14 @@ enum CliCommand {
         #[arg(short, long)]
         mnemonic_option: Option<String>,
     },
+    /// Request a TDX quote from a TS server and write it to a file
+    GetTdxQuote {
+        /// The socket address of the TS server, eg: 127.0.0.1:3002
+        tss_endpoint: String,
+        /// The filename to write the quote to. Defaults to quote.dat
```
HCastano, line 196:
```suggestion
        /// The filename to write the quote to. Defaults to `quote.dat`.
```

crates/test-cli/src/lib.rs
```diff
@@ -189,6 +189,14 @@ enum CliCommand {
         #[arg(short, long)]
         mnemonic_option: Option<String>,
     },
+    /// Request a TDX quote from a TS server and write it to a file
```
HCastano, line 192:
```suggestion
    /// Request a TDX quote from a TSS server and write it to a file.
```

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -77,6 +78,30 @@ pub async fn attest(
     Ok(StatusCode::OK)
 }
 
+/// Retrieve a quote by requesting a nonce from the chain and return the quote in the HTTP response
+/// body.
+/// This is used by node operators to get a quote for use in the `validate`, `change_endpoint`
+/// and `change_tss_accounts` extrinsics.
+pub async fn get_attest(
```
HCastano, line 85:
```suggestion
pub async fn request_attestation(
```

To match the extrinsic?

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -77,6 +78,30 @@ pub async fn attest(
     Ok(StatusCode::OK)
 }
 
+/// Retrieve a quote by requesting a nonce from the chain and return the quote in the HTTP response
+/// body.
```
HCastano, line 82:
```suggestion
/// body.

```

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -77,6 +78,30 @@ pub async fn attest(
...
+/// This is used by node operators to get a quote for use in the `validate`, `change_endpoint`
+/// and `change_tss_accounts` extrinsics.
+pub async fn get_attest(
+    State(app_state): State<AppState>,
+) -> Result<(StatusCode, Vec<u8>), AttestationErr> {
+    let (signer, x25519_secret) = get_signer_and_x25519_secret(&app_state.kv_store).await?;
+    let api = get_api(&app_state.configuration.endpoint).await?;
+    let rpc = get_rpc(&app_state.configuration.endpoint).await?;
+
+    // Request attestation to get nonce
```
HCastano, line 92:
```suggestion
```

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -77,6 +78,30 @@ pub async fn attest(
...
+    let nonce = request_attestation(&api, &rpc, signer.signer()).await?;
```
HCastano, line 93: As part of the `AttestationIssued` event we return not only the `nonce` but also the `block_number` at which that request was processed. 

I thought it would be useful for that second step, but thinking about it a bit more it might not be.

HCastano (approved):
Thanks for adding this!

Issue #1152, opened 2024-11-06
Check network jumpstart status in new session hander 


ameba23:
The staking extension pallet's `new_session_handler` gets called each session with the new validator set, selects new signers and initiates the reshare process.

I might be mistaken but i can't find anywhere in that function where we check if the number of current signers is zero.  That is, the jump start has not yet happened.

It looks like in this case it will select a complete new signing set for the reshare.

This maybe isn't a problem, because this reshare will fail and nothing should happen.

I checked the jumpstart DKG code in the registry pallet and i can't see that anything would be broken by `pallet_staking_extension::Signers` already containing a bunch of validator IDs when the jumpstart runs - it looks as though it will be just overwritten by the actual signers when they confirm the DKG is complete.

So maybe this is a non-issue, but i wanted to put it up in case it does end up causing some strangeness, or if we wanna put that check in anyway to be sure.

ameba23:
Closed with https://github.com/entropyxyz/entropy-core/pull/1175

PR #1175, opened 2024-11-20
Add no current signer guard

JesseAbram:
Adds a guard for no current signer which will also act as a not jumpstarted guard as jumpstarting sets signers. Doing it this way saves a look up and is a little more protective IMO


ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1152

ameba23 (approved):
:100:

PR #1185, opened 2024-11-25
Reshare in a spawned task and fix propagation pallet rotate keyshares endpoint lookup key

ameba23:
When fixing the reshare test in https://github.com/entropyxyz/entropy-core/pull/1162 i found two problems which mean the resharing will not work in production.

1. The lookup key for the rotate keyshares endpoint is incorrect
2. The http client used by the propagation pallet has a timeout set to two seconds.  When this timeout is met, the http client closes the connection, causing axum to kill the task handling the request. This causes the reshare to stop mid-protocol.  This is fixed here by running the protocol in a spawned task, as we do with DKG.

I have fixed these in https://github.com/entropyxyz/entropy-core/pull/1162 - but since that PR makes quite some changes to our test setup, and still needs some cleaning up, i wanted to put the fixes in a separate PR, so its clear what they are, and so it could be used for a fix for our current testnet (resharing will not work without these fixes).

If we want to fix things with a runtime upgrade and not need to update the TSS code, we could bump up the timeout instead of spawning a task for the reshare.  But i think long term it is better to use a separate task, because amount of time the reshare takes depends on many factors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@ runtime
 - Remove declare synced ([#1134](https://github.com/entropyxyz/entropy-core/pull/1134/))
 - Update programs to accept multiple oracle data ([#1153](https://github.com/entropyxyz/entropy-core/pull/1153/))
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


HCastano (approved):
Nice finds! I know how much of a pain this has been to fix.

I agree with the approach of fixing this in the TSS, seems more reasonable long term.

Also thanks for splitting this out into its own PR.

PR #1186, opened 2024-11-27
Bump synedrion to v0.2.0 release

ameba23:
This bumps synedrion from v0.2.0-beta.0 to the released v0.2.0.  Dependabot would probably eventually do this for us, but im not sure when that will happen and i would like to have it in before we make another release.

There is one small breaking change in the release: [`KeyShare.verifying_key()`](https://docs.rs/synedrion/latest/synedrion/struct.KeyShare.html#method.verifying_key) now returns an `Option<VerifyingKey>`.

And keyshare format has changed so pre-generated keyshares need updating.


crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -148,6 +149,10 @@ async fn test_reshare() {
         if new_signer_ids != old_signer_ids {
             break;
         }
+        if i > 100 {
```
ameba23, line 152: This is not relevant to this PR, but it was missing from https://github.com/entropyxyz/entropy-core/pull/1185 - and i got a CI error in this PR which i think was caused by it: https://github.com/entropyxyz/entropy-core/pull/1185

dvdplm (approved):

HCastano (approved):
I agree with David here, we should have the test deadline handled a bit better.

PR #1192, opened 2024-12-02
Fix error handling for TDX quote generation in production

ameba23:
In `entropy-tss`, real (non-mock) TDX quotes are currently only generated when the `production` feature flag is enabled.

Unfortunately that means the code behind that flag is not tested in CI.  I recently set up a test network on TDX hardware to test attestation, and `entropy-tss` would not compile with the `production` feature flag enabled due to the error type which `config-tsm::generate_quote` having changed. This PR fixes that.

I have also made an issue to have the error type improved before the next release of `configfs-tsm`: https://github.com/entropyxyz/configfs-tsm/issues/7

We could also consider adding `cargo check -p entropy-tss --feature production` to our CI.

### DIFF
--- a/crates/threshold-signature-server/src/attestation/api.rs
+++ b/crates/threshold-signature-server/src/attestation/api.rs
@@ -186,7 +186,8 @@ pub async fn create_quote(
         context,
     );
 
-    Ok(configfs_tsm::create_quote(input_data.0)?)
+    Ok(configfs_tsm::create_quote(input_data.0)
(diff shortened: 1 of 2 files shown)


HCastano (approved):
👍 

Yeah we definitely should be checking all the features during CI builds. I say we could even do it as part of this PR, but since we're not really testing on TDX hardware wait a bit to save ourselves the extra CI run time for now.

PR #1194, opened 2024-12-03
Add oracle data pointer check

JesseAbram:


pallets/transaction-pause/Cargo.toml
```diff
@@ -22,6 +22,7 @@ sp-std            ={ version="14.0.0", default-features=false }
 smallvec="1.13.2"
 
 pallet-balances={ version="29.0.0" }
+pallet-oracle  ={ version='0.3.0', path='../oracle', default-features=false }
```
ameba23, line 25: This is needed because the oracle config is now part of the programs config, and programs is a dependency of transaction-pause right?

ameba23 (approved):
:100: Good move, having checks like this is definitely a good thing

pallets/programs/src/lib.rs
```diff
@@ -218,10 +218,14 @@ pub mod pallet {
         ProgramAlreadySet,
         /// User owns too many programs.
         TooManyProgramsOwned,
-        /// Program is being used by an account
+        /// Program is being used by an account.
```
ameba23, line 221: I think adding these is gonna make @HCastano happy

pallets/transaction-pause/Cargo.toml
JesseAbram: yup exactly

pallets/programs/src/lib.rs
JesseAbram: yes, why I did it, but I also rolled my eyes while doing it

PR #1198, opened 2024-12-04
Test CLI command to retrieve quote and change endpoint / TSS account in one command

ameba23:
I am using the `change_endpoint` extrinsic to test attestation (by changing the endpoint to the same as what it was before, just to check the quote can be validated).

The test-cli has a command to retrieve a quote and write it to a file, and and another command to change the endpoint with a quote given as a string.  Its currently a bit tricky to convert the quote file to a string as its binary not utf8. Since these two things generally need to be done one after the other, this PR makes a single command which gets the quote and uses it in the `change_endpoint` or `change_threshold_accounts` extrinsic.

This PR also has some stuff tacked on which should really be in separate PRs:
- I have also made some minor changes to the error handling / arguments of the `change_endpoint` / `change_threshold_accounts` client functions.
- A little refactor of how we handle mnemonics in the test-cli.
- A fix for the `get-tdx-quote` command to make it also take the context in which the quote will be used.


crates/test-cli/src/lib.rs
```diff
@@ -713,3 +663,35 @@ impl StatusOutput {
...
+    let mnemonic = if let Some(mnemonic) = mnemonic_option {
+        mnemonic
+    } else {
+        std::env::var("DEPLOYER_MNEMONIC")
+            .map_err(|_| anyhow!("A mnemonic must be given either by the command line option or DEPLOYER_MNEMONIC environment variable"))?
+    };
+    Ok(<sr25519::Pair as Pair>::from_string(&mnemonic, None)?)
+}
+
+/// This is the same as [QuoteContext] but implements [ValueEnum]
```
ameba23, line 678: Its annoying to have to do this but implementing `clap::ValueEnum` for `QuoteContext` would mean having clap as a dependency of `entropy-shared`.

JesseAbram (approved):

PR #1204, opened 2024-12-06
Add TDX test network chainspec

ameba23:
This add a chainspec type for the setup i am currently using for a simple test network with the TSS servers running in TDX virtual machines.

It is just the development chainspec with a custom TSS endpoint (we need TSS nodes to have their public IP address), custom PCK certificates (we can't use the mock ones), and a custom accepted MRTD value from the entropy-tss VM image i am currently using.

I don't want to hardcode the IP address as i think it will probably change (hopefully we find something cheaper!), so it is read from an environment variable.

Currently the PCK is hard-coded.  Ideally i would like to read it in from a certificate file, using the same code that the staking extension pallet uses to extract and verify PCKs from certificates. But this would mean having the staking pallet be a direct dependency of the node CLI, which is kinda weird.  Maybe that code should be extracted to a separate crate. Since this code may change anyway if we switch to using [dcap-qvl](https://docs.rs/dcap-qvl), i am going to kick this decision down the road.

For details of how the TSS nodes are deployed see https://github.com/entropyxyz/tdx-build-system/issues/5

### DIFF
--- a/node/cli/src/chain_spec/dev.rs
+++ b/node/cli/src/chain_spec/dev.rs
@@ -13,7 +13,9 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-use crate::chain_spec::{get_account_id_from_seed, provisioning_certification_key, ChainSpec};
+use crate::chain_spec::{
+    get_account_id_from_seed, provisioning_certification_key, ChainSpec, MrtdValues,
+};
 use crate::endowed_accounts::endowed_accounts_dev;
 
 use entropy_runtime::{
@@ -123,6 +125,7 @@ pub fn development_config() -> ChainSpec {
             vec![],
             get_account_id_from_seed::<sr25519::Public>("Alice"),
             devnet_four_node_initial_tss_servers(),
+            None,
         ))
         .build()
 }
@@ -148,6 +151,7 @@ pub fn devnet_local_four_node_config() -> crate::chain_spec::ChainSpec {
             vec![],
             get_account_id_from_seed::<sr25519::Public>("Alice"),
             devnet_local_docker_four_node_initial_tss_servers(),
+            None,
         ))
         .build()
 }
@@ -169,6 +173,7 @@ pub fn development_genesis_config(
         String,
         BoundedVecEncodedVerifyingKey,
     )>,
+    accepted_mrtd_values: Option<MrtdValues>,
 ) -> serde_json::Value {
     // Note that any endowed_accounts added here will be included in the `elections` and
     // `technical_committee` genesis configs. If you don't want that, don't push those accounts to
@@ -283,10 +288,10 @@ pub fn development_genesis_config(
             max_instructions_per_programs: INITIAL_MAX_INSTRUCTIONS_PER_PROGRAM,
             total_signers: TOTAL_SIGNERS,
             threshold: SIGNER_THRESHOLD,
-            accepted_mrtd_values: vec![
+            accepted_mrtd_values: accepted_mrtd_values.unwrap_or(vec![
                 BoundedVec::try_from([0; 48].to_vec()).unwrap(),
                 BoundedVec::try_from([1; 48].to_vec()).unwrap(),
-            ],
+            ]),
             ..Default::default()
         },
         "programs": ProgramsConfig {
--- a/node/cli/src/chain_spec/mod.rs
+++ b/node/cli/src/chain_spec/mod.rs
@@ -48,7 +48,10 @@ use serde_json::json;
 use sp_authority_discovery::AuthorityId as AuthorityDiscoveryId;
 use sp_consensus_babe::AuthorityId as BabeId;
 use sp_core::{sr25519, Pair, Public};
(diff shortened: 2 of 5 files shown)


node/cli/src/chain_spec/dev.rs
```diff
@@ -283,10 +288,10 @@ pub fn development_genesis_config(
             max_instructions_per_programs: INITIAL_MAX_INSTRUCTIONS_PER_PROGRAM,
             total_signers: TOTAL_SIGNERS,
             threshold: SIGNER_THRESHOLD,
-            accepted_mrtd_values: vec![
+            accepted_mrtd_values: accepted_mrtd_values.unwrap_or(vec![
```
ameba23, line 291: Since in most cases we use the mock values, i've made this an option which if not given will use the mock values.

node/cli/src/chain_spec/mod.rs
```diff
@@ -215,3 +219,6 @@ pub fn authority_keys_from_seed(
         get_from_seed::<AuthorityDiscoveryId>(seed),
     )
 }
+
+/// Accepted build time measurement values for TDX attestation
+pub type MrtdValues = Vec<BoundedVec<u8, ConstU32<48>>>;
```
ameba23, line 224: This type alias is already present in the parameters pallet, so possibly we could avoid duplicating it by having that as a dependency, or putting it in entropy-shared.

JesseAbram (approved):

node/cli/src/chain_spec/tdx_testnet.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub static ref PCK: BoundedVecEncodedVerifyingKey = vec![
+    2, 166, 103, 136, 58, 157, 155, 124, 186, 75, 81, 133, 87, 255, 233, 182, 192, 125, 235, 230,
+    121, 173, 147, 108, 47, 190, 240, 181, 75, 181, 31, 148, 128,
+    ].try_into().unwrap();
+}
+
+fn tdx_devnet_four_node_initial_tss_servers(
+) -> Vec<(sp_runtime::AccountId32, TssX25519PublicKey, String, BoundedVecEncodedVerifyingKey)> {
+    let tss_ip = std::env::var("ENTROPY_TESTNET_TSS_IP")
+        .expect("ENTROPY_TESTNET_TSS_IP environment variable to be set");
```
HCastano, line 41: Why panic here instead of having a fallback to a default IP?

node/cli/src/chain_spec/tdx_testnet.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/// The configuration used for development.
+///
+/// Since Entropy requires at two-of-three threshold setup, and requires an additional relayer node,
+/// we spin up four validators: Alice, Bob, Charlie and Dave.
+pub fn development_config() -> ChainSpec {
+    ChainSpec::builder(wasm_binary_unwrap(), Default::default())
+        .with_name("Development")
+        .with_id("dev")
```
HCastano, line 81: These should be updated to be "TDX-something"

node/cli/src/chain_spec/tdx_testnet.rs
```diff
@@ -0,0 +1,97 @@
...
+    );
+
+    vec![alice, bob, charlie, dave]
+}
+
+/// The configuration used for development.
+///
+/// Since Entropy requires at two-of-three threshold setup, and requires an additional relayer node,
+/// we spin up four validators: Alice, Bob, Charlie and Dave.
+pub fn development_config() -> ChainSpec {
```
HCastano, line 78: Same here

node/cli/src/chain_spec/tdx_testnet.rs
```diff
@@ -0,0 +1,97 @@
...
+use sp_runtime::BoundedVec;
+
+/// The build time measurement value from the current entropy-tss VM images
+const ACCEPTED_MRTD: [u8; 48] = [
+    145, 235, 43, 68, 209, 65, 212, 236, 224, 159, 12, 117, 194, 197, 61, 36, 122, 60, 104, 237,
+    215, 250, 254, 138, 53, 32, 201, 66, 166, 4, 164, 7, 222, 3, 174, 109, 197, 248, 127, 39, 66,
+    139, 37, 56, 135, 49, 24, 183,
+];
+
+lazy_static::lazy_static! {
```
HCastano, line 31: A comment explaining how to generate this would be helpful

node/cli/src/chain_spec/mod.rs
HCastano: I'd say it's fine for now, if we use it more then we can move it somewhere better

HCastano (approved):
Couple comments, but looks good

node/cli/src/chain_spec/tdx_testnet.rs
ameba23: I'm not sure what default IP we can fallback to.  `localhost` wont work because these are the IPs which the TSS nodes use to communicate with each other, and from the CVM point of view localhost is not the same as localhost on the host.  I don't really want to hardcode the public IP of our current box.

Issue #1208, opened 2024-12-09
Staking pallet's `FailedAttestationCheck` error could be more descriptive


ameba23:
When we submit an attestation there are a number of reasons it could fail to validate - bad build time measurement, invalid signature, badly formed quote, bad input data.  These are variants of the attestation pallet's `Error`: https://github.com/entropyxyz/entropy-core/blob/808645f6af26c31a7fdc087bd242b52d02d0e861/pallets/attestation/src/lib.rs#L123

But the error we get back from the staking pallet on a failed attestation, will always be `FailedAttestationCheck`. https://github.com/entropyxyz/entropy-core/blob/808645f6af26c31a7fdc087bd242b52d02d0e861/pallets/staking/src/lib.rs#L431

If this was a 'normal' rust error, i would suggest having that error variant contain the attestation pallet `Error` as a tuple.  But i've had issues trying to do stuff like this with `#[pallet::Error]` annotated errors - eg: see the conversion of a `PckParseVerifyError` to a staking pallet error: https://github.com/entropyxyz/entropy-core/blob/808645f6af26c31a7fdc087bd242b52d02d0e861/pallets/staking/src/lib.rs#L355

PR #1209, opened 2024-12-10
Extract PCK certificate chain from quotes

ameba23:
I recently figured out the PCK certificate chain is actually included in the quote, encoded as PEM. I have an open PR to tdx-quote which allows getting the certificate chain from a quote, and i have moved the cert chain parsing logic over there: https://github.com/entropyxyz/tdx-quote/pull/12

This means that the extrinsics `validate` and `change_threshold_accounts` don't need to take both a quote and a certificate chain - just the quote.

My proposal is to change `AttestationHandler::verify_quote` to validate the quote with whatever PCK can be extracted from its certificate chain, and return the PCK upon successful verification - the staking pallet can then store it in the `ServerInfo`.

This also adds a `production` feature flag to `entropy` which propagates down to `pallet-attestation`. If present, real PCK cert chain verification will be used (previously we always mocked certificate chain verification).  By default it is not used.

Closes https://github.com/entropyxyz/entropy-core/issues/1208 by adding a custom error type to the attestation handler trait.

If this gets approved, i think there is maybe some refactoring that could be done as a followup - i'm not totally sure but i don't think the staking extension pallet needs to be a dependency of the attestation pallet anymore.  Which means we can have the attestation pallet be a dependency of the staking extension pallet, and move the `AttestationHandler` trait from `entropy-shared` to the attestation pallet.


runtime/Cargo.toml
```diff
@@ -273,3 +273,6 @@ try-runtime=[
   "pallet-utility/try-runtime",
   "pallet-vesting/try-runtime",
 ]
+# Enables real PCK certificate chain verification - which means TSS nodes much be runnning on TDX
```
HCastano, line 276:
```suggestion
# Enables real PCK certificate chain verification - which means TSS nodes must be runnning on TDX
```

Cargo.lock
HCastano, line 1: Did you do a full `cargo update` here? This seems like a lot of changes

node/cli/Cargo.toml
```diff
@@ -119,3 +119,6 @@ runtime-benchmarks=[
 # Enable features that allow the runtime to be tried and debugged. Name might be subject to change
 # in the near future.
 try-runtime=["entropy-runtime/try-runtime", "try-runtime-cli/try-runtime"]
+# Enables real PCK certificate chain verification - which means TSS nodes much be runnning on TDX
```
HCastano, line 122:
```suggestion
# Enables real PCK certificate chain verification - which means TSS nodes must be runnning on TDX
```

HCastano (commented):
Took a sneak peek here, seems reasonable. I like that we're able to move some of this out to the `tdx-quote` crate

Cargo.lock
ameba23: Yeah that was an accident.  I couldn't update the git version tdx-quote with `cargo update tdx-quote` because some crates were using the git version and others the release version.  So i lazily just did `cargo update` without thinking of the implications.  Im not sure whether to revert it or just see if things still work and leave it in if the do.

ameba23: In the end i have reverted the lockfile as there was cargo-deny related license issues popping up

crates/client/Cargo.toml
```diff
@@ -38,10 +38,12 @@ js-sys={ version="0.3.74", optional=true }
 tokio ={ version="1.42", features=["time"] }
 
 [dev-dependencies]
-serial_test          ="3.2.0"
-sp-keyring           ="34.0.0"
+serial_test="3.2.0"
+sp-keyring="34.0.0"
 entropy-testing-utils={ path="../testing-utils" }
-tdx-quote            ={ version="0.0.1", features=["mock"] }
+tdx-quote={ git="https://github.com/entropyxyz/tdx-quote.git", branch="peg/cert-chain-parse", features=[
```
ameba23, line 44: I'm planning to re-publish this crate if this PR gets approved. Ideally i would also like to fix https://github.com/entropyxyz/tdx-quote/issues/13 and https://github.com/entropyxyz/tdx-quote/issues/11 before publishing.

crates/shared/src/types.rs
```diff
@@ -183,12 +182,37 @@ impl<AccountId> AttestationHandler<AccountId> for () {
...
+        Ok(sp_runtime::BoundedVec::new())
     }
 
     fn request_quote(_attestee: &AccountId, _nonce: [u8; 32]) {}
 }
+
+/// An error when verifying a quote
+#[cfg(not(feature = "wasm"))]
+#[derive(Eq, PartialEq)]
+pub enum VerifyQuoteError {
```
ameba23, line 197: I added this custom error type because i couldn't figure out how to get the useful information out of a `sp_runtime::DispathError` without adding a lot of extra code.

pallets/staking/src/lib.rs
```diff
@@ -345,20 +339,45 @@ pub mod pallet {
...
+        /// PCK certificate chain public key is not well formed
         PckCertificateBadPublicKey,
+        /// Pck certificate could not be extracted from quote
         PckCertificateNoCertificate,
-        FailedAttestationCheck,
     }
 
-    impl<T> From<pck::PckParseVerifyError> for Error<T> {
-        fn from(error: pck::PckParseVerifyError) -> Self {
+    impl<T> From<VerifyQuoteError> for Error<T> {
```
ameba23, line 364: Maybe its possible to remove the need for this `impl` by implementing [`PalletError`](https://docs.rs/frame-support/latest/frame_support/traits/trait.PalletError.html) for `VerifyQuoteError`, so that we can have a tuple enum variant of `pallet::<T>::Error` containing it.  I would maybe try this in a followup - but i am afraid of going over the maximum encoded size of 4 bytes.

pallets/staking/src/lib.rs
```diff
@@ -345,20 +338,45 @@ pub mod pallet {
...
+        fn from(error: VerifyQuoteError) -> Self {
```
HCastano, line 364: Ah, so it was really frustrating trying to debug errors stemming from the quote verification code because they'd all basically be type erased into the pallet error. While this is pretty ugly it is quite useful.

Maybe worth a comment explaining this for future readers.

pallets/staking/src/lib.rs
HCastano: It seems like it's a compile time check anyways, so yeah would be worth trying

crates/shared/src/types.rs
```diff
@@ -165,10 +165,9 @@ pub trait AttestationHandler<AccountId> {
     fn verify_quote(
         attestee: &AccountId,
         x25519_public_key: X25519PublicKey,
-        provisioning_certification_key: BoundedVecEncodedVerifyingKey,
         quote: Vec<u8>,
         context: QuoteContext,
-    ) -> Result<(), sp_runtime::DispatchError>;
+    ) -> Result<BoundedVecEncodedVerifyingKey, VerifyQuoteError>;
```
HCastano, line 170: As part of the doc comments we'll want to mention that this is a PCK that gets returned

HCastano (approved):
Nice work, looks good 👍 

I'd need to double check the diff with `v0.3.0` to be sure, but this probably needs a `CHANGELOG` entry and a "Bump `spec_version`" tag.

> Which means we can have the attestation pallet be a dependency of the staking extension pallet, and move the AttestationHandler trait from entropy-shared to the attestation pallet.

It would be nice to keep this flexible using associated types, but I'd be curious to see what you have in mind for this first before commenting further.

ameba23:
> I'd need to double check the diff with `v0.3.0` to be sure, but this probably needs a `CHANGELOG` entry and a "Bump `spec_version`" tag.

Im still not super clear exactly what kinds of runtime changes mean we need to bump the spec version but i added that label for now as i figured better to have it and not need it than need it and not have it.

crates/shared/src/types.rs
```diff
@@ -183,12 +184,38 @@ impl<AccountId> AttestationHandler<AccountId> for () {
     fn verify_quote(
         _attestee: &AccountId,
         _x25519_public_key: X25519PublicKey,
-        _provisioning_certification_key: BoundedVecEncodedVerifyingKey,
         _quote: Vec<u8>,
         _context: QuoteContext,
-    ) -> Result<(), sp_runtime::DispatchError> {
-        Ok(())
+    ) -> Result<BoundedVecEncodedVerifyingKey, VerifyQuoteError> {
+        // Ok(sp_runtime::BoundedVec::new())
```
HCastano, line 190:
```suggestion
```

CHANGELOG.md
```diff
@@ -41,6 +45,9 @@ runtime
 - Add TSS endpoint to get TDX quote ([#1173](https://github.com/entropyxyz/entropy-core/pull/1173))
 - Add TDX test network chainspec ([#1204](https://github.com/entropyxyz/entropy-core/pull/1204))
 - Test CLI command to retrieve quote and change endpoint / TSS account in one command ([#1198](https://github.com/entropyxyz/entropy-core/pull/1198))
+- In ([#1209]()) a `production` feature flag was added to `entropy` which if enabled will use
+  non-mock verification of PCK certificate chains in TDX quotes, meaning TSS servers must be running
+  on TDX hardware
```
HCastano, line 50:
```suggestion
```

These sections should just have PR titles.

If anything we could maybe mention it in the `Breaking Changes`, but nbd. Can always add it later when prepping the release.

PR #1215, opened 2024-12-12
On-chain unresponsiveness reporting

HCastano:
This is another take on #667 but without subgroups.

This PR adds a new extrinsic, `report_unstable_peer()`, to the Staking pallet allowing
TSS nodes in the signing committee to report other TSS nodes in the committee.

Internally this extrinsic calls out to the Slashing pallet to note these reports.
Eventually something will be done to punish any repeat offenders (probably `chill` them
from the validator set), but that'll be work for a later PR.

I'm in the middle of another PR which adds the triggering mechanism to the TSS itself.


pallets/staking/src/lib.rs
```diff
@@ -90,6 +90,7 @@ pub mod pallet {
         + frame_system::Config
         + pallet_staking::Config
         + pallet_parameters::Config
+        + pallet_slashing::Config
```
HCastano, line 93: I know this is gonna trigger Jesse, but in a future PR I want to move this to be loosely coupled

JesseAbram: idk I just think it makes the code more unreadable and less flexible but wtv

JesseAbram (approved):

pallets/staking/Cargo.toml
```diff
@@ -34,6 +34,7 @@ p256              ={ version="0.13.2", default-features=false, features=["ecdsa"
 rand              ={ version="0.8.5", default-features=false, features=["alloc"] }
 
 pallet-parameters={ version="0.3.0", path="../parameters", default-features=false }
+pallet-slashing={ version="0.3.0", path="../slashing", default-features=false }
```
ameba23, line 37: Are you sure you are not going to run in to problems with needing the staking extension pallet to be a dependency of the slashing pallet in order to call `chill`? Or can you call chill on `pallet-staking` directly without having circular dependencies?

pallets/staking/src/lib.rs
```diff
@@ -735,6 +737,57 @@ pub mod pallet {
...
+            // For reporting purposes we need to know the validator account tied to the TSS account.
+            let reporter_validator_id = Self::threshold_to_stash(&reporter_tss_account)
+                .ok_or(Error::<T>::NoThresholdKey)?;
+            let offender_validator_id = Self::threshold_to_stash(&offender_tss_account)
+                .ok_or(Error::<T>::NoThresholdKey)?;
+
+            // Note: This operation is O(n), but with a small enough Signer group this should be
+            // fine to do on-chain.
+            let signers = Self::signers();
+            ensure!(signers.contains(&reporter_validator_id), Error::<T>::NotSigner);
```
ameba23, line 763: This means relayers cannot use this to report that they were unable to relay a signature request to a member of the signing committee.  Which is fine for now but maybe at some point we will want a separate extrinsic for that purpose.

ameba23 (approved):
:star: Looks great.

My only concern would be possible circular dependency problems - as even if we can call `pallet-staking`'s `chill` from the slashing pallet without needing `pallet-staking-extension`, we are probably anyway going to need stuff from the staking extension pallet there for example to be able to mark a bad signer for removal from the signer set at the next reshare, or remove a TSS node, or something along those lines. So i would probably have put this extrinsic in the slashing pallet.

Thinking about it, i expect your plan is to put a trait in `entropy-shared` in order to do this, like with the `AttestationHandler`.

pallets/staking/src/lib.rs
```diff
@@ -735,6 +737,57 @@ pub mod pallet {
...
+        ) -> DispatchResultWithPostInfo {
+            let reporter_tss_account = ensure_signed(origin)?;
+
+            // For reporting purposes we need to know the validator account tied to the TSS account.
+            let reporter_validator_id = Self::threshold_to_stash(&reporter_tss_account)
+                .ok_or(Error::<T>::NoThresholdKey)?;
+            let offender_validator_id = Self::threshold_to_stash(&offender_tss_account)
+                .ok_or(Error::<T>::NoThresholdKey)?;
+
+            // Note: This operation is O(n), but with a small enough Signer group this should be
```
ameba23, line 760: I would like a formal justification for the capitilization of Signer.  
```suggestion
            // Note: This operation is O(n), but with a small enough signer group this should be
```

HCastano:
> My only concern would be possible circular dependency problems - as even if we can call pallet-staking's chill from the slashing pallet without needing pallet-staking-extension, we are probably anyway going to need stuff from the staking extension pallet there for example to be able to mark a bad signer for removal from the signer set at the next reshare, or remove a TSS node, or something along those lines. 

Yeah so I'm not entirely sure how we'll be doing the last bit (calling `chill()`), but
I'm kinda hoping that the `Offences` pallet (which is what Slashing uses under the hood)
will just do that bit for us. If we have to interface with Staking directly then yes it
might be a problem.

> So i would probably have put this extrinsic in the slashing pallet.

The problem with putting it in the Slashing pallet is that we don't have access to the
types needed to identify validators or signers (e.g `ThresholdToStash`, `Signers`).


> Thinking about it, i expect your plan is to put a trait in entropy-shared in order to do this, like with the AttestationHandler.

Yeah I do want to do this eventually so the Slashing mechanism can be somewhat flexible
and configurable. This could also help with some of the circular dependencies that may
arise.

pallets/staking/Cargo.toml
HCastano: Discussed below

pallets/staking/src/lib.rs
HCastano: Yes that's a good point. I meant to bring it up in the PR description but forgot. The
essence of the question here is whether or not we give relayers the power to report. I'm
not entirely sure yet, but it's something we can decide on later

pallets/staking/src/lib.rs
HCastano: The storage struct is called `Signers`, QED? lol 

Pls don't make me run CI again for this 😅

pallets/staking/src/lib.rs
HCastano: Opened https://github.com/entropyxyz/entropy-core/issues/1223

Issue #1203, opened 2024-12-06
entropy-tss `/info` route to get account ID cannot be used until a connection to chain has been made


ameba23:
In https://github.com/entropyxyz/entropy-core/pull/1128 we remove the ability to pass in a pre-generated mnemonic to entropy-tss.  Instead, it is generated internally and the public key / account ID can be retrieved via the `/info` http route.

This does not effects our tests because we can still use the built-in development mnemonics.  But for testnet the workflow has to be changed.

There was a proposal to still include these public keys in the testnet chainspec, by starting the tss nodes and getting the public keys before compiling the chain.

There is a chicken and egg problem here, because entropy-tss does not start its http server until a successful connection to the chain has been made - it calls the `check_node_prerequisites` function before starting a TCP listener: https://github.com/entropyxyz/entropy-core/blob/d192cdc079684677bb01a566fcd27c24d7422df8/crates/threshold-signature-server/src/main.rs#L112

There are a number of possible ways around this, eg: starting the server before making that check, or starting a simpler version of the server with only the `/info` endpoint, or reading the account ID out from the logs.

But personally i would like to move away from having TSS servers pre-configured at genesis, as id like to keep things as close to production conditions as possible.

@HCastano made a proposal to have an extrinsic which allows us to add TSS details to an existing validator after genesis. I think this is a good move, but we need to be careful about how we handle validators with no associated TSS node - the jump start extrinsic handler would need some changes as it currently picks initial signers from the session pallet's `validators()` getter, not the staking extension's list of TSS nodes. https://github.com/entropyxyz/entropy-core/blob/d192cdc079684677bb01a566fcd27c24d7422df8/pallets/registry/src/lib.rs#L227

I expect there may be similar issues with picking a new signer for reshare in the new session handler, but we could add a jumpstart requirement that all validators must have an associated TSS node at that point.

ameba23:
Related to this, the `check_node_prerequisites` function also checks for a minimum balance for submitting extrinsics

Issue #1214, opened 2024-12-12
Attestation problems when restarting `entropy-tss` nodes


ameba23:
Currently we persist the TSS mnemonic from which account ID and x25519 keypair are derived to disk.

Since these are used in attestation, it is important that they are generated in the trusted domain and never leave.  Even if / when we [figure out how to do encryption with the hardware key](https://github.com/entropyxyz/tdx-build-system/issues/4), we still cannot have these keypairs survive a VM reboot, as there could be some manipulation of the VM image while it is down, and we wouldn't know until we did another attestation.

For this reason i think we need to have these secret keys be stored only in memory and re-generated every time entropy-tss launches.  This means we need to automate the process by which the new public keys are submitted to the chain - as we cannot expect the node operator to be around all hours of the day.

That is, as soon as entropy-tss starts, it should be able to submit an extrinsic with its TSS account ID and x25519 public key.  There are two issues here:

1. We need a funded account to be able to submit the extrinsic.  This would be solved if we get free transactions working.
2. The chain needs to known which stash account the TSS node should be associated with.  Currently it knows this because the host operator uses their stash account to submit the extrinsic.

The tricky thing about (2) is that the TSS node cannot know itself which stash account it is associated with. We can't include this information by customising the VM image, because ideally any change to the VM image will result in different measurements being used in the attestation (although we still haven't figured this part out and maybe there will be a way we can).

My proposal is that we have a local network service running as part of the chain node which the TSS node can communicate with. We can use [QEMU guest forwarding](https://github.com/entropyxyz/tdx-build-system/issues/5#issuecomment-2534845470) to ensure that this service is only exposed to the confidential virtual machine and not accessible publicly.

This service would verify a quote from the TSS server, and then either:
- Send a small amount of funds to it from the stash account so that the TSS server is able to submit extrinsics.  Like a mini-faucet which is only exposed locally.
OR:
- Submit extrinsics from the stash account on behalf of the TSS server, meaning the TSS server itself doesn't need to hold funds.

This is potentially quite a big change to our design, but right now i don't see many other options.

PR #1216, opened 2024-12-13
Non persistent TSS signer and x25519 keypair

ameba23:
`entropy-tss` has two keypairs which define its identity on the network: the TSS account used to sign extrinsics, and the x25519 encryption keypair.  Since both are generated in a confidential virtual machine and their public keys included in the attestation, authenticating with these keypairs proves that the attestation still holds.

However, because they are currently stored on disk, it is possible that the validator operator restarts the virtual machine, makes some modifications, and continues to use these keypairs without needing to make another attestation.  

This PR makes the signing / x25519 keypairs no longer persistent, but generated new every time entropy-tss launches (except in development / testing where test mnemonics are used).  If the `entropy-tss` process is killed for whatever reason, it will not be able to continue to participate in the protocols until the validator operator updates the TSS details using the `change_threshold_accounts` extrinsic. I am hoping to find a way to automate this - see https://github.com/entropyxyz/entropy-core/issues/1214

This PR has implications for our devops flow - the `--setup-only` command line option is no longer available. `entropy-tss` should be run only once, and the public keys retrieved using the `/info` http route.

Closes https://github.com/entropyxyz/entropy-core/issues/1203 by adding a boolean `ready` flag to the application state, which is set to true once the 'prerequisite checks' are complete.  These checks now also include a check the the TSS account id has been registered with the staking pallet, and the balance check is now mandatory (before, only a warning was logged if the TSS account had no funds).  In a non-ready state, all the the HTTP routes relating the the protocols will return an error.

This also has implications for slashing, as a TSS node should not be in a non-ready state for too long. How long is acceptable depends a bit on whether we are able to automate the process of the node getting funded and calling `change_threshold_accounts`.


crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -101,26 +93,6 @@ use crate::{
     validation::EncryptedSignedMessage,
 };
 
-#[tokio::test]
-#[serial]
-async fn test_get_signer_does_not_throw_err() {
```
ameba23, line 106: I think this set is no longer needed as the process of getting a signer from app state is now infallible

crates/threshold-signature-server/src/lib.rs
```diff
@@ -199,14 +205,90 @@ use crate::{
...
+        SubxtAccountId32(self.pair.public().0)
+    }
+
+    /// Get the x25519 public key
+    pub fn x25519_public_key(&self) -> [u8; 32] {
+        x25519_dalek::PublicKey::from(&self.x25519_secret).to_bytes()
+    }
+
+    /// Convenience function to get chain api and rpc
+    pub async fn get_api_rpc(
```
ameba23, line 285: Not related to this PR but i though once being here why not

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -406,45 +281,67 @@ pub async fn check_node_prerequisites(url: &str, account_id: &str) {
...
+                    tracing::error!("Unable to query the account balance of `{}`", &account_id);
+                    "Unable to query account balance".to_string()
+                })?;
+                if !has_minimum_balance {
+                    Err("Minimum balance not met".to_string())?
+                }
+                Ok(())
+            };
+
+            if let Err(error) = backoff::future::retry(backoff.clone(), balance_query).await {
```
ameba23, line 311: This balance check and the one below could maybe be combined into a state machine which makes both checks, but i think for now its ok to make one after the other

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -406,45 +281,67 @@ pub async fn check_node_prerequisites(url: &str, account_id: &str) {
         Ok((api, rpc))
     };
 
-    // Note: By default this will wait 15 minutes before it stops retry attempts.
-    let backoff = backoff::ExponentialBackoff::default();
-    match backoff::future::retry(backoff, connect_to_substrate_node).await {
+    // Never give up trying to connect
+    let backoff = backoff::ExponentialBackoff { max_elapsed_time: None, ..Default::default() };
```
ameba23, line 285: I've not looked at the `backoff` crate in too much detail but i think adding `max_elapsed_time: None` means it will keep checking indefinitely.

crates/threshold-signature-server/src/helpers/validator.rs
```diff
@@ -15,51 +15,29 @@
...
-    get_signer_from_hkdf(&hkdf)
-}
-
 /// Get the PairSigner as above, and also the x25519 encryption keypair for
 /// this threshold server
-pub async fn get_signer_and_x25519_secret(
-    kv: &KvManager,
-) -> Result<(PairSigner<EntropyConfig, sr25519::Pair>, StaticSecret), UserErr> {
-    let hkdf = get_hkdf(kv).await?;
+pub fn get_signer_and_x25519_secret(
```
ameba23, line 32: This is now only used when a `ValidatorName` (eg: `--alice`) is given, or in tests.

crates/threshold-signature-server/src/user/errors.rs
```diff
@@ -179,6 +179,8 @@ pub enum UserErr {
     TooFewSigners,
     #[error("Non signer sent from relayer")]
     IncorrectSigner,
+    #[error("Node has started fresh and not yet successfully set up")]
+    NotReady,
```
ameba23, line 183: If we want to react programmatically to this error (eg: for slashing) we could change the `impl IntoResponse` below to give a some special status code if this variant is present.

crates/threshold-signature-server/src/lib.rs
```diff
@@ -199,14 +205,90 @@ use crate::{
 
 #[derive(Clone)]
 pub struct AppState {
+    /// Tracks whether prerequisite checks have passed.
+    /// This means:
+    /// - Communication has been established with the chain node
+    /// - The TSS account is funded
+    /// - The TSS account is registered with the staking extension pallet
+    ready: Arc<RwLock<bool>>,
```
ameba23, line 213: This `bool` could maybe be replaced with an enum with the various states of readiness to make it easier to determine why the node is not ready: no connection to chain, no funds, or not registered with staking pallet.

JesseAbram:
are we fully set on not keeping this in the kvdb, I thought we were still on the fence about that, if so then I can push removing the whole kvdb as its purpose is now kinda obsolute and everything can be held in state, also probably needs to get devops eyes on this too as it will break their flow

HCastano (commented):
Generally looks good, but I need another review to finish up

ameba23:
> are we fully set on not keeping this in the kvdb, I thought we were still on the fence about that, if so then I can push removing the whole kvdb as its purpose is now kinda obsolute and everything can be held in state, 

Personally, i think it would be great to remove the kvdb.  Im still not totally sure what our best option is for keyshare storage - i am looking into using the SGX seal API which Hang from Phala suggested to us.  But i would suggest for now we keep it in memory only.

> also probably needs to get devops eyes on this too as it will break their flow

I have a call with @vitropy tomorrow to look at the setup we have on the TDX machine, which hopefully will make clear why these changes are needed.

crates/threshold-signature-server/src/helpers/launch.rs
ameba23: I don`t see what is to gain by limiting it.  It will be hard for the operator to see that the `entropy-tss` process has stopped if the VM is still up and running, so i am imagining `entropy-tss` is going to be automatically re-started if it bails, in which case it will continue to attempt to connect anyway.

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -406,45 +281,69 @@ pub async fn check_node_prerequisites(url: &str, account_id: &str) {
...
+                    e.to_string()
+                })?;
+                if !has_minimum_balance {
+                    Err("Minimum balance not met".to_string())?
+                }
+                Ok(())
+            };
+
+            if let Err(error) = backoff::future::retry(backoff.clone(), balance_query).await {
+                tracing::error!("This should never happen because backoff has no permanent errors or maximum timeout: {error}");
```
HCastano, line 312:
```suggestion
                tracing::error!(
                    "This should never happen because backoff has no permanent errors or maximum \
                    timeout: {error}"
                );
```

The long lines are a bit hard to read

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -406,45 +281,69 @@ pub async fn check_node_prerequisites(url: &str, account_id: &str) {
...
+                    .staking_extension()
+                    .threshold_to_stash(subxt::utils::AccountId32(*account_id.as_ref()));
+
+                let _stash_address = query_chain(&api, &rpc, stash_address_query, None)
+                    .await?
+                    .ok_or_else(|| {
+                        tracing::warn!("TSS account ID {account_id} not yet registered on-chain - you need to call `validate` or `change_threshold_accounts`");
+                        SubstrateError::NoEvent})?;
+                Ok(())
+            };
```
HCastano, line 333:
```suggestion
                let _stash_address = query_chain(&api, &rpc, stash_address_query, None)
                    .await?
                    .ok_or_else(|| {
                        tracing::warn!(
                            "TSS account ID {account_id} not yet registered on-chain - you need to \
                            call `validate` or `change_threshold_accounts`"
                        );

                        SubstrateError::NoEvent
                    })?;

                Ok(())
            };
```

Same thing here

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -406,45 +281,69 @@ pub async fn check_node_prerequisites(url: &str, account_id: &str) {
...
+
+            tracing::info!(
+                "Checking if our account ID has been registered on chain `{}`",
+                &account_id
+            );
+            if let Err(error) = backoff::future::retry(backoff, check_for_tss_account_id).await {
+                tracing::error!("This should never happen because backoff has no permanent errors or maximum timeout: {error}");
```
HCastano, line 340: I'd also change this error to be <100 characters

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -406,45 +281,69 @@ pub async fn check_node_prerequisites(url: &str, account_id: &str) {
...
             }
+            tracing::info!("TSS node passed all prerequisite checks and is ready");
+            app_state.make_ready();
         },
-        Err(_err) => {
-            tracing::error!("Unable to establish connection with Substrate node at `{}`", url);
-            panic!("Unable to establish connection with Substrate node.");
+        Err(error) => {
+            tracing::error!("This should never happen because backoff has no permanent errors or maximum timeout: {error:?}");
```
HCastano, line 346: Here too 😄

crates/threshold-signature-server/src/helpers/launch.rs
HCastano: Doesn't look like multiple balance queries like this are handled [natively by subxt](https://github.com/paritytech/subxt/issues/648)

crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -59,15 +57,15 @@ pub async fn attest(
     validate_new_attestation(block_number, &attestation_requests, &app_state.kv_store).await?;
 
     // Check whether there is an attestion request for us
-    if !attestation_requests.tss_account_ids.contains(&signer.signer().public().0) {
+    if !attestation_requests.tss_account_ids.contains(&app_state.pair.public().0) {
```
HCastano, line 60: Can we not use the `account_id()` getter here?

crates/threshold-signature-server/src/helpers/launch.rs
HCastano: By having it spin down after 15 (or some amount) of minutes it becomes a very clear error state for any operators, assuming they have some way to get notified that the process is down.

While it may "just get restarted" by the operator it will also signal to them that some manual action (in this case funding the accounts) needs to be taken.

crates/threshold-signature-server/src/node_info/api.rs
```diff
@@ -36,16 +35,18 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
...
-    let (signer, x25519_secret) = get_signer_and_x25519_secret(&app_state.kv_store).await?;
-    let tss_account = AccountId32(signer.signer().public().0);
-    let x25519_public_key = *x25519_dalek::PublicKey::from(&x25519_secret).as_bytes();
-
-    Ok(Json(TssPublicKeys { x25519_public_key, tss_account }))
+    Ok(Json(TssPublicKeys {
+        ready: app_state.is_ready(),
+        x25519_public_key: app_state.x25519_public_key(),
+        tss_account: app_state.subxt_account_id(),
+    }))
```
HCastano, line 51: If we had a `From` impl here we could do `app_state.into()`. But maybe not really worth since we only do the conversion in a single place

crates/threshold-signature-server/src/node_info/api.rs
```diff
@@ -36,16 +35,18 @@ pub async fn hashes() -> Json<Vec<HashingAlgorithm>> {
 /// Public signing and encryption keys associated with a TS server
 #[derive(Serialize, Deserialize, Clone, Debug, Eq, PartialEq)]
 pub struct TssPublicKeys {
+    /// Indicates that all prerequisite checks have passed
+    pub ready: bool,
     pub tss_account: AccountId32,
```
HCastano, line 40: Similar to the other place, might as well add doc comments here too

crates/threshold-signature-server/src/lib.rs
HCastano: Can you open a `Good First Issue` to go through the codebase and use this?

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -220,12 +220,14 @@ pub async fn sign_tx(
     State(app_state): State<AppState>,
     Json(encrypted_msg): Json<EncryptedSignedMessage>,
 ) -> Result<(StatusCode, Body), UserErr> {
-    let (signer, x25519_secret) = get_signer_and_x25519_secret(&app_state.kv_store).await?;
+    if !app_state.is_ready() {
```
HCastano, line 223: I do worry a bit that we'll forget this check in some new endpoint and it could cause problems. But maybe it'll be quite obvious when somebody tries to do something when the app isn't `ready`

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -98,8 +92,7 @@ pub async fn create_clients(
     values: Vec<Vec<u8>>,
     keys: Vec<String>,
     validator_name: &Option<ValidatorName>,
-) -> (IntoMakeService<Router>, KvManager) {
-    let listener_state = ListenerState::default();
+) -> (IntoMakeService<Router>, KvManager, SubxtAccountId32) {
```
HCastano, line 95: Do we need to return the account ID here? Or maybe said another way, is there no way to access the app state in the test context after this setup helper?

CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ runtime
 - In [#1147](https://github.com/entropyxyz/entropy-core/pull/1147) a field is added to the
   chainspec: `jump_started_signers` which allows the chain to be started in a pre-jumpstarted state
   for testing. If this is not desired it should be set to `None`.
+- In [#1216](https://github.com/entropyxyz/entropy-core/pull/1216) the `--setup-only` option for `entropy-tss`
```
HCastano, line 35: The second part of this breaking change kinda understates the impact this will have on the operator flow...we can maybe elaborate on this more when cutting the next release

HCastano (approved):
Nice work with this 💪 

cc @entropyxyz/system-reliability-engineers since this will impact your flow

crates/threshold-signature-server/src/helpers/launch.rs
ameba23: Ok i will add a limit but make an issue regarding the dev-ops-related workflow for how the host operator can be notified that the process has stopped.

crates/threshold-signature-server/src/user/api.rs
ameba23: Agree, this is a problem. The other way to approach this would be to start with a mini axum app which only exposes `/version` and `/info`, then when the checks have passed kill it and start our full axum app. But the disadvantage is that when an entropy-tss goes down and restarts, we would get a 404 when hitting eg: `/sign_tx`, which is hard to distinguish from some other problem.

crates/threshold-signature-server/src/helpers/tests.rs
ameba23: i hear you, but i am not sure how to access app state after we have called `.into_make_service()`

ameba23:
As this is a significant breaking change and i feel like there is still a bit of unsureness i am not gonna merge until we have talked about it at the core sync.

Having looked at the option of SGX-sealing the keyshare - i have concerns about how we can use a recovered keyshare if we have a new TSS account id - since the keyshare's party id is the TSS account id, and we need to sign protocol messages with it in order to do anything useful with the keyshare.

But that said i don't really see another option than this, because of needing the keypairs to prove attestation.

@JesseAbram @HCastano

crates/threshold-signature-server/src/helpers/launch.rs
ameba23: > Doesn't look like multiple balance queries like this are handled [natively by subxt](https://github.com/paritytech/subxt/issues/648)

Can you explain what you mean a bit more i dont get it.  It looks like the person who made that issue wants to check balances for a bunch of different accounts, but here we are making a balance query for one account several times one after the other.

PR #1222, opened 2024-12-17
Add oracle data to cli

JesseAbram:
Allows the Cli to post oracle data pointers to chain

pairs with https://github.com/entropyxyz/programs/pull/100


crates/test-cli/src/lib.rs
```diff
@@ -132,6 +133,8 @@ enum CliCommand {
         config_interface_file: Option<PathBuf>,
         /// The path to a file containing the program aux interface (defaults to empty)
         aux_data_interface_file: Option<PathBuf>,
+        /// The path to a file containing the program oracle data (defaults to empty)
```
ameba23, line 136:
```suggestion
        /// The path to a file containing the program oracle data headings (defaults to empty)
```
or similar - just to make clear that its the headings that this file contains because otherwise when i read i think of the actual data that the oracle retrieves

ameba23 (approved):
:star: all looks great.

At first i was thinking that since these headers are small and likely human readable they dont really need to be put in a file and could be given on the command line.  But actually it makes total sense since this file is generated automatically in the paired PR.

PR #1228, opened 2024-12-20
Report unstable peers from TSS

HCastano:
This is a follow up to #1215.

In this PR we enable TSS servers to report each other on-chain for "offline" or
connection related failures during signing (note that this PR doesn't include DKG or
proactive refresh) using the reporting extrinsic introduced in #1215.

When trying to figure out when to report I basically walked through the signing flow to
see where connections were being made (either inbound or outbound).

For outbound connections we basically report all the errors from
`open_protocol_connections()`. For inbound, we report peers that we're expecting to
connect to us during signing but never do.

As far as testing goes it was a bit tricky to hit all the different networking/connection
codepaths programmatically. In the end I did mange to get one unit test for inbound
connections and one for outbound connections (thanks for the help here @ameba23).

For those errors I couldn't test programmatically I did do some manual testing with hacked
Docker images (inserting crashes at connection points) - but I definitely could've missed
something here.

### DIFF
--- a/crates/threshold-signature-server/src/helpers/signing.rs
+++ b/crates/threshold-signature-server/src/helpers/signing.rs
@@ -91,12 +91,20 @@ pub async fn do_signing(
     )
     .await?;
 
-    // TODO (Nando): Maybe report here since we're expecting a connection from somebody and they
-    // never responded to us?
     let channels = {
-        let ready = timeout(Duration::from_secs(SETUP_TIMEOUT_SECONDS), rx_ready).await?;
-        let broadcast_out = ready??;
-        Channels(broadcast_out, rx_from_others)
+        match timeout(Duration::from_secs(SETUP_TIMEOUT_SECONDS), rx_ready).await {
+            Ok(ready) => {
+                let broadcast_out = ready??;
+                Channels(broadcast_out, rx_from_others)
+            },
+            Err(e) => {
+                let remaining_peers = app_state.protocol_peers(session_id);
+                return Err(ProtocolErr::Timeout {
+                    source: e,
+                    inactive_peers: Some(remaining_peers),
+                });
+            },
+        }
     };
 
     let result = signing_service
--- a/crates/threshold-signature-server/src/helpers/signing.rs
+++ b/crates/threshold-signature-server/src/helpers/signing.rs
@@ -98,10 +98,10 @@ pub async fn do_signing(
                 Channels(broadcast_out, rx_from_others)
             },
             Err(e) => {
-                let remaining_peers = app_state.unsubscribed_peers(&session_id)?;
+                let unsubscribed_peers = app_state.unsubscribed_peers(&session_id)?;
                 return Err(ProtocolErr::Timeout {
                     source: e,
-                    inactive_peers: Some(remaining_peers),
+                    inactive_peers: Some(unsubscribed_peers),
                 });
             },
         }
--- a/crates/threshold-signature-server/src/helpers/signing.rs
+++ b/crates/threshold-signature-server/src/helpers/signing.rs
@@ -82,14 +82,25 @@ pub async fn do_signing(
         .map_err(|_| ProtocolErr::SessionError("Error getting lock".to_string()))?
         .insert(session_id.clone(), listener);
 
-    open_protocol_connections(
+    let result = open_protocol_connections(
         &sign_context.sign_init.validators_info,
         &session_id,
         signer,
         state,
         &x25519_secret_key,
     )
-    .await?;
+    .await;
+
+    match result {
+        Ok(_) => (),
+        Err(ProtocolErr::ConnectionError { source: _, account_id: _ }) => {
+            todo!()
+        },
+        // Err(ProtocolErr::EncryptedConnection(_)) => todo!(),
+        // Err(ProtocolErr::BadSubscribeMessage(_)) => todo!(),
+        // Err(ProtocolErr::Subscribe(_)) => todo!(),
+        _ => todo!(),
+    }
 
     let channels = {
         let ready = timeout(Duration::from_secs(SETUP_TIMEOUT_SECONDS), rx_ready).await?;
--- a/crates/threshold-signature-server/src/helpers/signing.rs
+++ b/crates/threshold-signature-server/src/helpers/signing.rs
@@ -91,6 +91,8 @@ pub async fn do_signing(
     )
     .await?;
 
+    // TODO (Nando): Maybe report here since we're expecting a connection from somebody and they
+    // never responded to us?
     let channels = {
         let ready = timeout(Duration::from_secs(SETUP_TIMEOUT_SECONDS), rx_ready).await?;
         let broadcast_out = ready??;
--- a/crates/threshold-signature-server/src/helpers/signing.rs
+++ b/crates/threshold-signature-server/src/helpers/signing.rs
@@ -98,7 +98,7 @@ pub async fn do_signing(
                 Channels(broadcast_out, rx_from_others)
             },
             Err(e) => {
-                let remaining_peers = app_state.protocol_peers(session_id);
+                let remaining_peers = app_state.unsubscribed_peers(&session_id)?;
                 return Err(ProtocolErr::Timeout {
                     source: e,
                     inactive_peers: Some(remaining_peers),
--- a/crates/threshold-signature-server/src/helpers/signing.rs
+++ b/crates/threshold-signature-server/src/helpers/signing.rs
@@ -99,10 +99,7 @@ pub async fn do_signing(
             },
             Err(e) => {
                 let unsubscribed_peers = app_state.unsubscribed_peers(&session_id)?;
-                return Err(ProtocolErr::Timeout {
-                    source: e,
-                    inactive_peers: unsubscribed_peers,
-                });
+                return Err(ProtocolErr::Timeout { source: e, inactive_peers: unsubscribed_peers });
             },
         }
     };
--- a/crates/threshold-signature-server/src/helpers/signing.rs
+++ b/crates/threshold-signature-server/src/helpers/signing.rs
@@ -93,13 +93,26 @@ pub async fn do_signing(
 
     match result {
         Ok(_) => (),
-        Err(ProtocolErr::ConnectionError { source: _, account_id: _ }) => {
-            todo!()
+        Err(e)
+            if matches!(
+                e,
+                ProtocolErr::ConnectionError { .. }
+                    | ProtocolErr::EncryptedConnection { .. }
+                    | ProtocolErr::BadSubscribeMessage { .. }
+                    | ProtocolErr::Subscribe { .. }
+            ) =>
+        {
+            let _account_id = match e {
+                ProtocolErr::ConnectionError { ref account_id, .. } => account_id,
+                ProtocolErr::EncryptedConnection { ref account_id, .. } => account_id,
+                ProtocolErr::BadSubscribeMessage { ref account_id, .. } => account_id,
+                ProtocolErr::Subscribe { ref account_id, .. } => account_id,
+                _ => unreachable!(),
+            }.clone();
+
+            return Err(e);
         },
-        // Err(ProtocolErr::EncryptedConnection(_)) => todo!(),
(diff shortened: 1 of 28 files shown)


crates/threshold-signature-server/src/signing_client/api.rs
```diff
@@ -139,6 +139,8 @@ async fn handle_socket_result(socket: WebSocket, app_state: AppState) {
     if let Err(err) = handle_socket(socket, app_state).await {
         tracing::warn!("Websocket connection closed unexpectedly {:?}", err);
         // TODO here we should inform the chain that signing failed
+        //
+        // TODO (Nando): Report error up here?
```
HCastano, line 143: I think the answer here is _no_ since this would be bubbled up when we try and make a (failed) `ws/` connection anyways

ameba23: I haven't totally thought through but seems reasonable.  The only errors we could catch here but not elsewhere are relating to incoming connections which do not relate to an existing listener.  Eg: connections from random people or with a bad 'subscribe' message. I think we can just ignore these.

ameba23:
As for (2):
- A TSS server just doesn't show up.  This should be easy to test.
- A TSS connects and submits a valid subscribe message, but then drops the connection.  We can write a test helper that does this - i think there already is something similar for testing bad subscribe messages.
- A TSS server drops the connection in the middle of a protocol run.  This is hard to test. The only thing i can think of is to wrap `setup_client` https://github.com/entropyxyz/entropy-core/blob/a5c69b2a1d6e1fe571ba7194899343db50f834e4/crates/threshold-signature-server/src/helpers/tests.rs#L74 in a [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) which kills the TSS server after some amount of time which is likely to be in the middle of the protocol.  Since the signing protocol is fast, this might be too unreliable.  But it would probably work for the DKG/reshare protocols. Kinda ugly though.

ameba23:
Probably the most common problem will be that a TSS server is just not available for whatever reason.

I think its worth noting that the relayer node will immediately know this, because they will be unable to make the `sign_tx` request.  But currently they do nothing about it, other than logging an error: https://github.com/entropyxyz/entropy-core/blob/a5c69b2a1d6e1fe571ba7194899343db50f834e4/crates/threshold-signature-server/src/user/api.rs#L206

Really here they should tell the other TSS servers that they can abandon the signing protocol. There is no point in waiting for this one to connect, as they have failed to receive the signature request.

Otherwise, if the unresponsive node is due to make an outgoing ws connection, its hard to say how long we should wait for them before reporting unresponsiveness. Are they offline or just being slow?

HCastano:
@ameba23 thanks for the feedback!

> A TSS server just doesn't show up. This should be easy to test.

This would just end up with a failure at the point of the relayer, not necessarily the
signers (which we don't deal with right now).

This ties into your second comment though, we can have a follow up in which we just deal
with the relayer <-> signer failure case.

> A TSS connects and submits a valid subscribe message, but then drops the connection. We
> can write a test helper that does this - i think there already is something similar for
> testing bad subscribe messages.

Okay yeah, I can try and look into this.

> A TSS server drops the connection in the middle of a protocol run. This is hard to
> test. The only thing i can think of is to wrap setup_client

Hmm this doesn't seem ideal, and if we want to run it in tests we may end up with
nondeterministic tests...I'll poke around the code here and see if I can think of
anything else.

HCastano:
Peg, would you be able to point me to the bad subscribe test you were talking about?

I was playing around with throwing some errors into the [protocol test helpers](https://github.com/entropyxyz/entropy-core/blob/master/crates/protocol/tests/helpers/mod.rs) but looks like they're too low level for what we wan to trigger in this case.

ameba23:
> Peg, would you be able to point me to the bad subscribe test you were talking about?

Here we start the signing process and then send a subscribe message from someone not in the signing group:

https://github.com/entropyxyz/entropy-core/blob/58069f5a2d28ef4b59714a1d063c479493cbd305/crates/threshold-signature-server/src/user/tests.rs#L726

So i guess we would change it so that it is someone who is in the signing group, who sends a subscribe message and then does nothing.  The problem is, we need to stop that node from connecting as normal at the same time. So i guess we would need `spawn_testing_validators` to somehow omit one node.  Or maybe theres some other way to go about it.

crates/threshold-signature-server/src/helpers/signing.rs
```diff
@@ -91,6 +91,8 @@ pub async fn do_signing(
     )
     .await?;
 
+    // TODO (Nando): Maybe report here since we're expecting a connection from somebody and they
```
HCastano, line 94: @ameba23 do you know if there's a way to hold other parties accountable here?

My current problem: I know in `open_protocol_connections()` that there are some parties that will initiate the protocol with us. However, if they don't then we trigger this timeout and bail - I'd like to be able to know who they were so I can report them.

I can figure out who we are expecting to connect with us using the implementation details in `open_protocol_connections()` (bit of a leak imo), but from that I can't really assign "blame" to the party who didn't end up connecting with us (I only who know I expected to connect, but not who actually failed to connect).

Another question here. Say we're expecting two peers to connect with us, and only one of them does. I think because there's _some_ data that went through the channel this check would pass, right? If that's true then it might also be tricky for me to report the second peer from here

ameba23: > I only who know I expected to connect, but not who actually failed to connect

Not sure if this answers your question, but we can see which nodes did not yet connect (or which we did not yet connect to) for a given session id using `AppState.listener_state` which is a hashmap of session IDs to [`Listener`s](https://github.com/entropyxyz/entropy-core/blob/df7f6cdafd80f528c3235a910653bf8ea1234fb1/crates/protocol/src/listener.rs#L34).  [`Listener.validators`](https://github.com/entropyxyz/entropy-core/blob/df7f6cdafd80f528c3235a910653bf8ea1234fb1/crates/protocol/src/listener.rs#L43) tracks the remaining TSS nodes we have not yet connected to or received a connection from.

Whenever we connect to or receive a connection from one of these we call `listener.subscribe()` which internally removes that TSS node from the remaining `validators`.

So if we want to report all peers who failed to make an incoming connection when the timeout was reached, we could maybe do something like this with lines 97-99 below:

```rust
match timeout(Duration::from_secs(SETUP_TIMEOUT_SECONDS), rx_ready).await {
  Ok(ready) => {
    let broadcast_out = ready??;
    Channels(broadcast_out, rx_from_others)
  }
  Err(_) => {
    let listeners = app_state.listeners.lock()?;
    let listener = listeners.get(&session_id).ok_or(ProcotolErr::NoSessionId)?;
    let remaining_validators: Vec<_> = listener.validators.keys().map(|id| SubxtAccountId32(id)).collect();
    return Error(ProtocolErr::TimedOutWaitingForConnectionsFrom(remaining_validators));
  }
}
```

This code wont work as it is, but you get the idea.

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -370,6 +372,56 @@ pub async fn sign_tx(
...
+        ProtocolErr::Timeout { inactive_peers, .. } => inactive_peers.clone(),
+        _ => vec![],
+    };
+
+    // This is a non-reportable error, so we don't do any further processing with the error
+    if peers_to_report.is_empty() {
+        return Err(error.to_string());
+    }
+
+    tracing::debug!("Reporting `{:?}` for `{}`", peers_to_report.clone(), error.to_string());
```
HCastano, line 400: In the logs these (and basically all the other `AccountId`s from this PR) are formatted as byte arrays. I can make them SS58 addresses, but this would potentially give us inconsistent logs with other events which use byte arrays. While they're harder to read I'd rather have more consistency across our logs - but lmk what you think

crates/threshold-signature-server/src/helpers/signing.rs
HCastano: Super helpful, thanks 🙏

crates/threshold-signature-server/src/user/api.rs
ameba23: +1 for being consistent everywhere.  if i remember right subxt doesn't have the ss58 codec but sp-core does. At one point this was an issue as in some places we didn't have sp-core, but im pretty sure we sorted that and all our crates which would need to log something have it now.

ameba23 (approved):
:100: this is a big step towards getting slashing finished.

As i understand the scope of this PR is only to report errors from the signing protocol, but it looks like the groundwork has been done to add this to DKG and resharing, and basically we would just need to add a call to `handle_protocol_errors`.

As for the hard to test cases i'm happy to leave it for now and if we don't come up with a way of doing automated tests we can maybe have instructions for doing a manual test in the pre-release checklist.
 
I haven't looked in much detail at how these errors get handled on the chain side. We should bear in mind that many of them could also be our own fault (eg: failed incoming connection because we lost connectivity ourself, and encryption errors because of corrupt data our end).  But i guess this is accounted for in that the slashing pallet gets a sufficient number of reports from different peers before taking action.

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -370,6 +372,56 @@ pub async fn sign_tx(
...
+        | ProtocolErr::BadSubscribeMessage { account_id, .. }
+        | ProtocolErr::Subscribe { account_id, .. } => vec![account_id.clone()],
+
+        ProtocolErr::Timeout { inactive_peers, .. } => inactive_peers.clone(),
+        _ => vec![],
+    };
+
+    // This is a non-reportable error, so we don't do any further processing with the error
+    if peers_to_report.is_empty() {
+        return Err(error.to_string());
```
JesseAbram, line 397: as for the response here shouldn't this be Ok() as in the function here did not error out, everything progressed normally within the context of this function

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -370,6 +372,56 @@ pub async fn sign_tx(
...
+
+        if let Err(tx_error) =
+            submit_transaction(api, rpc, signer, &report_unstable_peer_tx, None).await
+        {
+            failed_reports.push(format!("{}", tx_error));
+        }
+    }
+
+    if failed_reports.is_empty() {
+        Err(error.to_string())
```
JesseAbram, line 415: same here

crates/threshold-signature-server/src/signing_client/api.rs
```diff
@@ -293,7 +292,19 @@ pub async fn get_channels(
...
-    let ready = timeout(Duration::from_secs(SETUP_TIMEOUT_SECONDS), rx_ready).await?;
-    let broadcast_out = ready??;
-    Ok(Channels(broadcast_out, rx_from_others))
+    match timeout(Duration::from_secs(SETUP_TIMEOUT_SECONDS), rx_ready).await {
+        Ok(ready) => {
+            let broadcast_out = ready??;
+            Ok(Channels(broadcast_out, rx_from_others))
+        },
+        Err(e) => {
+            let unsubscribed_peers = state.unsubscribed_peers(session_id).map_err(|_| {
```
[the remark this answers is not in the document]
HCastano: There are two `unsubscribed_peers()` methods - the one on `ListenerState` (which is what's used) here returns a `SubscribeErr` so we have to map it.

The one on `AppState` returns a `ProtocolErr` but we don't have access to that here

crates/threshold-signature-server/src/user/api.rs
HCastano: I do agree with you - but if we change this to an `Ok<String>` it'll mean that the caller will end up needing to treat the `Ok` case as a failure/error, because that's what it is. By only returning errors were we can save ourselves this and just do an `unwrap_err()` at the call site.

We also don't do any extra processing with the `Err` case - we log it and move on. If we had some different behaviour at the call site, then yes could make sense.

crates/threshold-signature-server/src/user/api.rs
HCastano: Opened https://github.com/entropyxyz/entropy-core/issues/1271. Could be a Friday activity for someone lol

PR #1237, opened 2025-01-06
Remove manual password input

### LINKED ISSUE #1220: Remove manual password input for kvdb
Its still not totally clear whether we are keeping or ditching the kvdb.  But what i think definately needs to change before we deploy entropy-tss to a confidential virtual machine is remove the manual password input.  There will be no way for the validator operator to enter a password via a terminal, as there will be no shell access to the virtual machine.

The quickest was to do this would be to change this line:
https://github.com/entropyxyz/entropy-core/blob/a1e36983135cd6333197101b4285bb715e4d7e5e/crates/threshold-signature-server/src/helpers/launch.rs#L146

To use `PasswordMethod::NoPassword`.

JesseAbram:
Closes #1220 

Adds a default password path at .password.txt for ease in testing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ test_db_3002
 
 # Keyshare files from the test CLI
 crates/test-cli/keyshare-*
+.password.*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ runtime
   extrinsics no longer take a PCK certificate chain. Rather, the certificate chain is extracted from the
   provided quote. The test CLI `change-threshold-accounts` command also no longer takes a PCK
   certificate chain.
+- In [#1237](https://github.com/entropyxyz/entropy-core/pull/1237) manual password input is removed and a
+ password file is now necessary (unless running alice, bob etc).
 
 ### Added
 - In [#1128](https://github.com/entropyxyz/entropy-core/pull/1128) an `/info` route was added to `entropy-tss`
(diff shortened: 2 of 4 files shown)


HCastano (commented):
I would approach this a bit differently.

The behaviour here should be that by default we use `PasswordMethod::NoPassword`. We may
need some dummy/default password for encrypting the DB for now (but medium term this
shouldn't be needed).

If a `--password-file` flag is used, use it to encrypt and load the DB. However, don't
have a default password file created or lying around.

I'm kinda approaching this from this mindset where operators shouldn't have to deal with
the encryption in any way. Instead the KVDB should just get encrypted by the fact that it
lives in TEE memory.

JesseAbram:
I mean that is fair, but are we on the fence still about removing the kvdb cuz if so then #1216 should wait, if we are then the whole kvdb should get removed and I can add onto 1216 and then this would no longer be a thing

ameba23 (approved):
Im approving this, just because i think its better than what we currently have as there will be no way to manually enter a password.

But its going to be tricky for operators to add this password file when running on TDX as there will be no ssh access to the virtual machine in which entropy-tss runs.  It is possible to make filesystem changes but its not simple and i'm not sure we should encourage people to do it.

> I mean that is fair, but are we on the fence still about removing the kvdb cuz if so then https://github.com/entropyxyz/entropy-core/pull/1216 should wait, if we are then the whole kvdb should get removed and I can add onto 1216 and then this would no longer be a thing

I think we probably are going to end up ditching the kvdb. As it looks to me right now our options for storing the keyshare are either SGX sealing, or in-memory only.

JesseAbram:
Closed due to talks on #1247

PR #1242, opened 2025-01-08
Add cli options for adding validator

JesseAbram:


crates/client/src/client.rs
```diff
@@ -525,3 +536,145 @@ pub async fn get_tdx_quote(
...
+pub fn deconstruct_session_keys_string(session_keys: String) -> Result<SessionKeys, String> {
+    if session_keys.len() != 256 {
+        return Err(String::from("Session keys len cannot have length be more or less than 256"));
+    }
+
+    let err = || String::from("Internal Math Error");
+    let len = session_keys.len();
+    let mut session_keys_u8: Vec<u8> = Vec::with_capacity(128);
+    let mut iter = session_keys.chars();
+    for _ in (0..len).step_by(2) {
```
ameba23, line 665: Im a bit confused - isn't this converting a hex string to a `Vec<u8>`?  Why not use the `hex` crate?

crates/client/src/client.rs
```diff
@@ -525,3 +536,145 @@ pub async fn get_tdx_quote(
...
+    })
+}
+
+/// Deconstructs a string session key into SessionKeys type
+pub fn deconstruct_session_keys_string(session_keys: String) -> Result<SessionKeys, String> {
+    if session_keys.len() != 256 {
+        return Err(String::from("Session keys len cannot have length be more or less than 256"));
+    }
+
+    let err = || String::from("Internal Math Error");
```
ameba23, line 661:
```suggestion
    let err = || String::from("Session keys must be given as hex");
```

crates/client/src/client.rs
```diff
@@ -525,3 +536,145 @@ pub async fn get_tdx_quote(
...
+    Ok(result_event)
+}
+
+/// Deconstructs a session key into SessionKeys type
+pub fn deconstruct_session_keys(session_keys: Vec<u8>) -> Result<SessionKeys, String> {
+    if session_keys.len() != 128 {
+        return Err(String::from("Session keys len cannot have length be more or less than 128"));
+    }
+
+    let babe: [u8; 32] = session_keys[0..32].try_into().unwrap();
```
ameba23, line 642: Should probably return an error rather than unwrap.

Also, is this the format we get the session keys from the node in?  I would have expected json or something.

crates/client/src/tests.rs
```diff
@@ -282,3 +284,127 @@ async fn test_get_oracle_headings() {
...
+
+#[tokio::test]
+#[serial]
+async fn test_set_session_key_and_declare_validate() {
+    let one = AccountKeyring::Ferdie;
+    let substrate_context = test_context_stationary().await;
+
+    let api = get_api(&substrate_context.node_proc.ws_url).await.unwrap();
+    let rpc = get_rpc(&substrate_context.node_proc.ws_url).await.unwrap();
+    let session_key = "0x833b9e467823236c10348fd26f114ecf10f46e1be41b6a2ba4fab0b915a0b0affc9d3031694d1c65b711b8c5f79d57f734c9fa01366164ecc3ff8d84a04759160c2cf916dcabb8135c7605106052f9baf8e45a4f348cef2785f83a9674bec964ac3013772500af6236b1f6491a7024cacf69557b4b89779afdbc017de364e166".to_string();
```
ameba23, line 316: Be interested to know how you got this value

ameba23 (approved):
:1st_place_medal: Great work for getting this over the line.

It would be great to try this out - i'm guessing it will be tricky to add a node to our existing docker-compose setup because of network-redirecting stuff, but maybe not.  Otherwise i guess we have to wait until we have a deployment with this.

crates/client/src/client.rs
JesseAbram: nope just a hex string lol, and ya will handle the error better

JesseAbram:
> 🥇 Great work for getting this over the line.
> 
> It would be great to try this out - i'm guessing it will be tricky to add a node to our existing docker-compose setup because of network-redirecting stuff, but maybe not. Otherwise i guess we have to wait until we have a deployment with this.

ya zach is working on it, I am going to back fill this code to version 0.3.0 (not merge it just so he can try it out on testnet and without the tdx baggage)

PR #1250, opened 2025-01-14
Allow different versions for programs

JesseAbram:
Add a way for programs to have different versions

Can be made more efficient by caching runtime version and using cached instead of spinning up a new one (#1251)

### DIFF
--- a/crates/threshold-signature-server/src/helpers/user.rs
+++ b/crates/threshold-signature-server/src/helpers/user.rs
@@ -161,6 +161,7 @@ pub async fn compute_hash(
     }
 }
 
+/// Gets and launches a runtime for a program version
 pub fn get_programs_runtime(version: u8, fuel: u64) -> Result<Runtime, UserErr> {
     match version {
         0 => Ok(Runtime::new(ProgramConfig { fuel })),
--- a/crates/threshold-signature-server/src/helpers/user.rs
+++ b/crates/threshold-signature-server/src/helpers/user.rs
@@ -16,7 +16,7 @@
 //! Utilities relating to the user
 use std::time::Duration;
 
-use entropy_programs_runtime::Runtime;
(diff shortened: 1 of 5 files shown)


crates/threshold-signature-server/src/helpers/user.rs
```diff
@@ -154,8 +154,17 @@ pub async fn compute_hash(
...
             let program_info = get_program(api, rpc, &programs_data[*i].program_pointer).await?;
+            let mut runtime = get_programs_runtime(program_info.version_number, fuel)?;
             runtime.custom_hash(program_info.bytecode.as_slice(), message).map_err(|e| e.into())
         },
         _ => Err(UserErr::UnknownHashingAlgorithm),
     }
 }
+
+/// Gets and launches a runtime for a program version
+pub fn get_programs_runtime(version: u8, fuel: u64) -> Result<Runtime, UserErr> {
```
ameba23, line 165: The return type of this function would need to change if we added another supported runtime version, no?  Or are you thinking to have the version number be passed into a single `Runtime` struct which manages versions internally?  If so, i think we don't need this function as we can push this logic over to the programs repo, and if not, the return type should probably be a trait object or an enum.

ameba23 (approved):
Approving as there are no red flags, but i don't totally understand how having multiple supported runtime versions is going to look in practice.

crates/threshold-signature-server/src/helpers/user.rs
JesseAbram: ya good point, I broke this up into evaluation, so the return type should be consistent, the enum would just create match statments in the code line which I was trying to avoid, I think the new way is nice and a light touch

ameba23 (approved):
:100: This is much better, nice one.

PR #1254, opened 2025-01-16
Reshare e2e test

JesseAbram:


ameba23 (approved):
:star2: This is great as i am really depressed about the fact that our CI is still not reliably passing after all the things we have tried.

I could only get this to work locally by manually adding the line:
```rust
reshare-test=["entropy-runtime/reshare-test"]
```
to `./node/cli/Cargo.toml`.  So please either add that or document how to run the test without needing to add that.

Also, i am not so sure about the second part of the test where we wait 80 seconds for another reshare.  I ran the test 3 times locally, and twice is passed but once it failed with:

```
---- validator::tests::test_reshare_e2e stdout ----
thread 'validator::tests::test_reshare_e2e' panicked at crates/threshold-signature-server/src/validator/tests.rs:297:5:
assertion `left != right` failed
```

Yes, probably this is because i have slow hardware.  But i think ideally the second reshare test should use the same code as the first bit, where we first wait to see if signers have changed, then wait another 10s for the rotation.

JesseAbram:
> 🌟 This is great as i am really depressed about the fact that our CI is still not reliably passing after all the things we have tried.
> 
> I could only get this to work locally by manually adding the line:
> 
> ```rust
> reshare-test=["entropy-runtime/reshare-test"]
> ```
> 
> to `./node/cli/Cargo.toml`. So please either add that or document how to run the test without needing to add that.
> 
hmmmm interesting worked locally and on the CI, but will add

RELEASE_CHECKLIST.md
```diff
@@ -5,6 +5,9 @@ tagged as the final release.
 
 ## Pre-Prep
 - [ ] Inform relevant parties that you're preparing a release (e.g, by posting on Discord)
+- [ ] Manually run e2e reshare test 
+(first compile node with features=reshare-test and then run
```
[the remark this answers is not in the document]
JesseAbram: gonna add based on the other comment, but weird I don't see it locally or on CI

RELEASE_CHECKLIST.md
```diff
@@ -5,6 +5,9 @@ tagged as the final release.
 
 ## Pre-Prep
 - [ ] Inform relevant parties that you're preparing a release (e.g, by posting on Discord)
+- [ ] Manually run e2e reshare test
```
HCastano, line 8: I would try and avoid running this manually and instead make it a part of the release CI job. We have an `on > push > release/...` stage where we can add an extra testing stage which encompasses this.

For your own

RELEASE_CHECKLIST.md
```diff
@@ -5,6 +5,9 @@ tagged as the final release.
 
 ## Pre-Prep
 - [ ] Inform relevant parties that you're preparing a release (e.g, by posting on Discord)
+- [ ] Manually run e2e reshare test 
+(first compile node with features=reshare-test and then run 
+```cargo test --release --features=reshare-test -- --test test_reshare_e2e```)
```
HCastano, line 10:
```suggestion
`cargo test --release --features=reshare-test -- --test test_reshare_e2e`)
```

Single backticks for inline formatted code is alright

runtime/src/constants.rs
```diff
@@ -78,7 +78,10 @@ pub mod time {
 
     // NOTE: Currently it is not possible to change the epoch duration after the chain has started.
     //       Attempting to do so will brick block production.
+    #[cfg(not(feature = "reshare-test"))]
     pub const EPOCH_DURATION_IN_BLOCKS: BlockNumber = 4 * HOURS;
+    #[cfg(feature = "reshare-test")]
+    pub const EPOCH_DURATION_IN_BLOCKS: BlockNumber = 15 / (SECS_PER_BLOCK as BlockNumber);
```
HCastano, line 84: Why define this in terms of `SECS_PER_BLOCK` instead of having something fixed for tests? Say, `1 * MINUTE` or something like that?

crates/threshold-signature-server/Cargo.toml
```diff
@@ -113,4 +113,5 @@ alice       =[]
 bob         =[]
 # Enable this feature to run the integration tests for the wasm API of entropy-protocol
 # This requires the entropy-protocol node-js module to be built and so is not run by default
-wasm_test=[]
+wasm_test   =[]
+reshare-test=[]
```
HCastano, line 117: We should add some comments explaining the feature here and in the other TOML files

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -239,6 +239,64 @@ async fn test_reshare_basic() {
...
+        spawn_testing_validators(crate::helpers::tests::ChainSpecType::IntegrationJumpStarted)
+            .await;
+
+    let force_authoring = true;
+    let context =
+        test_node_process_testing_state(ChainSpecType::IntegrationJumpStarted, force_authoring)
+            .await;
+    let api = get_api(&context[0].ws_url).await.unwrap();
+    let rpc = get_rpc(&context[0].ws_url).await.unwrap();
+    let client = reqwest::Client::new();
```
HCastano, line 259: We can replace this with `spawn_tss_nodes_and_start_chain()`

HCastano (commented):
Main thing for me is to try and get this to run in CI but only on release (or test release) tags

runtime/src/constants.rs
JesseAbram: because I want it less than one minuet and I I get an overflow problem that way

RELEASE_CHECKLIST.md
JesseAbram: should be doable but let me do in another PR so I can work with zach of vi on that

JesseAbram: #1270

PR #1258, opened 2025-01-22
Add entropy-shared check script

ameba23:
This is a small PR which adds a script which runs `cargo check` on `entropy-shared` with all the different combinations of feature flags / targets which will be used. 

This is useful to check that changes to entropy-shared have not broken anything, without needing to run the entire CI pipeline.

I found this useful recently when getting into a big muddle after making some changes there, so i would like to keep this script in the scripts directory for next time.

### DIFF
--- a/scripts/check-entropy-shared.sh
+++ b/scripts/check-entropy-shared.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+# This script checks that entropy-shared will compile with all the different combinations of feature
+# flags / targets which will be used. This is useful to check that changes to entropy-shared have
+# not broken anything without needing to run the entire CI pipeline.
+#
+# Used by entropy-tss
+cargo check -p entropy-shared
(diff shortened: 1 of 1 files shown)


HCastano (approved):
Looks fine, but two things:

1. I imagine there's a more Rust-y way of checking feature combinations, maybe there's an existing crate we can leverage?
2. Probably worthwhile to add this check into CI so that if something breaks we can know before merging code into `master`

ameba23:
>     2. Probably worthwhile to add this check into CI so that if something breaks we can know before merging code into `master`
The CI pipleline does make all these checks but in some cases only when it compiles stuff eg: when it compiles `entropy-protocol`  for wasm in order to run the nodejs bindings tests. So im not sure we gain anything by adding this to CI, The idea of this is just to be able to check stuff quickly without needing to wait for CI.

Issue #1247, opened 2025-01-13
Design for a recovery feature using other entropy-tss nodes as key-providers


ameba23:
We have been discussing the problem of not having encrypted persistent storage with which to store the keyshares, which means we are vulnerable to lose keyshares if the VM needs to restart. 

This is a proposal for a possible design of a recovery feature.

Each entropy-tss node acts as a key server, providing symmetric encryption keys for persistent storage for other entropy-tss nodes.

On `entropy-tss` launch:
 - Check if there is an existing encrypted key-value store:
   - If there is not an existing encrypted store:
     - Generate a new TSS account ID
     - Select a TSS node to be our keyserver, by querying the chain for all TSS nodes and selecting one deterministically based on TSS account id (we should be ok to do this without any funds since no extrinsics needed).
     - Request an encryption key, calling `/request-encryption-key` and giving our TSS account ID to be used as a lookup key under which to store this key for retrieval later, as well as a quote (the handler for this endpoint is explained below). 
     - Encrypt our key-value store using this key. Also, store the server info of the TSS node used as a key-server, as well as our TSS account ID, in plaintext.
   - If there is an existing encrypted store:
     - Stored with it will be the server info of the chosen keyserver, as well as the account ID. Use these to call `/request-encryption-key` to get the key for the key-value store.

In the `/request-encryption-key` HTTP route handler:
 - POST request body should contain:
   - TSS account ID
   - An ephemeral encryption public key which can be used to provide an encrypted response to this request only.
   - A TDX quote, which should contain the TSS account ID and ephemeral public key in the input data.
 - Verify the given quote, and respond with an error if it cannot be verified. 
 - If we already have an encryption key associated with this account ID, respond with it (encrypted with the given ephemeral encryption key).
 - Otherwise, generate one, store it in our key-value store, and respond with it (encrypted with the given ephemeral encryption key).

This relies on there being at least two entropy-tss nodes at genesis - otherwise there is a chicken-and-egg problem where the first node does not have another node to use as a key server.

JesseAbram:
Instead of requesting a key, would be an easier flow especially in the spin up to create your own key, store it in memory (in the tdx machine) and then send it to other validators to be stored in their kvdb at periodic intervals, which sounds less secure but because of the tdx it is more secure and for me at least easier to manage. This can also allow you to "reshare" this key as validators leave, and maybe at intervals like a pre update, can also maybe have the chain manage the state of who has what

JesseAbram:
[...]

PR #1259, opened 2025-01-22
Add endpoints to migrate an encrypted key-value store dump for upgrading `entropy-tss` 

ameba23:
This relates to, https://github.com/entropyxyz/entropy-core/issues/1247 and is described in this comment:  https://github.com/entropyxyz/entropy-core/issues/1247#issuecomment-2603935848

This provides HTTP endpoints which allow us to export the key-value store as a raw encrypted key-value pair dump, and import it again on a new version of entropy-tss.

The workflow will be something like:

- Do GET `/backup_encrypted_db` to get a payload containing the encrypted db, as well as details of the 'key provider'.
- Stop the VM.
- Start a new VM with the new version of `entropy-tss`
- Do POST `/backup_encrypted_db` giving the payload from step 1 in the request body, which would put the db / key provider files back.  This route can only be used when the node is in a 'not yet ready' state.

Currently based off https://github.com/entropyxyz/entropy-core/pull/1249

Following core sync call - it seems like we maybe don't want or need this feature.

### DIFF
--- a/crates/client/src/client.rs
+++ b/crates/client/src/client.rs
@@ -43,7 +43,7 @@ use crate::{
     user::{
         self, get_all_signers_from_chain, get_validators_not_signer_for_relay, UserSignatureRequest,
     },
-    Hasher,
+    Hasher, TssPublicKeys,
 };
 pub use crate::{
     chain_api::{get_api, get_rpc},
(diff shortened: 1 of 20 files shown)


ameba23:
@JesseAbram @HCastano as it looks like we maybe don't want this, i am thinking to close it for now and possibly re-open later if we do a version upgrade runthrough and find that it is a problem that all TSS nodes change their IDs.

PR #1263, opened 2025-01-27
Add an extra TSS state for connected to chain but not funded / fully ready

ameba23:
This is a sub-pr of https://github.com/entropyxyz/entropy-core/pull/1249

### DIFF
--- a/crates/threshold-signature-server/src/lib.rs
+++ b/crates/threshold-signature-server/src/lib.rs
@@ -181,7 +181,7 @@ use entropy_shared::X25519PublicKey;
 use sp_core::{crypto::AccountId32, sr25519, Pair};
 use std::{
     collections::HashMap,
-    sync::{Arc, RwLock},
+    sync::{Arc, PoisonError, RwLock},
 };
 use subxt::{
     backend::legacy::LegacyRpcMethods, tx::PairSigner, utils::AccountId32 as SubxtAccountId32,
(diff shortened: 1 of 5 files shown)


crates/threshold-signature-server/src/lib.rs
```diff
@@ -256,17 +277,41 @@ impl AppState {
...
+        Ok(())
+    }
+
     /// Mark the node as ready. This is called once when the prerequisite checks have passed.
-    pub fn make_ready(&self) {
-        let mut is_ready = self.ready.write().unwrap();
-        *is_ready = true;
+    pub fn make_ready(&self) -> Result<(), PoisonError<std::sync::RwLockWriteGuard<'_, TssState>>> {
+        let mut tss_state = self.tss_state.write()?;
+        *tss_state = TssState::Ready;
```
HCastano, line 313: Do we want to be able to go from both of the other states to `Ready`, or enforce a specific order here?

HCastano (approved):

crates/threshold-signature-server/src/lib.rs
ameba23: In production, this will always be called after `connected_to_chain_node()` has been called, so we could enforce that that is the case.

But in our test helper functions we use this to jump straight to the ready state and don't bother checking for a chain connection.  So if we wanted to enforce an order we would also need to call `connected_to_chain_node()` in those helpers.

Im not sure it really matters either way.

PR #1266, opened 2025-01-28
Add no hash option

JesseAbram:
Adds a no hash option. After talking with @rh0delta a possibility to send a hash to be signed is a good use case, can do this hackily with custom hash, but this should be an out of the box option

### DIFF
--- a/crates/shared/src/types.rs
+++ b/crates/shared/src/types.rs
@@ -105,6 +105,7 @@ pub enum HashingAlgorithm {
     Sha3,
     Keccak,
     Blake2_256,
+    NoHash,
     Custom(usize),
 }
 
--- a/crates/shared/src/types.rs
+++ b/crates/shared/src/types.rs
@@ -105,6 +105,7 @@ pub enum HashingAlgorithm {
     Sha3,
     Keccak,
     Blake2_256,
+    /// If you are trying to sign a hashed message you can use this, length requirement is 64 digits
     NoHash,
     Custom(usize),
 }
--- a/crates/shared/src/types.rs
+++ b/crates/shared/src/types.rs
@@ -105,8 +105,8 @@ pub enum HashingAlgorithm {
     Sha3,
     Keccak,
     Blake2_256,
-    /// If you are trying to sign a hashed message you can use this, length requirement is 64 digits
-    NoHash,
+    /// An algorithm which produces the same output as its input.
+    Identity,
     Custom(usize),
 }
 
--- a/crates/threshold-signature-server/src/user/tests.rs
+++ b/crates/threshold-signature-server/src/user/tests.rs
@@ -957,14 +957,15 @@ async fn test_compute_hash() {
     let expected_hash = blake3::hash(PREIMAGE_SHOULD_SUCCEED).as_bytes().to_vec();
     assert_eq!(message_hash.to_vec(), expected_hash);
 
-    let no_hash = vec![0u8; 32];
-    let no_hash_too_long = vec![0u8; 33];
+    const EXPECTED_MAX_HASH_LENGTH: usize = 32;
+    let no_hash = vec![0u8; EXPECTED_MAX_HASH_LENGTH];
+    let no_hash_too_long = vec![0u8; EXPECTED_MAX_HASH_LENGTH + 1];
 
     // no hash
     let message_hash_no_hash = compute_hash(
         &api,
         &rpc,
-        &HashingAlgorithm::NoHash,
+        &HashingAlgorithm::Identity,
         10000000u64,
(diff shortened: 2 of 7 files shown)


crates/shared/src/types.rs
```diff
@@ -105,6 +105,8 @@ pub enum HashingAlgorithm {
     Sha3,
     Keccak,
     Blake2_256,
+    /// If you are trying to sign a hashed message you can use this, length requirement is 64 digits
+    NoHash,
```
HCastano, line 109:
```suggestion
    Identity,
```

This is also the terminology used by Substrate for a hashing function which produces the same output as its input

crates/shared/src/types.rs
```diff
@@ -105,6 +105,8 @@ pub enum HashingAlgorithm {
     Sha3,
     Keccak,
     Blake2_256,
+    /// If you are trying to sign a hashed message you can use this, length requirement is 64 digits
```
HCastano, line 108:
```suggestion
    /// An algorithm which produces the same output as its input.
```

Not sure we should be having such specific "advice" or use cases as part of the doc comments. All the hashers output 32 bytes (256-bits) here as well.

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -956,6 +956,36 @@ async fn test_compute_hash() {
     // custom hash program uses blake 3 to hash
     let expected_hash = blake3::hash(PREIMAGE_SHOULD_SUCCEED).as_bytes().to_vec();
     assert_eq!(message_hash.to_vec(), expected_hash);
+
+    let no_hash = vec![0u8; 32];
+    let no_hash_too_long = vec![0u8; 33];
```
HCastano, line 961: We should do something like this to make it clear that 32 bytes is the expected max size.

```rust
const EXPECTED_MAX_HASH_LENGTH = 32;
let no_hash = vec![0u8; EXPECTED_MAX_HASH_LENGTH];
let no_hash_too_long = vec![0u8; EXPECTED_HASH_LENGTH + 1];
```

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -956,6 +956,36 @@ async fn test_compute_hash() {
...
+        &api,
+        &rpc,
+        &HashingAlgorithm::NoHash,
+        10000000u64,
+        &vec![ProgramInstance { program_pointer: program_hash, program_config: vec![] }],
+        &no_hash_too_long,
+    )
+    .await;
+
+    assert!(message_hash_no_hash_too_long.is_err());
```
HCastano, line 988: Let's match on the actual error type here (I guess that conversion error?)

deny.toml
```diff
@@ -33,7 +33,6 @@ exceptions=[
   { allow=["AGPL-3.0"], name="entropy-client" },
   { allow=["AGPL-3.0"], name="entropy-testing-utils" },
   { allow=["AGPL-3.0"], name="entropy-test-cli" },
-  { allow=["AGPL-3.0"], name="entropy-create-test-keyshares" },
```
HCastano, line 36: nit: I'd open  a different PR to try and fix this stuff

HCastano (approved):
Makes sense to me to include this out of the box. Similar to the custom hash though, could be problematic if misused.

Few comments, main thing is that I'd change the variant name from `NoHash` to `Identity`.

crates/shared/src/types.rs
```diff
@@ -105,6 +105,8 @@ pub enum HashingAlgorithm {
...
+    NoHash,
     Custom(usize),
```
ameba23, line 110: Not related to this PR but just noticed this.  We should avoid using `usize` in shared types because on wasm usize is 32 bits and on native x64_64 its 64 bits.  Not sure it really matters with json, but with binary formats this can cause mayhem.  I can make an issue.

crates/shared/src/types.rs
JesseAbram: yes works but Im just gonna be pedantic tho, cuz doesn't that mean it is not a hashing algorithim if it has the same output lol

crates/shared/src/types.rs
JesseAbram: id be happy to fix it now but I feel like @HCastano would nit it

HCastano: @JesseAbram I'm just trying to get you more green squares 😆 but yeah, I'd do this in a follow up just so we can test it properly

ameba23: Are there any prizes for lots of green squares?

HCastano: @ameba23 You'll get a longer seadoo ride with @JesseAbram

PR #1269, opened 2025-01-29
Fix licensing issue

JesseAbram:


HCastano (approved):
Thanks for doing this in its own PR.

Instead of blanket allowing `Zlib` though I would add the two crates licensed with it (`const_format` and `const_format_proc_macros`) as entries into the `exceptions` table - as we've done with `ring` for example.

JesseAbram:
> Thanks for doing this in its own PR.
> 
> Instead of blanket allowing `Zlib` though I would add the two crates licensed with it (`const_format` and `const_format_proc_macros`) as entries into the `exceptions` table - as we've done with `ring` for example.

done I mean Zlib does seem to fall within what is acceptable but added as exception

PR #1272, opened 2025-01-30
HashingAlgorithm enum should use u32 not usize

ameba23:
See https://github.com/entropyxyz/entropy-core/pull/1266#discussion_r1933391577

### DIFF
--- a/crates/shared/src/types.rs
+++ b/crates/shared/src/types.rs
@@ -107,7 +107,7 @@ pub enum HashingAlgorithm {
     Blake2_256,
     /// An algorithm which produces the same output as its input.
     Identity,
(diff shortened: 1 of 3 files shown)


JesseAbram (approved):

HCastano (approved):
Wonder if we can have tests on different platforms to verify the size is as expected, something like `test_check_hash_pointer_out_of_bounds()`?

ameba23:
> Wonder if we can have tests on different platforms to verify the size is as expected, something like `test_check_hash_pointer_out_of_bounds()`?

Im fairly confident we wont have issues as u32 must always be <= usize. But if you think we need to add some cross platform tests we could.

PR #1275, opened 2025-01-31
Add request limit cache

JesseAbram:
Originally started off as writing a threaded test for multiple txs, found out in the ```increment_or_wipe_request_limit``` function there was a ```reserve_key``` kvdb call that did not handle the overloading tx well and caused failures. This was solved by adding a cache and moving the request limit there 

The cache probably should have been used before and will open up an issue about moving things that were stored in the kvdb that shouldn't to it.  #1280 

As well will open an issue to create a deletion strategy for request limit as it is O(n) (this problem existed with the kvdb too) #1281 

Also maybe we should merge other app_state hashmaps into cache #1282


ameba23 (approved):
The signing stress test is great :star:

And the cache is definitely an improvement, since this stuff does not need to be stored on disk. 

But i would have done it a bit differently - rather than a generalised cache i would have separate data structures for the different things we need to store.

So for the request limit i would have the property of `AppState` be:

`request_limit: Arc<RwLock<HashMap<VerifyingKey, RequestLimitStorage>>>`

and have similar properties for the other things we need to store (we could put them in a sub-struct if `AppState` is starting to get a mess).

The advantages:
- Less time spent waiting for lock to become free, as with everything behind a single RwLock the chances of it being locked are much higher. I think putting them all behind the same `Arc` is fine.
- Faster because you don't need to serialize and deserialize values.
- The type system guarantees that entries are well formed - impossible to have a failed lookup or corrupt value.

crates/threshold-signature-server/src/lib.rs
```diff
@@ -344,6 +352,46 @@ impl AppState {
         ))
     }
 
+    pub fn write_to_cache(&self, key: String, value: Vec<u8>) -> anyhow::Result<()> {
```
ameba23, line 355: nit: Im not super keen on using anyhow here as strictly speaking this is library code. Probably it doesn't really matter but in best case we make have a `CacheError`

crates/threshold-signature-server/src/lib.rs
```diff
@@ -255,13 +258,17 @@ pub struct AppState {
...
+pub type Cache = Arc<RwLock<HashMap<String, Vec<u8>>>>;
+
 impl AppState {
     /// Setup AppState with given secret keys
     pub fn new(
         configuration: Configuration,
         kv_store: KvManager,
         pair: sr25519::Pair,
         x25519_secret: StaticSecret,
+        cache: Cache,
```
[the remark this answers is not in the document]
JesseAbram: ya but if we want to test some pre state it would be easier......however I did create the unsafe calls for that

crates/threshold-signature-server/src/lib.rs
JesseAbram: Im ok with that, I saw your method I think I was thinking of a method with multiple errors then never ended up writing one

JesseAbram:
> The signing stress test is great ⭐
> 
> And the cache is definitely an improvement, since this stuff does not need to be stored on disk.
> 
> But i would have done it a bit differently - rather than a generalised cache i would have separate data structures for the different things we need to store.
> 
> So for the request limit i would have the property of `AppState` be:
> 
> `request_limit: Arc<RwLock<HashMap<VerifyingKey, RequestLimitStorage>>>`
> 
> and have similar properties for the other things we need to store (we could put them in a sub-struct if `AppState` is starting to get a mess).
> 
> The advantages:
> 
> * Less time spent waiting for lock to become free, as with everything behind a single RwLock the chances of it being locked are much higher. I think putting them all behind the same `Arc` is fine.
> * Faster because you don't need to serialize and deserialize values.
> * The type system guarantees that entries are well formed - impossible to have a failed lookup or corrupt value.

hmmmmm I see the benefit, the issue being it will be writing a lot more code instead of a generalized get call or write call it becomes multiple functions for each storage unit, im on the fence

crates/threshold-signature-server/src/lib.rs
```diff
@@ -255,13 +258,17 @@ pub struct AppState {
     pub attestation_nonces: Arc<RwLock<HashMap<X25519PublicKey, [u8; 32]>>>,
 }
 
+/// A global cache type for the TSS
+pub type Cache = Arc<RwLock<HashMap<String, Vec<u8>>>>;
```
HCastano, line 262: By using a `Vec<u8>` value we lose type guarantees here for no real benefit. We know exactly what type(s) we want to store, so imo we should just use them directly

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -602,7 +601,7 @@ pub async fn check_for_key(account: &str, kv: &KvManager) -> Result<bool, UserEr
 /// Checks the request limit
 pub async fn request_limit_check(
     rpc: &LegacyRpcMethods<EntropyConfig>,
-    kv_store: &KvManager,
+    app_state: &AppState,
```
HCastano, line 604: I'd try and limit the scope of what we pass into here. We don't need the whole of `AppState`, just the cache

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -429,6 +440,84 @@ async fn signature_request_with_derived_account_works() {
     clean_tests();
 }
 
+#[tokio::test]
+#[serial]
+async fn signature_request_overload() {
```
HCastano, line 445: What are the actual assertions needed for this test? I don't see anything (either success or failure)

HCastano (commented):
I'm with Peg here, we should be leveraging the type system here more + also defining the cache a bit better (but I guess that's being discussed in #1282).

crates/threshold-signature-server/src/lib.rs
```diff
@@ -344,6 +352,46 @@ impl AppState {
         ))
     }
 
+    pub fn write_to_cache(&self, key: String, value: Vec<u8>) -> anyhow::Result<()> {
+        self.clear_poisioned_cache();
```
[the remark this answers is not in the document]
JesseAbram: for one block, which would be better than a poisioned cache forever making it unable to sign

crates/threshold-signature-server/src/user/tests.rs
JesseAbram: that wwe get a verified signature https://github.com/entropyxyz/entropy-core/pull/1275/files#diff-e1d8c0027f2fda5b3998b919bb461f5935ad1f6690b719c637968150be6f5e76R502-R506

JesseAbram:
> @JesseAbram
> 
> @dvdplm suggested that we can probably get away with a non-poisoning lock here instead of explicitly ignoring the poisoned lock. [`parking_lot::RwLock`](https://docs.rs/parking_lot/latest/parking_lot/type.RwLock.html#differences-from-the-standard-library-rwlock) fits the description here, so maybe consider using that?

From the docs, "No poisoning, the lock is released normally on panic." which doesn't help the initial concern of the data not being updated.......which again is mostly being off by one which isn't a big deal, however locking the sign flow is, anyways can look in more on monday, make an issue

HCastano:
> From the docs, "No poisoning, the lock is released normally on panic." which doesn't help the initial concern of the data not being updated.......which again is mostly being off by one which isn't a big deal, however locking the sign flow is, anyways can look in more on monday, make an issue

One of the benefits here is that we lean into the fact that we're not offering any guarantees around a non-poisoned lock state - as opposed to hiding that from the caller by clearing the poisoned state internally

JesseAbram:
> > right but that means we either do this for all the rwlocks here or just this one? Im writting this one, I know that a poision lock should not lock the request limit Im not sure about the other rwlocks in app_state. would you rather be explicit about this or have two different rwlock libraries
> 
> Not sure - depends on the other data. Let's leave it as is for now, and make a concrete decision in #1282?

lol pretty much what I was thinking, also to be honest talking about this mutex problem is fun id like to bring it up in our next call.......just cuz idk it is fun

crates/threshold-signature-server/src/user/api.rs
JesseAbram: currently all hellper functions are in app state #1282 should handle that by creating a sub impl of cache

PR #1279, opened 2025-02-04
Ignore another failing integration test

HCastano:
This is a follow up to #1274 - looks like there are more failing tests
than I originally thought.


JesseAbram (approved):

ameba23 (approved):
I think its the jump start which fails here, so we could switch this out to use the `IntegrationTestJumpStarted` chainspec. But i'm not sure theres much point as it would not be testing anything which isn't covered elsewhere - the idea of this one was to test the whole run-through.

PR #1285, opened 2025-02-06
Use `spawn_blocking` when processing synedrion messages

ameba23:
This was suggested by @fjarri here:

https://github.com/entropyxyz/manul/pull/91#issuecomment-2639203564

Since synedrion's message processing methods are synchronous and computationally heavy, it does not make sense to run them in async tokio tasks as this may prevent the executor from moving other tasks forward.

This PR uses [`tokio::task::spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) to run them as blocking, syncronous tasks instead.

Unfortunately this does not appear to give any improvement in performance when testing signing with 8 parties.

Since we will be soon swapping this out for the protocol execution loop provided by manul, it probably makes no difference whether or not we merge this.  But gonna put it up anyway.

### DIFF
--- a/crates/protocol/src/execute_protocol.rs
+++ b/crates/protocol/src/execute_protocol.rs
@@ -31,7 +31,7 @@ use synedrion::{
     AuxInfo, KeyResharingInputs, KeyShare, NewHolder, OldHolder, PrehashedMessage,
     RecoverableSignature, ThresholdKeyShare,
 };
-use tokio::sync::mpsc;
+use tokio::{sync::mpsc, task::spawn_blocking};
 
 use crate::{
     errors::{GenericProtocolError, ProtocolExecutionErr},
@@ -100,7 +100,7 @@ where
             let tx = tx.clone();
             let my_id = my_id.clone();
             let destination = destination.clone();
-            tokio::spawn(async move {
+            spawn_blocking(move || {
                 session_arc
                     .make_message(&mut OsRng, &destination)
                     .map(|(message, artifact)| {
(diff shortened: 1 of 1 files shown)


crates/protocol/src/execute_protocol.rs
```diff
@@ -154,7 +154,8 @@ where
                                 let tx = process_tx.clone();
                                 tokio::spawn(async move {
                                     let result = session_arc.process_message(&mut OsRng, preprocessed);
-                                    if tx.send(result).await.is_err() {
+
+                                    if futures::executor::block_on(tx.send(result)).is_err() {
```
ameba23, line 158: Here we have the issue that the channel is async but we are now in a sync block.  However because `tx.send` is likely to be considerably faster than `session.process_message`, i think it still makes sense to do this.

The alternative would be to use `std::sync::mpsc` rather than `tokio::sync::mpsc`, but then we have the problem that we can't use the receiver with `tokio::select!`.

HCastano (approved):
I don't really know what's going on behind the scenes in the protocol session code, but from the `spawn_blocking` link you provided it mentions that if the workload is CPU bound we could benefit from using something like Rayon.

Maybe worth looking into in a different PR.

fjarri:
I am not sure `rayon` would fit naturally here since we need to create new tasks and add them to the pool as we receive messages. But maybe it allows that too. In any case, in my tests `spawn_blocking` works well enough.

PR #1289, opened 2025-02-10
Revert "Bump parity-scale-codec from 3.7.2 to 3.7.4 in the patch-dependencies group"

ameba23:
Reverts entropyxyz/entropy-core#1286 which was automatically merged despite breaking CI.

This is a short term proposal to get CI to pass on master again for now.

Long term i think we should be able to re-open that PR and fix the issue with having duplicate indexes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7795,9 +7795,9 @@ dependencies = [
 
 [[package]]
 name = "parity-scale-codec"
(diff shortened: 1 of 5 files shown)


HCastano (approved):
Thanks for catching this.

Do you mind making a follow up issue so we can follow up with why it merged with CI failures?

Issue #1281, opened 2025-02-05
Deletion strategy for request limit


JesseAbram:
As of now the request limiter will grow at O(n), each new account request increases the storage.

It only needs to be stored for one block so maybe a double map where we can easily delete old data every block or when no longer signer or something

ameba23:
Agree - maybe we can store block number for everybody, rather than a per-user block number.

Eg:
```rust
struct AppState {
  request_limit:  Arc<RwLock<RequestLimiter>>, 
}

struct RequestLimiter {
  current_block_number: BlockNumber,
  request_count: HashMap<VerifyingKey, u32>,
}

impl RequestLimiter {
  fn increment(&mut self, verifying_key: VerifyingKey, block_number: Blocknumber) {
    if block_number != self.block_number {
      self.block_number = block_number;
      // Start fresh for everybody
      self.request_count = HashMap::new();
    }
    request_count.entry(verifying_key).and_modify(|count| *count += 1).or_insert(1);
  }
}
```

JesseAbram:
> Agree - maybe we can store block number for everybody, rather than a per-user block number.
> 
> Eg:
> 
> struct AppState {
>   request_limit:  Arc<RwLock<RequestLimiter>>, 
> }
> 
> struct RequestLimiter {
>   block_number: BlockNumber,
>   request_count: HashMap<VerifyingKey, u32>,
> }
> 
> impl RequestLimiter {
>   fn increment(&mut self, verifying_key: VerifyingKey, current_block_number: BlockNumber) {
>     if current_block_number != self.block_number {
>       self.block_number = current_block_number;
>       // Start fresh for everybody
>       self.request_count = HashMap::new();
>     }
>     self.request_count.entry(verifying_key).and_modify(|count| *count += 1).or_insert(1);
>   }
> }

didn't see this but ya was thinking this, single map, one block number and everycheck if the blocknumber is greater then just clear the whole mapping, simple easy, and won't grow that big

Issue #1280, opened 2025-02-05
Add more items to cache


JesseAbram:
Certain items in the TSS should be in the cache not in the KVDB

mainly all the ```LATEST_BLOCK_NUMBER_*``` in the validate calls

They should grab the current block number - 1 when they spin up and can lose the number if they ever go down no need for the kvdb

PR #1290, opened 2025-02-10
Create cache in app state

### LINKED ISSUE #1282: Merging All HashMap in app state to generalized cache
There are many hashmaps in app_state now, they can maybe all potentially be merged into the generalized cache

JesseAbram:
Closes #1282

### DIFF
--- a/crates/threshold-signature-server/src/helpers/app_state.rs
+++ b/crates/threshold-signature-server/src/helpers/app_state.rs
@@ -1,0 +1,228 @@
+use crate::{
+    chain_api::{get_api, get_rpc, EntropyConfig},
+    launch::Configuration,
+    signing_client::ListenerState,
+    user::api::*,
+};
+use anyhow::anyhow;
+use entropy_kvdb::kv_manager::KvManager;
+use entropy_shared::X25519PublicKey;
+use sp_core::{crypto::AccountId32, sr25519, Pair};
+use std::{
+    collections::HashMap,
+    sync::{Arc, PoisonError, RwLock},
+};
+use subxt::{
+    backend::legacy::LegacyRpcMethods, tx::PairSigner, utils::AccountId32 as SubxtAccountId32,
+    OnlineClient,
+};
+use x25519_dalek::StaticSecret;
+
+/// Represents the state relating to the prerequisite checks
+#[derive(Clone, PartialEq, Eq)]
+pub enum TssState {
+    /// Initial state where no connection to chain node has been made
+    NoChainConnection,
+    /// Connection is made to the chain node but the account may not be yet funded
+    ReadOnlyChainConnection,
+    /// Fully ready and able to participate in the protocols
+    Ready,
+}
+
+impl TssState {
+    fn new() -> Self {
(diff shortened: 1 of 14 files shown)


crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -0,0 +1,247 @@
+use crate::{
```
ameba23, line 1: Maybe add the license header

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    /// Clears a poisioned lock from request limit
+    pub fn clear_poisioned_request_limit(&self) {
+        if self.request_limit.is_poisoned() {
+            self.request_limit.clear_poison()
+        }
+    }
+}
+
+#[derive(Clone)]
```
ameba23, line 172:
```suggestion
/// Application state struct which is cloned and made available to every axum HTTP route handler function 
#[derive(Clone)]
```

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -0,0 +1,247 @@
...
+pub struct AppState {
+    /// Tracks incoming protocol connections with other TSS nodes
+    pub listener_state: ListenerState,
```
ameba23, line 175: If we wanted, we could also put `ListenerState` inside `Cache`.  But it would be a fairly big refactor so maybe something for another PR.

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -0,0 +1,247 @@
...
+    fn is_ready(&self) -> bool {
+        self == &TssState::Ready
+    }
+
+    fn can_read_from_chain(&self) -> bool {
+        self != &TssState::NoChainConnection
+    }
+}
+
+#[derive(Clone)]
```
ameba23, line 46:
```suggestion
/// In-memory store of application state
#[derive(Clone)]
```
or something like that

ameba23 (approved):
:1st_place_medal: makes things a lot simpler i think

For next steps I think we should also move the latest block number storage into the cache - as in these things:  https://github.com/entropyxyz/entropy-core/blob/9b01aff0541529edb6693fcf73989fe9750a550f/crates/threshold-signature-server/src/helpers/launch.rs#L50-L54

JesseAbram:
> 🥇 makes things a lot simpler i think
> 
> For next steps I think we should also move the latest block number storage into the cache - as in these things:
> 
> https://github.com/entropyxyz/entropy-core/blob/9b01aff0541529edb6693fcf73989fe9750a550f/crates/threshold-signature-server/src/helpers/launch.rs#L50-L54

yup have an issue out for it #1280 Imma get to it probably this week but #1281 first

crates/threshold-signature-server/src/helpers/app_state.rs
JesseAbram: not that hard, added it in

PR #1291, opened 2025-02-12
Deletion for request limit

JesseAbram:


crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -181,12 +176,60 @@ impl Cache {
...
+    }
+
     /// Clears a poisioned lock from request limit
     pub fn clear_poisioned_request_limit(&self) {
         if self.request_limit.is_poisoned() {
             self.request_limit.clear_poison()
         }
     }
+
+    /// Write to block numbers
```
HCastano, line 197:
```suggestion
      /// Write the given block number to the `block_number` cache.
```

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -181,12 +176,60 @@ impl Cache {
...
+        self.clear_poisioned_block_numbers();
+        let mut block_numbers = self
+            .block_numbers
+            .write()
+            .map_err(|_| anyhow!("Error getting write write_to_block_numbers lock"))?;
+        block_numbers.insert(key, value);
+        Ok(())
+    }
+
+    /// Check if key exists in request limit
```
HCastano, line 208:
```suggestion
    /// Check if the given block number exists in the cache.
```

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -181,12 +176,60 @@ impl Cache {
...
+        Ok(block_numbers.contains_key(key))
+    }
+
+    /// Reads from block numbers will error if no value, call exists_in_request_limit to check
+    pub fn read_from_block_numbers(&self, key: &String) -> anyhow::Result<Option<u32>> {
+        self.clear_poisioned_block_numbers();
+        let block_numbers = self
+            .block_numbers
+            .read()
+            .map_err(|_| anyhow!("Error getting read read_from_block_numbers lock"))?;
```
HCastano, line 224: We can use the `exists_in_block_numbers()` method here, and then only `get` if that check passes

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -181,12 +176,60 @@ impl Cache {
...
+    pub fn exists_in_block_numbers(&self, key: &String) -> anyhow::Result<bool> {
+        self.clear_poisioned_block_numbers();
+        let block_numbers = self
+            .block_numbers
+            .read()
+            .map_err(|_| anyhow!("Error getting read exists_in_block_numbers lock"))?;
+        Ok(block_numbers.contains_key(key))
+    }
+
+    /// Reads from block numbers will error if no value, call exists_in_request_limit to check
```
HCastano, line 218:
```suggestion
    /// Returns the number of requests handled so far at the given block number.
```

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -181,12 +176,60 @@ impl Cache {
...
+    pub fn read_from_block_numbers(&self, key: &String) -> anyhow::Result<Option<u32>> {
+        self.clear_poisioned_block_numbers();
+        let block_numbers = self
+            .block_numbers
+            .read()
+            .map_err(|_| anyhow!("Error getting read read_from_block_numbers lock"))?;
+        Ok(block_numbers.get(key).cloned())
+    }
```
HCastano, line 226:
```suggestion
    }

```

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -626,44 +628,19 @@ pub async fn request_limit_check(
...
-        // same block incrememnt request amount
-        if request_info.request_amount <= request_limit {
-            cache.write_to_request_limit(
-                verifying_key,
-                RequestLimitStorage {
-                    block_number,
-                    request_amount: request_info.request_amount + 1,
-                },
-            )?;
+        // incrememnt request amount
```
HCastano, line 638:
```suggestion
        // increment request amount
```

crates/threshold-signature-server/src/user/tests.rs
```diff
@@ -1880,6 +1877,20 @@ async fn test_increment_or_wipe_request_limit() {
     .map_err(|e| e.to_string());
     assert_eq!(err_too_many_requests, Err("Too many requests - wait a block".to_string()));
 
+    run_to_block(&rpc, 1).await;
+    // no error whole mapping is empty
```
HCastano, line 1881:
```suggestion
    // no error while mapping is empty
```

HCastano (approved):
I do like that this is simpler than tracking a limit per user.

Two things though:
1. We'll need to increase the current limits, by say, a factor of 100 to account for many users using the same limit
2. We need to consider that a single user could grief that rest of the users "on" a TSS now

JesseAbram:
> I do like that this is simpler than tracking a limit per user.
> 
> Two things though:
> 
> 1. We'll need to increase the current limits, by say, a factor of 100 to account for many users using the same limit
> 2. We need to consider that a single user could grief that rest of the users "on" a TSS now

oh no we still are doing that. We just nuke the whole mapping every block so there isn't a build up in memory.

- Before checking request limit mapping we check to see if a new block has happened,
- If it does we nuke the whole request limit mapping so everyone is back at 0 

also the limit for a single user is stored in parameters and is governed by chain governance

crates/threshold-signature-server/src/helpers/app_state.rs
JesseAbram: ya i had thought about that, but idk that seems less generalizable I can't think of why we would need an error if doesn't exist but maybe we would

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -67,13 +66,15 @@ pub struct Cache {
...
-    pub request_limit: Arc<RwLock<HashMap<String, RequestLimitStorage>>>,
+    pub request_limit: Arc<RwLock<HashMap<String, u32>>>,
     /// Storage for encryption key backups for other TSS nodes
     /// Maps TSS account id to encryption key
     pub encryption_key_backup_provider: Arc<RwLock<HashMap<AccountId32, [u8; 32]>>>,
     /// Storage for quote nonces for other TSS nodes wanting to make encryption key backups
     /// Maps response x25519 public key to quote nonce
     pub attestation_nonces: Arc<RwLock<HashMap<X25519PublicKey, [u8; 32]>>>,
+    /// Collection of block numbers to store
+    pub block_numbers: Arc<RwLock<HashMap<String, u32>>>,
```
ameba23, line 77: +1 for keeping all the stored block numbers in one place.

But here's my nitpick: Why store them in a hashmap and not a struct?  This makes it possible to attempt to look up a block number which doesn't exist.  We know at compile time all the keys we are going to need in our map, so they can just as well be struct fields.

I would do this:
```rust
struct Cache {
  ...
  pub block_numbers: Arc<BlockNumbers>
}

struct BlockNumbers {
  request_limit: RwLock<u32>,
  new_user: RwLock<u32>,
  reshare: RwLock<u32>,
  attest: RwLock<u32>,
}
```
Also this way you can read one while another one is locked for writing.

crates/threshold-signature-server/src/helpers/launch.rs
```diff
@@ -50,6 +50,7 @@ pub const DEFAULT_EVE_MNEMONIC: &str =
 pub const LATEST_BLOCK_NUMBER_NEW_USER: &str = "LATEST_BLOCK_NUMBER_NEW_USER";
 pub const LATEST_BLOCK_NUMBER_RESHARE: &str = "LATEST_BLOCK_NUMBER_RESHARE";
 pub const LATEST_BLOCK_NUMBER_ATTEST: &str = "LATEST_BLOCK_NUMBER_ATTEST";
+pub const LATEST_BLOCK_NUMBER: &str = "LATEST_BLOCK_NUMBER";
```
ameba23, line 53: Should this not be 
```suggestion
pub const LATEST_BLOCK_NUMBER_REQUEST_LIMIT: &str = "LATEST_BLOCK_NUMBER_REQUEST_LIMIT";
```

crates/threshold-signature-server/src/user/api.rs
```diff
@@ -626,44 +628,19 @@ pub async fn request_limit_check(
...
-        if request_info.request_amount <= request_limit {
-            cache.write_to_request_limit(
-                verifying_key,
-                RequestLimitStorage {
-                    block_number,
-                    request_amount: request_info.request_amount + 1,
-                },
-            )?;
+        // increment request amount
+        if request_amount <= request_limit {
```
ameba23, line 639: nit: here we lock to read, unlock ,then lock to write.  we could do both in one go.

ameba23 (approved):
:1st_place_medal:

crates/threshold-signature-server/src/helpers/launch.rs
JesseAbram: although we are using it for request limit all this is a lazy latest block and can be used in other places, the other ones are specific for last time x was updated this one since it resets every block can be generalizable if needed

crates/threshold-signature-server/src/helpers/app_state.rs
JesseAbram: ya I like this but can do in next PR #1280

PR #1293, opened 2025-02-13
Add block numbers to cache

### LINKED ISSUE #1280: Add more items to cache

JesseAbram:
Closes #1280

### DIFF
--- a/crates/threshold-signature-server/src/helpers/app_state.rs
+++ b/crates/threshold-signature-server/src/helpers/app_state.rs
@@ -59,6 +59,7 @@ impl TssState {
     }
 }
 
+/// Fields related to blocknumbers being stored in cache
 #[derive(Debug, Deserialize, Serialize)]
 pub enum BlockNumberFields {
     LatestBlock,
@@ -68,6 +69,7 @@ pub enum BlockNumberFields {
     ProactiveRefresh,
 }
 
+/// Blocknumbers being stored in cache
 #[derive(Default, Clone)]
 pub struct BlockNumbers {
     pub latest_block: Arc<RwLock<u32>>,
--- a/crates/threshold-signature-server/src/helpers/app_state.rs
+++ b/crates/threshold-signature-server/src/helpers/app_state.rs
@@ -22,6 +22,7 @@ use crate::{
 use anyhow::anyhow;
 use entropy_kvdb::kv_manager::KvManager;
 use entropy_shared::X25519PublicKey;
+use serde::{Deserialize, Serialize};
 use sp_core::{crypto::AccountId32, sr25519, Pair};
 use std::{
     collections::HashMap,
@@ -58,6 +59,16 @@ impl TssState {
     }
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub enum BlockNumberFields {
+    LatestBlock,
+}
+
+#[derive(Default, Clone)]
+pub struct BlockNumbers {
+    latest_block: Arc<RwLock<u32>>,
+}
+
 /// In-memory store of application state
 #[derive(Clone)]
 pub struct Cache {
@@ -74,7 +85,7 @@ pub struct Cache {
     /// Maps response x25519 public key to quote nonce
     pub attestation_nonces: Arc<RwLock<HashMap<X25519PublicKey, [u8; 32]>>>,
     /// Collection of block numbers to store
-    pub block_numbers: Arc<RwLock<HashMap<String, u32>>>,
+    pub block_numbers: Arc<BlockNumbers>,
 }
 
 impl Default for Cache {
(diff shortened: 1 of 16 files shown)


crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -58,6 +59,26 @@ impl TssState {
...
+    NewUser,
+    Reshare,
+    Attest,
+    ProactiveRefresh,
+}
+
+/// Blocknumbers being stored in cache
+#[derive(Default, Clone)]
+pub struct BlockNumbers {
+    pub latest_block: Arc<RwLock<u32>>,
```
ameba23, line 75: Since the whole struct is in an `Arc`, i don't think each field needs to also be in an `Arc`.

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -195,40 +216,41 @@ impl Cache {
...
-        if self.block_numbers.is_poisoned() {
-            self.block_numbers.clear_poison()
+    pub fn clear_poisioned_block_numbers(&self, lock: &Arc<RwLock<u32>>) {
+        if lock.is_poisoned() {
+            lock.clear_poison()
+        }
+    }
+
+    /// Gets block number field in block numbers
+    pub fn get_block_number_target(&self, key: &BlockNumberFields) -> Arc<RwLock<u32>> {
```
ameba23, line 247: This works great but another possible way would be to have a custom struct wrapping `RwLock<32>`:

```rust
struct BlockNumber(RwLock<u32>)

impl BlockNumber {
  fn read_from_blocknumber {
    ...
  }
  fn write_to_blocknumber {
    ...
  }
}

struct BlockNumbers {
  latest_block: BlockNumber,
  new_user: BlockNumber, 
  ...
}
```

crates/threshold-signature-server/src/main.rs
```diff
@@ -64,22 +65,22 @@ async fn main() -> anyhow::Result<()> {
...
             // we already can expose the `/info` http route
             if let Err(error) =
-                entropy_tss::launch::check_node_prerequisites(app_state, key_option).await
+                entropy_tss::launch::check_node_prerequisites(app_state.clone(), key_option).await
             {
                 tracing::error!("Prerequistite checks failed: {} - terminating.", error);
                 process::exit(1);
             }
+
+            if let Err(error) = get_block_number_and_setup_latest_block_number(app_state).await {
```
ameba23, line 80: +1 for moving this after the prerequisite checks so that we definitely have a connection to the chain.

crates/threshold-signature-server/src/helpers/tests.rs
```diff
@@ -69,21 +69,20 @@ pub async fn initialize_test_logger() {
     *LOGGER.get_or_init(|| instrumentation.setup()).await
 }
 
-pub async fn setup_client() -> KvManager {
+pub async fn setup_client() -> AppState {
```
ameba23, line 72: +1 this is useful to have access to the whole of AppState in tests

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -195,40 +216,41 @@ impl Cache {
...
-            .map_err(|_| anyhow!("Error getting read exists_in_block_numbers lock"))?;
-        Ok(block_numbers.contains_key(key))
-    }
-
     /// Returns the number of requests handled so far at the given block number.
-    pub fn read_from_block_numbers(&self, key: &String) -> anyhow::Result<Option<u32>> {
-        self.clear_poisioned_block_numbers();
-        let block_numbers = self
-            .block_numbers
+    pub fn read_from_block_numbers(&self, key: &BlockNumberFields) -> anyhow::Result<u32> {
```
ameba23, line 230: grumble grumble (anyhow)

ameba23 (approved):
:star:

Great work i think this is a big improvement

crates/threshold-signature-server/src/helpers/app_state.rs
JesseAbram: it doesn't but if you don't you can't derive clone and it gets a lot harder to work with

crates/threshold-signature-server/src/helpers/app_state.rs
JesseAbram: Ill make an issue lol

crates/threshold-signature-server/src/main.rs
JesseAbram: yeee thought I would let the previous backup code handle for that

crates/threshold-signature-server/src/helpers/app_state.rs
JesseAbram: #1310

PR #1294, opened 2025-02-14
Revert Dependabot PR #1292

HCastano:
Looks like this was auto merged even though the CI failed on it. A simliar thing happened
a few weeks ago, so we should bump up the priority for looking into this.

PR #1301, opened 2025-02-18
Revert lockfile following merging #1294

ameba23:
I made a mistake when resolving conflicts in order to merge https://github.com/entropyxyz/entropy-core/pull/1294

(needed for CI to pass)

And ended up with the change still being present in the lockfile.  This puts it back.


ameba23:
Annoyingly this doesn't seem to actually solve the problem.  I am not sure why. @HCastano @JesseAbram

PR #1304, opened 2025-02-18
Add test reshare e2e to ci on release

### LINKED ISSUE #1270: Add test_reshare_e2e to CI
In #1254 ```test_reshare_e2e``` was added. Add it to the release CI and remove the manual step from the release checklist

JesseAbram:
Closes #1270 

Adds the e2e test to the release check pipeline


.github/workflows/test-resahre-e2e
```diff
@@ -0,0 +1,30 @@
...
+      - release/v[0-9]+.[0-9]+.[0-9]+\+?** # Release with build identifier.
+      # Then, also, basically any release-ish name in the `test/` tagspace.
+      - test/**release/**
+  schedule:
+    # Build the default branch weekly. Our version of "nightly."
+    - cron: 55 4 * * 0   # Every Sunday at 4:55 AM.
+
+jobs:
+  test-e2e-reshare:
+    runs-on: ubuntu-latest
```
ameba23, line 19: Just in case you are not aware - `ubuntu-latest` will give us a standard github runner.  Our other tests run on `core-build-runner` which has much better specs. This might be needed to do a release build - i think the standard runners only have like 14gb storage.

.github/workflows/test-resahre-e2e
```diff
@@ -0,0 +1,30 @@
...
+  push:
+    tags:
+      # First, match SemVer.org conformant tags in the `release/` tagspace.
+      - release/v[0-9]+.[0-9]+.[0-9]+-?**  # Release, or pre-release build.
+      - release/v[0-9]+.[0-9]+.[0-9]+\+?** # Release with build identifier.
+      # Then, also, basically any release-ish name in the `test/` tagspace.
+      - test/**release/**
+  schedule:
+    # Build the default branch weekly. Our version of "nightly."
+    - cron: 55 4 * * 0   # Every Sunday at 4:55 AM.
```
ameba23, line 15: I don't understand what the schedule is for

ameba23 (approved):
:star:

Great to have this in.

One nitpick: we could put this at the beginning of the [main release pipeline](https://github.com/entropyxyz/entropy-core/blob/master/.github/workflows/build-and-release.yaml) so that the release will only get built if this passes.

Can we try this out by tagging this with `test/release/test-release-workflow`?

.github/workflows/test-resahre-e2e
cooldracula: Agree, I think this should be removed if this is running as its own action.  If this gets added to our main release pipeline, per [peg's comment](https://github.com/entropyxyz/entropy-core/pull/1304#pullrequestreview-2624995611), the schedule in that yaml file should remain.

RELEASE_CHECKLIST.md
```diff
@@ -5,9 +5,6 @@ tagged as the final release.
 
 ## Pre-Prep
 - [ ] Inform relevant parties that you're preparing a release (e.g, by posting on Discord)
-- [ ] Manually run e2e reshare test
```
cooldracula, line 8: Would it be useful to update this to mention the tests are running as part of our pipeline, or that the tests are release blocking, if you go with [Peg's good advice](https://github.com/entropyxyz/entropy-core/pull/1304#pullrequestreview-2624995611)?

cooldracula (changes requested):
It looks good.  One ask: can we rename the file to `test-reshare-e2e.yaml`?

.github/workflows/test-resahre-e2e
JesseAbram: yup sounds good, was just following the other release convention

ameba23 (approved):
:1st_place_medal:

cooldracula (approved):

JesseAbram:
About to delete the release pipeline but can confirm it is tested and works https://github.com/entropyxyz/entropy-core/actions/runs/13417074322/job/37480419216

PR #1305, opened 2025-02-19
`/version` HTTP route gives measurement value on production builds

ameba23:
This changes the entropy-tss `/version` HTTP handler to give information about the build:
- Whether or not it is build in 'production' mode
- If it is built with the `production` feature flag present, also give the measurement value.

This allows us to get the measurement value from a new production release in order to add it to the parameters pallet.

It could also be useful for debugging failed attestations as it gives a way of checking the measurement values are what they should be.

This is a breaking change - as any scripts which are programmatically using the output of the `/version` endpoint will need updating. @entropyxyz/system-reliability-engineers  if you rather need JSON output let me know.

This is currently based off https://github.com/entropyxyz/entropy-core/pull/1303 - github will rebase to master if / when that gets merged.


JesseAbram (approved):

HCastano:
@ameba23 as an FYI there's https://github.com/entropyxyz/entropy-core/issues/627 issue which means the version formatting might not be working correctly anyways. 

I would format the measurement details differently though. At least something on a single line, but maybe JSON would be better as you suggested

ameba23:
> I would format the measurement details differently though. At least something on a single line, but maybe JSON would be better as you suggested

I have changed it to JSON. Here is example output:

`{"cargo_package_version":"0.3.0","git_tag_commit":"release/v0.3.0-rc.1-156-g990494c5","build":"NonProduction"}`

@HCastano give me a thumbs up if you think this is better.  Otherwise i can put it back to how it was but on a single line.

crates/threshold-signature-server/src/node_info/api.rs
```diff
@@ -19,10 +19,51 @@ use serde::{Deserialize, Serialize};
 use strum::IntoEnumIterator;
 use subxt::utils::AccountId32;
 
-/// Returns the version and commit data
+/// Version infomation - the output of the `/version` HTTP endpoint
```
HCastano, line 22:
```suggestion
/// Version information - the output of the `/version` HTTP endpoint
```

HCastano (approved):
CI failure looks unrelated to this

PR #1313, opened 2025-02-21
Github actions workflow for building CVM image on release

ameba23:
This builds a confidential virtual machine image containing the entropy-tss binary when a release is made.

Eventually the binary could be built reproducibly from within yocto or guix - see: https://github.com/entropyxyz/meta-entropy-tss/issues/2

But for now, entropy-tss is built in a docker container with our usual release workflow, and the binary bundled into the image.

Pairs with: https://github.com/entropyxyz/yocto-build/pull/6

### DIFF
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -76,15 +76,14 @@ jobs:
         env:
           ENTROPY_TSS_BINARY_URI: ${{needs.build-entropy-tss.jobs.build.outputs.artifact-url}}
         run: |
-          echo "Using artifact url: ${{needs.build-entropy-tss.jobs.build.outputs.artifact-url}}"
           make image-base
-          find . -name 'core-image-minimal-tdx-gcp'
+          echo "Using artifact url: ${{needs.build-entropy-tss.jobs.build.outputs.artifact-url}}"
 
       - name: Upload build artifacts
         uses: actions/upload-artifact@v4
         with:
           name: images
-          path: /home/runner/work/yocto-build/yocto-build/reproducible-build/artifacts-base/core-image-minimal-tdx-gcp.rootfs.wic.tar.gz
+          path: /home/runner/work/entropy-core/entropy-core/reproducible-build/artifacts-base/core-image-minimal-tdx-gcp.rootfs.wic.tar.gz
           if-no-files-found: error
 
   release:
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -18,7 +18,7 @@ jobs:
 
   git-ref-basename:
     uses: ./.github/workflows/git-ref-basename.yaml
-  
+
   test-e2e-reshare:
     runs-on: core-build-runner
     steps:
@@ -55,6 +55,34 @@ jobs:
       DOCKER_HUB_USERNAME: ${{ secrets.DOCKER_HUB_USERNAME }}
       DOCKER_HUB_CI_TOKEN: ${{ secrets.DOCKER_HUB_CI_TOKEN }}
       CI_MACHINE_USER_TOKEN: ${{ secrets.CI_MACHINE_USER_TOKEN }}
+
+  build-entropy-tss-cvm-image:
+    name: Build confidential virtual machine image for entropy-tss
+    needs:
+      - build-entropy-tss
+    runs-on: core-build-runner
+    steps:
+      - name: Check out the yocto-build repository
+        uses: actions/checkout@v3
+        with:
+          repository: 'entropyxyz/yocto-build'
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+            version: latest
+
+      - name: Build
+        env:
+          ENTROPY_TSS_BINARY_URI: ${{need.build-entropy-tss.outputs.artifact-url}}
+        run: |
+          make image-base
+
+      - name: Upload build artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: images
+          path: ~/work/yocto-build/yocto-build/reproducible-build/artifacts-base/core-image-minimal-tdx-gcp.rootfs.wic.tar.gz
 
   release:
     name: Publish new release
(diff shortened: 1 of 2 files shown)


ameba23:
I have not yet tested the binaries from the release pipeline (built in with the `entropyxyz/entropy-tss` docker image) with our CVM image, but just to note that there are some difference from the known working one i have been using (which was built on ubuntu 24.10 with cargo.

Built with `entropyxyz/entropy-tss`:
```
$ ldd entropy-tss_v0.3.0_linux_amd64
	linux-vdso.so.1 (0x0000795a8da88000)
	libssl.so.1.1 => not found
	libcrypto.so.1.1 => not found
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x0000795a8da54000)
	libpthread.so.0 => /usr/lib/libpthread.so.0 (0x0000795a8c1c8000)
	libm.so.6 => /usr/lib/libm.so.6 (0x0000795a8c0d0000)
	libdl.so.2 => /usr/lib/libdl.so.2 (0x0000795a8c0c9000)
	libc.so.6 => /usr/lib/libc.so.6 (0x0000795a8bed7000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x0000795a8da8a000)
```

Known working configuration:

```
$ ldd ../src/entropy/meta-entropy-tss/recipes-core/entropy-tss/entropy-tss
	linux-vdso.so.1 (0x00007d2aca3d0000)
	libssl.so.3 => /usr/lib/libssl.so.3 (0x00007d2ac8924000)
	libcrypto.so.3 => /usr/lib/libcrypto.so.3 (0x00007d2ac8400000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x00007d2ac88f6000)
	libm.so.6 => /usr/lib/libm.so.6 (0x00007d2ac8308000)
	libc.so.6 => /usr/lib/libc.so.6 (0x00007d2ac8116000)
	/lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007d2aca3d2000)   
```

ameba23:
The issues in my [above comment](https://github.com/entropyxyz/entropy-core/pull/1313#issuecomment-2674621631) are now fixed.

But there are some on-going issues with this:
- https://github.com/entropyxyz/entropy-core/issues/1328
- [upload-artifact](https://github.com/actions/upload-artifact?tab=readme-ov-file#zip-archives) gives us a URL to a zip archive, containing binaries for both x86 and arm. What we expect in [the entropy-tss bitbake recipe](https://github.com/entropyxyz/meta-entropy-tss/blob/main/recipes-core/entropy-tss/entropy-tss.bb) is the URL of the x86 binary (not in a zip archive).  So **either** we need to update that script to handle unzipping the file and storing it somewhere, **or** we need a way of uploading the unzipped x86 binary somewhere using something other than upload-artifact.
- To generate non-mock attestations, `entropy-tss` needs to be compiled with the `production` feature flag present. Likewise, to verify real PCK certificate chains and not allow mock quotes, `entropy` also needs to be compiled with the `production` feature flag present. So we need a way to pass this flag to the docker build setup.

Then the final problem - which i propose to do in a follow-up PR.  We need to boot the CVM image, and hit the `/version` http route in order to get the measurement values which need to be included in the TDX testnet chainspec:

https://github.com/entropyxyz/entropy-core/blob/f436763b80076dd0213a18e333040f1066db1c06/node/cli/src/chain_spec/tdx_testnet.rs#L25

This means we ideally need to programmatically deploy the CVM to google cloud, boot it and get the measurement. Which would mean setting up a service account for GCP.

We do have some scripts for generating the measurements from CVM images without booting them:

https://github.com/entropyxyz/yocto-build/blob/fd348438306218496e09c6e41159fa3407ea7a3d/Makefile#L26-L36

But i have not yet tested them and so for now i would propose we deploy and boot the image to be sure.

ameba23:
Im also having [this issue with upload-artifact](https://github.com/actions/upload-artifact/issues/51).

Possible ways forward:

1. Switch to hosting the binary with a binary pastebin such as http://0x0.st/ with a retention time of 24 hours. This means we are dependent on that service existing / working to run our pipeline.  Which i realise is not ideal.
2. Pass a github authorization token when downloading the file:

```
wget --header='Authorization: token $GITHUB_TOKEN' -O entropy-tss $ENTROPY_TSS_URI
```

This would mean passing this token to the docker image used to do the poky build:

```
env:
  MY_TOKEN: ${{ secrets.GITHUB_TOKEN }}
run: |
          make image-base
```

ameba23:
Having tried several ways to do this with yocto, it seems they made it intentionally difficult to dynamically build images with custom inputs - everything which effects the build needs to be committed in order to guarantee reproducibilty.

So you can download a file from a remote resource as we are doing here, but you need to provide its hash and the hash needs to be committed to the image build scripts.

This issue stands whether we copy in the pre-built binary as we are doing now, or build from rust source in the yocto recipe. As even then it would need to be hard-coded into the repo which revision of the source code to build from.

Possible ways forward:
- Separate the entropy-core release pipeline from the CVM image build pipeline.  So the CVM image will not be part of the entropy-core release. Rather, the entropy-core release happens first and then we make a CVM image release using it.
- Always build releases from a 'release' branch of entropy-core and have the yocto recipe pull from there. I think this could work but it means we can't make successful test releases from other branches.
- Programmatically make  a commit to `meta-entropy-tss` from within the release pipeline.

HCastano:
@ameba23 

> Possible ways forward:
> - Separate the entropy-core release pipeline from the CVM image build pipeline.  So the CVM image will not be part of the entropy-core release. Rather, the entropy-core release happens first and then we make a CVM image release using it.
> - Always build releases from a 'release' branch of entropy-core and have the yocto recipe pull from there. I think this could work but it means we can't make successful test releases from other branches.
> - Programmatically make  a commit to `meta-entropy-tss` from within the release pipeline.

I would be okay with (1). This seems like a reasonable approach. What would the downsides be here though?

For (2), we can probably do something similar to our existing CI with a `test/release-*` branch nomenclature which runs through the entire release pipeline.

It seems like you went with (3) in #1339 though, why go with that approach over the others?

ameba23:
Closing this because we came to the decision to keep the entropy-core and CVM release pipelines separate.

Issue #1328, opened 2025-03-04
Build and release workflow fails following release of rustup 1.28.0


ameba23:
While working on https://github.com/entropyxyz/entropy-core/pull/1313 which involves making some changes to the release pipeline, the pipeline ran successfully several times, but then i started getting failures which appeared to be unrelated to changes i had made.

So i tried making a test release from the head of master branch.  `build-entropy` and `build-entropy-tss` both failed: https://github.com/entropyxyz/entropy-core/actions/runs/13649121649

It looks like the problem might be that when building for arm64, only the arm64 toolchain is installed. But the x86_64 toolchain is needed as that is the platform the container is running on:
```
#21 [linux/amd64 build 7/7] RUN --mount=type=ssh     --mount=type=secret,id=credentials     /bin/bash -o pipefail -c '$(         grep "export GITHUB_TOKEN" /run/secrets/credentials         | cut -d "#" -f 1 | tr -d "\""         || echo true     )     && [ -n "${GITHUB_TOKEN}" ]         && git config --global             url.https://entropyxyz:${GITHUB_TOKEN}@github.com/entropyxyz.insteadOf             ssh://git@github.com/entropyxyz         || true'     && mkdir -p ~/.ssh     && echo "github.com ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOMqqnkVzrm0SdG6UOoqKLsabgH5C9okWi0dh2l9GKJl"         > ~/.ssh/known_hosts     && if [ "amd64" = linux/amd64 ]; then         export RUST_PLATFORM=x86_64;         export BINUTILS_PATH=/usr/bin;     else         export RUST_PLATFORM=aarch64;         export BINUTILS_PATH=/usr/${RUST_PLATFORM}-linux-gnu/bin;     fi; $HOME/.cargo/bin/rustup target add "${RUST_PLATFORM}-unknown-linux-gnu"     && $HOME/.cargo/bin/rustup component add --target wasm32-unknown-unknown rust-src     && if [ "li
#21 0.079 error: toolchain 'stable-x86_64-unknown-linux-gnu' is not installed
#21 0.079 help: run `rustup toolchain install stable-x86_64-unknown-linux-gnu` to install it
#21 0.087 error: toolchain 'stable-x86_64-unknown-linux-gnu' is not installed
#21 0.087 help: run `rustup toolchain install stable-x86_64-unknown-linux-gnu` to install it
[...]

PR #1329, opened 2025-03-04
Dockerfile - Always install the native rust toolchain and add other platforms as a target

ameba23:
This is a possible solution to https://github.com/entropyxyz/entropy-core/issues/1328

Rather than installing the rust toolchain of the target platform, we always install the x86 toolchain (since that is the platform we are building on), and when building for arm64 add that as a target using `cargo target add`.

If i have understood correctly we wont be able to see if this fixes the issue tagging a release on this branch, because the release pipeline pulls the build container image from Dockerhub. Im not sure how to test it without pushing these changes to Dockerhub.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,13 +33,12 @@ RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs \
     | sh -s -- -y --no-modify-path --profile minimal \
       --default-toolchain none \
     && $HOME/.cargo/bin/rustup default "${RUST_STABLE_VERSION}" \
-    && if [ "amd64" = ${TARGETPLATFORM#"linux/"} ]; then \
-        export RUST_PLATFORM=x86_64; \
-    else \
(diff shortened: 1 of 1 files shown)


ameba23:
@entropyxyz/system-reliability-engineers as i mentioned on the call i have the feeling that if we merge this PR and make a release tag, the process will still fail.

To demonstrate what i mean i just pushed a release tag on this branch and you can see it fails with the same error as before: https://github.com/entropyxyz/entropy-core/actions/runs/13663456698/job/38199624997

I think this is because the image used by the [docker build-push-action](https://github.com/docker/build-push-action) here:

https://github.com/entropyxyz/entropy-core/blob/8e92dc1715ec55b603175c3851688165baf0eba5/.github/workflows/container-build-and-push.yaml#L83

is coming from dockerhub and not directly from the `Dockerfile` in this repo.

I am making sense? We can for sure go over this on a call, but sometimes i can express myself better in writing.

Issue #1332, opened 2025-03-05
CI `node-test` job doesn't run on non-Rust changes


HCastano:
Our `node-test` CI job doesn't run unless there are changes in Rust files living in a
particular set of folders.

The motivation for this was to avoid having to run an expensive and long CI check for
non-Rust related changes (e.g, a `README` edit).

Now that we have branch protection rules though, we can't merge a PR unless that check
has passed - which can't happen on PRs which don't have changes in those specific Rust
folders (e.g https://github.com/entropyxyz/entropy-core/pull/1330).

We should change the CI job logic so that it succeeds in the case that no Rust code
changes are made. I think this can be done from the workflow file itself, so we don't
need to muck with the branch protection rules.

ameba23:
It seems strange that the branch protection rules work this way - ideally we want to block merging if a CI task has failed - not if it has not been triggered.

PR #1333, opened 2025-03-05
Always run node-test workflow but only actually run the tests if rust code has been modified

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1332

Probably there is a less verbose way to do this.

### DIFF
--- a/.github/workflows/build-and-run-node-test.yaml
+++ b/.github/workflows/build-and-run-node-test.yaml
@@ -18,7 +18,7 @@ jobs:
             echo "changes_detected=false" >> $GITHUB_ENV
           fi
       - name: Set output
-        run: echo "::set-output name=changes_detected::${{ env.changes_detected }}"
+        run: echo "changes_detected=${{ env.changes_detected }}" >> $GITHUB_OUTPUT
 
   node-test:
     needs: check-paths
--- a/.github/workflows/build-and-run-node-test.yaml
+++ b/.github/workflows/build-and-run-node-test.yaml
@@ -24,11 +24,11 @@ jobs:
 
   node-test:
     needs: check-paths
-    if: needs.check-paths.outputs.run_tests == 'true'
     runs-on: core-build-runner
     steps:
       - uses: actions/checkout@v4
       - name: Increase swap
+        if: needs.check-paths.outputs.run_tests == 'true'
         run: |
           sudo swapoff -a
           sudo dd if=/dev/zero of=/swapfile bs=1G count=8
@@ -37,8 +37,10 @@ jobs:
           sudo swapon /swapfile
           grep Swap /proc/meminfo
       - name: Install dependencies
+        if: needs.check-paths.outputs.run_tests == 'true'
         uses: ./.github/actions/install-dependencies/
       - name: Build entropy-protocol nodejs package
+        if: needs.check-paths.outputs.run_tests == 'true'
         run: |
           cd crates/protocol
           make build-nodejs-testing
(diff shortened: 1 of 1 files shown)


ameba23:
Hmm not sure why i am still seeing this:

.github/workflows/build-and-run-node-test.yaml
```diff
@@ -2,14 +2,29 @@
 name: "Build and test"
 on:
   push:
-    paths:
-      - "node/**"
-      - "crates/**"
-      - "pallets/**"
-      - "runtime/**"
+    branches:
+      - "*"
```
HCastano, line 6:
```suggestion
```

Just the [`on: push`](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-a-single-event) should be sufficient here

.github/workflows/build-and-run-node-test.yaml
```diff
@@ -2,14 +2,29 @@
...
+          if git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep -E "^(node/|crates/|pallets/|runtime/)"; then
+            echo "changes_detected=true" >> $GITHUB_ENV
+          else
+            echo "changes_detected=false" >> $GITHUB_ENV
+      - name: Set output
+        run: echo "::set-output name=changes_detected::${{ env.changes_detected }}"
+
+  node-test:
+    needs: check-paths
+    if: needs.check-paths.outputs.run_tests == 'true'
```
HCastano, line 27: I think here we need a second case here. Like `if needs... != 'true'`, `run: | ... something that evaluates to true/exits the job`

HCastano (commented):
Your screenshot didn't fully upload btw - but I assume you meant that the job isn't running

.github/workflows/build-and-run-node-test.yaml
ameba23: Cant get that to work: https://github.com/entropyxyz/entropy-core/actions/runs/13693650026/job/38291254216?pr=1333

.github/workflows/build-and-run-node-test.yaml
ameba23: I tried having a separate 'no-op' workflow for this case.  But it seems the branch protection rules are interested in the particular job 'node-test' running succesfully, not that that the whole 'Build and test' workflow running successfully.

I don't have access to the branch protection rules settings so i can't check if there is a way round that by modifying them.

My alternative. but ugly, is to run each step conditionally so that the node-test always runs. But even that doesn't seem to work.

ameba23:
Can't get this to work. I am starting to wonder if the problem is something else - as i would expect to be able to see the logs of the `node-test` job even if the case that it skips the tests (as it does log a message in this case).  But the job is just sitting there in a 'pending' state.

ameba23:
I had a hunch it might be because node-test uses `core-build-runner` and the other check (which do run) use github provided runners. But changing that didn't help either.

I feel like i can't move forwards with this without having access to the repository settings.

.github/workflows/build-and-run-node-test.yaml
HCastano: Yeah you're right, it does operate on _jobs_, not _workflows_.

> My alternative, but ugly, is to run each step conditionally so that the node-test always runs. But even that doesn't seem to work.

So having conditional steps within the `node-test` job as opposed to separate jobs, right?

ameba23:
Thank you @HCastano ! Looks like this is working now.

.github/workflows/build-and-run-node-test.yaml
```diff
@@ -3,11 +3,31 @@ name: "Build and test"
 on: ["push"]
 
 jobs:
- node-test:
+  check-paths:
```
HCastano, line 6: nit: I think this'll make it a bit easier to understand what the check is going from the other jobs

```suggestion
  code-paths-changed:
```

E.g, we now do: "needs.code-paths-changed.outputs.run_tests == true"

HCastano (approved):
Nice work, thanks

PR #1339, opened 2025-03-07
Update release workflow to push to meta-entropy-tss before building CVM image

ameba23:
This is a sub-PR to https://github.com/entropyxyz/entropy-core/pull/1313

This makes a commit to meta-entropy-tss during the release pipeline.  For explanation of why see https://github.com/entropyxyz/entropy-core/pull/1313#issuecomment-2705728497

This currently will not work because it requires setting up a github app with an installation access token to push to that repo which i am unable to do because i don't have the right permissions.


ameba23:
Closing this because we came to the decision to keep the entropy-core and CVM release pipelines separate

PR #1345, opened 2025-03-11
Add 90 minute timeout for node-test workfow

### LINKED ISSUE #1336: Reduce `node-test` job timeout
Looks like the [`node-test` job ran for 6 hours](https://github.com/entropyxyz/entropy-core/actions/runs/13694167823/job/38292789647?pr=1334) in #1334 before timing out.

Two things to look into here:
1. We should reduce the job timeout to something like 90 minutes (current job runs in ~50 mins), there's no need to be spinning our wheels for that long
2. We should figure out _why_ it spun its wheels for so long since this is something I haven't seen before.

cc @entropyxyz/system-reliability-engineers

cooldracula:
Fixes #1336

### DIFF
--- a/.github/workflows/build-and-run-node-test.yaml
+++ b/.github/workflows/build-and-run-node-test.yaml
@@ -5,6 +5,7 @@ on: ["push"]
 jobs:
  node-test:
     runs-on: core-build-runner
+    timeout-minutes: 90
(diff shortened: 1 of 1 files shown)


ameba23 (approved):
:star:

PR #1346, opened 2025-03-11
Fix enum variant name for production build details

ameba23:
This is a small fix needed to compile `entropy-tss` in with the `production` feature enabled.

Ideally problems like this should be picked up in CI - we should probably be running `cargo check --features=production`.


HCastano (approved):
Good catch.

Yep, we can probably run with `--all-features` to make this a bit more robust.

PR #1349, opened 2025-03-12
Release in production mode unless 'non-TDX' is specified in release tag

ameba23:
This makes it possible to build entropy with the `production` feature flag in the release pipeline if the release tag contains the string 'production'.

An alternative would be to make production mode default and 'mock' need to be specified.  But i think this is fine for now.


HCastano (approved):
I would rather have it the other way around. I'd like the `X.Y.Z` tag to be TDX enabled by default, and something like `X.Y.Z+non-TDX` otherwise.

We'll also want to update the release checklist and maybe the [release management](https://github.com/entropyxyz/meta/wiki/Release-management) wiki to reflect whatever convention we go with.

PR #1350, opened 2025-03-12
Fix attest validate function

JesseAbram:
Fixes the attestation validate function's race condition. Not sure if this is actually needed or we should remove this attest code


HCastano (approved):
> ... we should remove this attest code

Do you mind elaborating on this? Like just removing the validation during the attest flow, or the attest flow generally?

ameba23 (approved):
:100:

I propose to actually remove this code in an upcoming PR.  But i would like to merge this first, anyway. As i think it's possible we will decide to put this code back in if there comes a situation where we need attestations to be initiated by the chain.

Issue #1351, opened 2025-03-13
Dead code related to attestation


ameba23:
In https://github.com/entropyxyz/entropy-core/pull/1109 the attestation flow was changed to be 'pull based'.
Previously, an offchain worker would request a quote from a TSS node in response to some event. 

Since 1109, the node operator requests a quote from the TSS, the TSS requests a nonce from the chain which is included in the quote, the quote is returned to the client (the node operator) and the node operator then submits the quote in an extrinsic signed with the stash account (rather than the TSS account doing this as before).

There is some code related to the old way of doing this still around - some ignored tests, the propagation pallet code, and the POST `/attest` endpoint.

We decided to keep this stuff around while we were still designing the attestation feature as there might be still some cases where we need an attestation to be initiated by the chain rather than the node operator.

I propose to remove this code for now to keep things clean.

But i think it could maybe be still useful a possible solution to the issue of having TSS nodes pre-configured at genesis (issue for this coming soon).

PR #1352, opened 2025-03-13
Remove dead code releating to attestation

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1351


ameba23:
Not sure if we still need to re-generate benchmarks etc for the propagation pallet following this

JesseAbram (approved):

Issue #1354, opened 2025-03-13
Include provisioning certification key (PCK) in entropy-tss `/info` output


ameba23:
For the TSS nodes which are hardcoded at genesis, the proposed release workflow is:

- Build the CVM image at the entropy-tss commit we intend to make a release from
- Start all 4 TSS nodes
- Call `/version` on one of them to the the measurement value for this build
- Call `/info` on each of them to get their public keys
- Make a PR to `entropy-core` adding the measurement value and TSS public keys to the TDX testnet chainspec
- Build an `entropy-core` release, linking to the assocated CVM image in the release notes
- Start the 4 chain nodes

The issue is that `/info` currently only gives us the TSS account ID and x25519 public key, not the PCK.  To get the PCK, the `/info` handler should generate a quote internally.

Tagging @entropyxyz/system-reliability-engineers as this is relevant to the deployment workflow

Issue #1353, opened 2025-03-13
Attestation for TSS nodes which are added at genesis


ameba23:
With our current testnet deployments, we start the chain we 4 chain nodes and 4 TSS nodes pre-configured at genesis.

This means the TSS nodes are presumed to be pre-attested: their provisioning certification keys (PCKs) are hard coded into the genesis config, and they never have to make an on-chain attestation, unless they later change their keys or IP address.  Only nodes which later join the running network have to make an attestation.

Probably this is fine for testing, but in a production network we ideally want anyone to be able to verify that all TSS nodes are playing by the rules.

There are several possible ways we could resolve this:
1. Initially start the chain only with chain nodes. That is, there will be at least 2 chain nodes on the network which do not have associated TSS nodes, and all TSS nodes must join via the `validate` extrinsic.
2. Have the chain initiate an attestation request for the genesis TSS nodes at block 1. This means using the [dead attestation code](https://github.com/entropyxyz/entropy-core/pull/1352). I am not keen on this as if the attestation fails, we already have assumed it wouldn't and stored their PCK.
3. In the genesis config, include a full quote with the TSS node info for the pre-configured nodes. In theory this means anybody can verify that the initial TSS nodes are playing by the rules. Nonce-freshness is not guaranteed - im not sure if this is an issue.

ameba23:
Related issue which also explains the release / deployment workflow: https://github.com/entropyxyz/entropy-core/issues/1354

PR #1357, opened 2025-03-14
Include Provisioning Certification Key (PCK) in the entropy-tss `/info` HTTP route output

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1354

Example output of `/info` on this branch:

```json
{"ready":false,"tss_account":"5Dy7r8pTEoJJDGRrebQvFyWWfKCpTJiXxz7NxbKeh8zXE7Vk","x25519_public_key":[40,170,149,217,225,231,193,134,157,146,161,94,118,146,134,201,179,206,106,186,35,6,93,138,104,203,205,68,208,90,255,7],"provisioning_certification_key":[2,35,153,56,144,219,98,192,9,186,39,114,167,154,75,24,93,39,159,234,180,105,135,89,110,203,179,93,192,164,177,214,78]}
```

### DIFF
--- a/crates/threshold-signature-server/src/attestation/api.rs
+++ b/crates/threshold-signature-server/src/attestation/api.rs
@@ -121,7 +121,7 @@ pub fn get_pck(
 ) -> Result<BoundedVecEncodedVerifyingKey, AttestationErr> {
     let quote_raw = configfs_tsm::create_quote([0; 64])
         .map_err(|e| AttestationErr::QuoteGeneration(format!("{:?}", e)))?;
-    let quote = Quote::from_bytes(&quote_raw).unwrap();
+    let quote = Quote::from_bytes(&quote_raw)?;
     let pck = quote.verify().map_err(|e| {
         AttestationErr::QuoteGeneration(format!("Could not get PCK from quote {:?}", e))
     })?;
--- a/crates/threshold-signature-server/src/attestation/api.rs
+++ b/crates/threshold-signature-server/src/attestation/api.rs
@@ -30,7 +30,7 @@ use axum::{
 use entropy_client::user::request_attestation;
 use entropy_shared::{
     attestation::{compute_quote_measurement, QuoteContext, QuoteInputData, VerifyQuoteError},
-    OcwMessageAttestationRequest,
+    BoundedVecEncodedVerifyingKey, OcwMessageAttestationRequest,
 };
 use parity_scale_codec::Decode;
 use serde::Deserialize;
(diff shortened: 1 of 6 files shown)


crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -188,6 +188,44 @@ pub fn get_measurement_value() -> Result<[u8; 32], AttestationErr> {
     Ok(compute_quote_measurement(&quote))
 }
 
+/// Get our Provisioning Certification Key (PCK)
+/// This generates a quote and gets the public key from it
+#[cfg(feature = "production")]
+pub fn get_pck(
+    _tss_account: SubxtAccountId32,
```
ameba23, line 195: This argument is not used, but its there so that we have the same API regardless of feature flags, so we can always use the function called `get_pck`

JesseAbram (approved):

HCastano (approved):
Looks good. I'd also add a `CHANGELOG` entry since it changes the format of the `/info` endpoint.

As for the release workflow you proposed, it seems a bit fragile. What happens in the case where we need to terminate the TSS servers put in the chainspec between the CVM and final release steps?

I guess maybe one thing do to here is to build a custom chainspec later on - like after we have a running network, and then publish that back to the repo as a reference. It would decouple the `entropy-core` release from the chainspec build step

ameba23:
> As for the release workflow you proposed, it seems a bit fragile. What happens in the case where we need to terminate the TSS servers put in the chainspec between the CVM and final release steps?
>
I agree. I don't think this is a complete solution as it currently is as external people can't verify the attestations of the genesis nodes.  I'd be up for switching to option number one in https://github.com/entropyxyz/entropy-core/issues/1353
 
But i think this is good enough to be able to make a testnet release and see everything working.

If we have to terminate the TSS servers we have these options:
- They recover their keys automatically using the recovery workflow. This won't yet work until https://github.com/entropyxyz/meta-entropy-tss/pull/5 is merged, and also backups are not made until all TSS nodes are connected to their chain nodes. It also wont work if all 4 nodes are taken down at the same time.
- We update their keys to the new ones using the `change_tss_accounts` extrinsic, submitted with the associated stash account.
- If all else fails we will have to make another commit to the chainspec and re-release the chain node. 

> I guess maybe one thing do to here is to build a custom chainspec later on - like after we have a running network, and then publish that back to the repo as a reference. It would decouple the `entropy-core` release from the chainspec build step

I don't totally get what you mean.  Once we have a running network (assuming that includes the chain) we can't change the genesis config. Maybe we can move this conversation to https://github.com/entropyxyz/entropy-core/issues/1353

PR #1359, opened 2025-03-18
Validate program blobs onchain

JesseAbram:


JesseAbram:
ok ya this won't work 

https://github.com/bytecodealliance/wasmtime/issues/6540

ameba23:
> ok ya this won't work

Shame.  I'm curious how come it seemed so much like it was gonna work.

Issue #1197, opened 2024-12-04
`entropy-tss` application data is stored in current working directory in which binary is run


ameba23:
The directory where the key-value database is stored for `entropy-tss` is determined by the function `entropy_kvdb::get_db_path`: https://github.com/entropyxyz/entropy-core/blob/3df5d39c2fee019280dcce3f084a59dc2c06d803/crates/kvdb/src/lib.rs#L21

It calls [`std::env::current_dir`](https://doc.rust-lang.org/std/env/fn.current_dir.html) to get the current working directory, then adds `.entropy/production/db` or `.entropy/testing/db`.

Usually when i run entropy-tss from the command line this is fine - it creates a `.entropy` in whatever directory i am in, and i can easily clean it up afterwards.

However recently when setting it up to run continuously in a VM enclave using a systemd service, it took me a while to figure out where the db was being stored as the process was being started by systemd, not my personal user.

The files end up in the root directory (`/.entropy`). This is not a conventional place to put a hidden directory containing application data.

I don't know about other platforms but i think for linux the convention is to put writable system-wide application data in `/var/lib/<application name>` or `/etc/<application name>` and user application data should follow the XDG base directory specification (eg: with a crate such as [`microxdg`](https://docs.rs/microxdg)).

My proposal would be that if the `production` feature flag is present, we should attempt to use either `/etc/entropy` or `/var/lib/entropy`, and if we get a permissions error fall back to either the current working directory, or XDG directory (`~/.local/share/entropy` or `~/.local/state/entropy`).

PR #1361, opened 2025-03-19
Change `entropy-tss` storage location for TDX production builds

ameba23:
This relates to this issue: https://github.com/entropyxyz/entropy-core/issues/1197

This changes the directory where `entropy-tss` stores its key-value store but only for TDX production builds.  The base directory for these builds is now `/persist` rather than the working directory where the binary is run.

This pairs with https://github.com/entropyxyz/meta-entropy-tss/pull/5 which mounts a persistent filesystem to the path `/persist` (the rest of the filesystem is in-memory only).

Rather than fix this issue complete by setting the storage directory to something conventional, i am trying to minimize the amount of breaking changes to our current setup since we are talking about doing another non-TDX deployment soon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ runtime
 - Attestation quote verification should check both run-time and build-time measurement values ([#1303](https://github.com/entropyxyz/entropy-core/pull/1303))
 - `/version` HTTP route gives measurement value on production builds ([#1305](https://github.com/entropyxyz/entropy-core/pull/1305))
 - Include Provisioning Certification Key (PCK) in the entropy-tss `/info` HTTP route output ([#1357](https://github.com/entropyxyz/entropy-core/pull/1357))
+- Change entropy-tss storage location for TDX production builds ([#1361](https://github.com/entropyxyz/entropy-core/pull/1361))
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ runtime
 - Attestation quote verification should check both run-time and build-time measurement values ([#1303](https://github.com/entropyxyz/entropy-core/pull/1303))
 - `/version` HTTP route gives measurement value on production builds ([#1305](https://github.com/entropyxyz/entropy-core/pull/1305))
 - Include Provisioning Certification Key (PCK) in the entropy-tss `/info` HTTP route output ([#1357](https://github.com/entropyxyz/entropy-core/pull/1357))
+- Change entropy-tss storage location for TDX production builds ([#1361](https://github.com/entropyxyz/entropy-core/pull/1361))
```
ameba23, line 79: I am not considering this a breaking change because until now we have not been doing deployments with production builds.  This should not effect our current deployment workflow.

crates/kvdb/src/lib.rs
```diff
@@ -18,10 +18,26 @@ pub mod encrypted_sled;
 pub mod kv_manager;
 use std::{fs, path::PathBuf};
 
-pub fn get_db_path(testing: bool) -> String {
-    let mut root: PathBuf = std::env::current_dir().expect("could not get home directory");
+/// The intended environment entropy-tss is build for
```
HCastano, line 21:
```suggestion
/// The intended environment entropy-tss is built for
```

HCastano (approved):
Nice this seems very reasonable. 

While not a breaking change, @entropyxyz/system-reliability-engineers may still be interested in this change

PR #1366, opened 2025-03-20
Update release checklist with instructions to build CVM image for TDX

ameba23:
This add instructions to the release checklist adding the workflow for a TDX deployment as things currently are.

This relates to https://github.com/entropyxyz/entropy-core/issues/1353 and https://github.com/entropyxyz/entropy-core/issues/1354

I don't necessarily want to merge this as it is now, as things are going to get better once we:
- Have a deployment setup using GCP. 
- Can automate getting the TDX measurement value from a CVM image build: https://github.com/entropyxyz/yocto-build/issues/13

But i wanted to put this up to show @entropyxyz/system-reliability-engineers what the deployment flow is going to need to look like (deploy TSS nodes before chain nodes).

And, when everybody complains at what a complicated and fragile procedure this is, to use it as argumentation that we have the genesis chain nodes not have associated TSS servers. I think for the first TDX testnet deployment we should leave things as they are here, but we should work towards being able to have all TSS servers be deployed and attested after genesis.


RELEASE_CHECKLIST.md
```diff
@@ -60,13 +60,36 @@ tagged as the final release.
     - Make sure nothing has gone into `master` in the meantime or you may have you repeat the
       previous steps!
 
+## Build CVM image for `entropy-tss`
+
+If this is a release intended for running a test network on TDX using Google Cloud Platform, do the
+following:
+
+- Make a PR to [`meta-entropy-tss`](https://github.com/entropyxyz/meta-entropy-tss) updating the
+  revision of entropy-tss to the release branch: [here](https://github.com/entropyxyz/meta-entropy-tss/blob/b621096b36ab13703f72954dab37fd47c2f642e9/recipes-core/entropy-tss/entropy-tss.bb#L42-L43).
```
HCastano, line 69: Can you clarify what the `SRCREV` is, it is just the Git hash of the commit we want to release at?

If so, why do we also need to specify the branch in the `SRC_URI`?

RELEASE_CHECKLIST.md
```diff
@@ -60,13 +60,36 @@ tagged as the final release.
...
+- Merge that PR.
+- Make a release in the [`yocto-build`](https://github.com/entropyxyz/yocto-build) repo, by cloning
```
HCastano, line 71:
```suggestion
- Make a release in the [`yocto-build`](https://github.com/entropyxyz/yocto-build) repo by cloning
```

RELEASE_CHECKLIST.md
```diff
@@ -60,13 +60,36 @@ tagged as the final release.
...
+  the repo and doing `git tag -s release/vX.Y.Z-rc.1 && git push origin release/vX.Y.Z-rc.1`
+- The CVM image will be built in CI.
+- Deploy the TSS nodes and get their details. There will soon be an automated way to do this but for
+  now we just have [this script](https://github.com/entropyxyz/yocto-build/blob/main/gcp-deploy) to
+  deploy a single node which you can use like this:
+   - Download the CVM image from the release artifacts of the build you just created
+   - Run the script with the name of the release tag and the path to the image: `./gcp-deploy release/vX.Y.Z.rc1 core-image-minimal-tdx-gcp.rootfs.wic.tar.gz`
```
HCastano, line 78: Is the `core-image-minimal...` bit static? Or under what circumstances would it change?

RELEASE_CHECKLIST.md
```diff
@@ -60,13 +60,36 @@ tagged as the final release.
...
+   - Get the IPs of the TSS nodes (listed under `EXTERNAL_IP` in the output of the deploy script)
+   - On one of them, get the TDX measurement value of this build from the output of `curl <ip address>:3001/version`
+   - For each of them, get the TSS public keys from the output of `curl <ip address>:3001/info`
+- Make a commit to the release branch putting the measurement value and TSS public keys in the TDX
```
HCastano, line 82: So this is the step I was saying in a previous PR that we might be able to skip.

E.g, in the release branch we do something like:

```
build-spec --chain tdx-testnet > tdx-testnet.json
```

And then patch that _later_ (post-release but pre-deployment) with all this info.

After we have the network up and running we can then commit the patched file back into the repo so other people want to join the testnet can just use it directly, e.g `entropy --chain entropy-core/res/patched-tdx-testnet.json`

HCastano (approved):
Approving since this is the reality of the release process today and it might be helpful to have it in.

> And, when everybody complains at what a complicated and fragile procedure this is, to use it as argumentation that we have the genesis chain nodes not have associated TSS servers. I think for the first TDX testnet deployment we should leave things as they are here, but we should work towards being able to have all TSS servers be deployed and attested after genesis.

lol - I agree with you though. We can start with this and iterate towards something better as those other issues get closed

RELEASE_CHECKLIST.md
ameba23: i don't think you need to specify both.  i expect you can leave off the branch bit and have `SRCREV` be either a commit hash, tag, or branch name.  But i haven't checked docs / tried it.

RELEASE_CHECKLIST.md
ameba23: Its the yocto 'distribution name'.  We might end up deciding to change it somepoint.  I guess we could make this generic by taking the first `*.rootfs..wic.tar.gz` file that we find

RELEASE_CHECKLIST.md
ameba23: Ah cool! That would be great, and save a lot of pain here.

ameba23:
Putting this back to 'draft' since it looks like the chainspec-modifying part will be moved to the deployment pipeline rather than the entropy-core release pipeline, which will simplify things here

ameba23:
I have removed the stuff about deploying TSS servers and getting their details - this will be handled elsewhere and not be part of the release check list.  For future reference, this is the bit i removed:

Deploy the TSS nodes and get their details. There will soon be an automated way to do this but for
now we just have [this script](https://github.com/entropyxyz/yocto-build/blob/main/gcp-deploy) to
deploy a single node which you can use like this:
   - Download the CVM image from the release artifacts of the build you just created
   - Run the script with the name of the release tag and the path to the image: `./gcp-deploy release/vX.Y.Z.rc1 core-image-minimal-tdx-gcp.rootfs.wic.tar.gz`
   - Get the IPs of the TSS nodes (listed under `EXTERNAL_IP` in the output of the deploy script)
   - On one of them, get the TDX measurement value of this build from the output of `curl <ip address>:3001/version`
   - For each of them, get the TSS public keys from the output of `curl <ip address>:3001/info`

PR #1367, opened 2025-03-24
Build devnet chainspec script takes a chainspec given as an argument

ameba23:
This is a tiny, non-breaking change to `./scripts/build-devnet-chainspec.sh` script, which makes it take a chainspec type given as a command line argument, defaulting to `devnet` if none is given.

This is so the script can be used with the `tdx-testnet` chainspec.

For discussion of why we would use this, see https://github.com/entropyxyz/entropy-core/pull/1366#discussion_r2006618363

### DIFF
--- a/scripts/build-devnet-chainspec.sh
+++ b/scripts/build-devnet-chainspec.sh
@@ -1,2 +1,3 @@
 # This script is meant to be run once before launching the devnet to build the inital chain spec
-./target/release/entropy build-spec --disable-default-bootnode --raw --chain=devnet > chains/devnet.json
+CHAINSPEC=${1:-"devnet"}
+ENTROPY_TESTNET_TSS_IP=0.0.0.0:3001 ./target/release/entropy build-spec --disable-default-bootnode --raw --chain=$CHAINSPEC > chains/${CHAINSPEC}.json


scripts/build-devnet-chainspec.sh
```diff
@@ -1,2 +1,3 @@
 # This script is meant to be run once before launching the devnet to build the inital chain spec
-./target/release/entropy build-spec --disable-default-bootnode --raw --chain=devnet > chains/devnet.json
+CHAINSPEC=${1:-"devnet"}
+ENTROPY_TESTNET_TSS_IP=0.0.0.0:3001 ./target/release/entropy build-spec --disable-default-bootnode --raw --chain=$CHAINSPEC > chains/${CHAINSPEC}.json
```
ameba23, line 3: I found that if i don't set `ENTROPY_TESTNET_TSS_IP` to something i get:

```
thread 'main' panicked at node/cli/src/chain_spec/tdx_testnet.rs:38:10:
ENTROPY_TESTNET_TSS_IP environment variable to be set: NotPresent
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I dont think it matters what it is set to in this context.

HCastano: I would rather handle this here

https://github.com/entropyxyz/entropy-core/blob/4ac57ddf6e01ec4f8809b8f8eae91424c70a2e17/node/cli/src/chain_spec/tdx_testnet.rs#L38

by falling back to some well known IP (like the one above).

HCastano, line 3: Also, depending on which step of the release process you want to use this in, you should note that the `--raw` chainspec isn't easily patchable (it's a hex-file).

So if you're using this as a base you'll need to remove that.

HCastano (approved):

scripts/build-devnet-chainspec.sh
ameba23: Ah thanks.

Hmm at this point im wondering if it actually makes sense to use this script or start fresh.

As this could potentially automate the process of calling `/info` on the TSS endpoints and patching the JSON file using those outputs.

ameba23:
Im closing this in favor of https://github.com/entropyxyz/entropy-core/pull/1370

If we end up deciding we don't like that workflow we can reopen this

Issue #1199, opened 2024-12-05
Possible improvements to entropy-tss test environment


ameba23:
This issue bundles together 3 possible improvements to the entropy-tss testing setup following https://github.com/entropyxyz/entropy-core/pull/1147 which could be done as follow ups:

- [ ] Use a struct as a return value from the `spawn_tss_nodes_and_start_chain` function - as it currently returns a tuple with 5 fields - suggested by @HCastano : https://github.com/entropyxyz/entropy-core/pull/1147#discussion_r1867266898
- [ ] Consider using feature flags to allow enabling the entropy-tss unsafe API when calling `spawn_tss_nodes_and_start_chain` from `entropy-testing-utils`, as @dvdplm suggested:  https://github.com/entropyxyz/entropy-core/pull/1147/files/18c91cc6e3fa81888bc08eefb1676c6a956afa27#r1864869983
- [ ] Consider putting back the development chainspec as a variant of `ChainSpecType` to allow testing using the development chainspec.  I removed it because we currently don't use it. See comment: https://github.com/entropyxyz/entropy-core/pull/1147#discussion_r1869788741

PR #1368, opened 2025-03-24
Create return type for spawn tss nodes

JesseAbram:
Related #1199

### DIFF
--- a/crates/threshold-signature-server/src/backup_provider/tests.rs
+++ b/crates/threshold-signature-server/src/backup_provider/tests.rs
@@ -29,7 +29,7 @@ use entropy_kvdb::clean_tests;
 use entropy_shared::user::ValidatorInfo;
 use entropy_testing_utils::{
     constants::{TSS_ACCOUNTS, X25519_PUBLIC_KEYS},
-    helpers::{spawn_tss_nodes_and_start_chain, TssTestingResult},
+    helpers::spawn_tss_nodes_and_start_chain,
     ChainSpecType,
 };
 use serial_test::serial;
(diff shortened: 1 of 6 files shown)


crates/threshold-signature-server/src/backup_provider/tests.rs
```diff
@@ -66,8 +71,13 @@ async fn backup_provider_unit_test() {
...
 
-    let (_ctx, _api, _rpc, _validator_ips, _validator_ids) =
-        spawn_tss_nodes_and_start_chain(ChainSpecType::IntegrationJumpStarted).await;
+    let TssTestingResult {
+        substrate_context: _ctx,
+        api: _api,
+        rpc: _rpc,
+        validator_ips: _validator_ips,
+        validator_ids: _validator_ids,
+    } = spawn_tss_nodes_and_start_chain(ChainSpecType::IntegrationJumpStarted).await;
```
HCastano, line 80: We shouldn't destructure this whole thing if we're not going to use any of the fields.

I'd recommend we do something like: 

```suggestion
let _spawn_result = spawn_tss_nodes_and_start_chain(ChainSpecType::IntegrationJumpStarted).await;
```

crates/threshold-signature-server/src/backup_provider/tests.rs
```diff
@@ -41,8 +41,13 @@ async fn backup_provider_test() {
...
 
-    let (_ctx, api, rpc, _validator_ips, _validator_ids) =
-        spawn_tss_nodes_and_start_chain(ChainSpecType::IntegrationJumpStarted).await;
+    let TssTestingResult {
+        substrate_context: _ctx,
+        api,
+        rpc,
+        validator_ips: _validator_ips,
+        validator_ids: _validator_ids,
+    } = spawn_tss_nodes_and_start_chain(ChainSpecType::IntegrationJumpStarted).await;
```
HCastano, line 50: Here it's also a bit strange to me to destructure and ignore most fields. We can use the two we need directly, e.g `spawn_result.api` or `spawn_result.rpc`.

HCastano (commented):
I would try and avoid destructuring the result each time, especially since we have a lot of ignored fields.

We can either use them directly, or maybe break this down into two structs. One that focuses on just the connections (API + RPC) and the other which focuses on the TSS handles.

ameba23 (approved):
:star:

Approving because i think its already an improvement, but i agree with @HCastano that since most of the time we only need api and rpc, its neater not to destructure.

It is kinda substrate's fault that we don't just have one struct for interacting with the chain api. We could make a struct that wraps api and rpc, but that would be a bigger PR.

ameba23 (approved):
:100:

PR #1370, opened 2025-03-25
Build `tdx-testnet` chainspec using TSS details from running genesis TSS nodes

ameba23:
This builds the TDX testnet chainspec given the IPs of the running genesis TSS nodes, and dynamically builds the chainspec by hitting their `/info` endpoints to get their public keys.

This would not be used in production (i think ideally we should not have TSS nodes running at genesis in production but thats [another discussion](https://github.com/entropyxyz/entropy-core/issues/1353)).

But for setting up a testnet it simplifies things a lot as we don't need to manually add these details into to a JSON file during the release process.  We just need to know their IPs (and hopefully we can get these programmatically when deploying them).

@entropyxyz/system-reliability-engineers


node/cli/src/chain_spec/tdx_testnet.rs
```diff
@@ -87,8 +77,19 @@ pub fn tdx_testnet_config() -> ChainSpec {
...
-            tdx_devnet_four_node_initial_tss_servers(),
-            Some(vec![BoundedVec::try_from(ACCEPTED_MEASUREMENT.to_vec()).unwrap()]),
+            tdx_devnet_four_node_initial_tss_servers(tss_endpoints),
+            Some(measurement_values),
         ))
         .build()
 }
+
+/// Get the measurement value for the currently deployed TSS nodes
+fn get_measurement_values(tss_endpoints: &[String; 4]) -> MeasurementValues {
```
ameba23, line 87: This might not be needed when the measurement values are built during the image build pipeline - see https://github.com/entropyxyz/yocto-build/issues/13

node/cli/src/command.rs
```diff
@@ -76,6 +76,16 @@ impl SubstrateCli for Cli {
     // | testnet           | Four nodes, Two threshold servers, Own Seed, Testnet Configuration |
     // | tdx-testnet       | Four nodes, Four threshold servers, Alice Bob Chalie and Dave, Development Configuration adapted for TDX testnet |
     fn load_spec(&self, id: &str) -> Result<Box<dyn sc_service::ChainSpec>, String> {
+        #[cfg(feature = "production")]
+        if id == "tdx-testnet" {
+            return Ok(Box::new(chain_spec::tdx_testnet::tdx_testnet_config([
+                "127.0.0.1:3001".to_string(),
```
ameba23, line 82: These will be read in from a file, environment var or command line arg

ameba23:
Following our meeting, i am probably gonna close this PR in favor of moving the chainspec-modifying into the ansible playbook where there is already some chainspec-modifying going on: https://github.com/entropyxyz/ansible-collection-entropy/blob/dev/playbooks/generate-chain-spec.yaml

ameba23:
I am closing this in favor of https://github.com/entropyxyz/ansible-collection-entropy/issues/37

But may re-open if that turns out to be complicated or doesn't get done for whatever reason.

Issue #1376, opened 2025-03-31
`entropy` fails to compile with the `runtime-benchmarks` feature present


ameba23:
When i do `cargo build -p entropy --release --features runtime-benchmarks` 

i get: 
```
error[E0275]: overflow evaluating the requirement `pallet_babe::pallet::Call<entropy_runtime::Runtime>: std::marker::Send`
     |
note: required because it appears within the type `entropy_runtime::RuntimeCall`
    --> /home/pumkin/src/entropy/entropy-core/runtime/src/lib.rs:1520:1
     |
1520 | / construct_runtime!(
1521 | |   pub enum Runtime
1522 | |   {
1523 | |     System: frame_system = 1,
...    |
1575 | | );
     | |_^
note: required because it appears within the type `UncheckedExtrinsic<MultiAddress<AccountId32, u32>, RuntimeCall, MultiSignature, ...>`
    --> /home/pumkin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/sp-runtime-32.0.0/src/generic/unchecked_extrinsic.rs:69:12
     |
69   | pub struct UncheckedExtrinsic<Address, Call, Signature, Extra>
     |            ^^^^^^^^^^^^^^^^^^
     = note: required for `UncheckedExtrinsic<MultiAddress<AccountId32, u32>, RuntimeCall, MultiSignature, ...>` to implement `Member`
     = note: required for `Block<Header<u32, BlakeTwo256>, UncheckedExtrinsic<MultiAddress<..., ...>, ..., ..., ...>>` to implement `BlockT`
     = note: the full name for the type has been written to '/home/pumkin/src/entropy/entropy-core/target/release/deps/entropy-95e888898a5e8da0.long-type-12860491654769157762.txt'
     = note: consider using `--verbose` to print the full type name to the console
     = note: this error originates in the macro `frame_support::construct_runtime` (in Nightly builds, run with -Z macro-backtrace for more info)

For more information about this error, try `rustc --explain E0275`.
error: could not compile `entropy` (bin "entropy") due to 1 previous error
```

with `rustc 1.85.1 (4eb161250 2025-03-15)`

ameba23:
I have a hunch that this follows updating rust.  I tried to compile with rust 1.84 but i get:

```
error: failed to parse manifest at `/home/pumkin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/hex-literal-1.0.0/Cargo.toml`

Caused by:
[...]

PR #1378, opened 2025-04-01
Bump various substrate crates relating to the `runtime-benchmarks` feature

ameba23:
This tries to address https://github.com/entropyxyz/entropy-core/issues/1376

I started doing this an realized it is going to turn into a full substrate update dependency hell. I don't think i am going to see it though.

The problem is we can't roll back to rust 1.84 to run our benchmarks as we already have dependencies which rely on 2024 edition (1.85).

This is the downside of having dependabot eagerly bumping things all the time. We could have avoided this problem by checking in CI that entropy will compile with the `runtime-benchmarks` feature.


ameba23:
Closing in favor of https://github.com/entropyxyz/entropy-core/pull/1381

PR #1380, opened 2025-04-01
Repeatedly attempt to retrieve TSS db backup and on failure start fresh 

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1379

When requesting both the nonce and the encryption key, we use the `backoff` crate to repeatedly send requests to the backup provider (another TSS node) for 15 mins. This is for the event that they themselves are still getting setup.

On failure, we copy the current db directory to `<directory name>-backup-<timestamp>` and start fresh.

PR #1385, opened 2025-04-03
Repeatedly make encryption key backup attempts when starting `entropy-tss`

ameba23:
In https://github.com/entropyxyz/entropy-core/pull/1380 made it so that we repeatedly make attempts to **recover** a backup of an encryption key.

But we found we also need to do that when **making** a backup - otherwise in some cases the 'backup provider' is being asked to make a backup before they are fully setup (connected to chain etc) themselves. 

See this comment for logs showing this happening:

https://github.com/entropyxyz/entropy-core/pull/1380#issuecomment-2772159965

This PR uses the `backoff` crate to repeatedly request backups for 15 minutes.


ameba23:
> looks good but do you think we can get a regression test?

I've added one which demonstrates that the request will go through even when the backup provider is offline at the point of making the first attempt.

I am not gonna add a negative test to show we give up after 15min, as that would add 15min to our CI pipeline.

JesseAbram (approved):

PR #1402, opened 2025-04-17
Update benchmarks to V2

JesseAbram:


pallets/parameters/src/benchmarking.rs
```diff
@@ -30,64 +29,67 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
...
+#[benchmarks]
+mod benchmarks {
+    use super::*;
+
+    #[benchmark]
+    fn change_request_limit() {
+        let origin = T::UpdateOrigin::try_successful_origin().unwrap();
+
+        #[extrinsic_call]
+        _(origin as T::RuntimeOrigin, 15.clone());
```
ameba23, line 41: im confused as to why we don't need to specify that its the `change_request_limit` extrinsic here (and with the other ones) - whereas we did before.  Like how does it know which one we are calling?

ameba23: ah, i checked the docs and i get it - the `_` is replaced with the name of the function - which matches the name of the extrinsic

pallets/programs/src/benchmarking.rs
```diff
@@ -39,104 +39,117 @@ fn assert_last_event<T: Config>(generic_event: <T as Config>::RuntimeEvent) {
...
-      BoundedVec::try_from(oracle_data_pointers[0].clone()).unwrap(),
-      BoundedVec::default(),
-    );
-
-    let (_oracle_length, hash_input_with_oracle) =
-    ProgramsPallet::<T>::get_length_and_hash_of_oracle(&oracle_data_pointers, hash_input).unwrap();
-
-    let program_hash = T::Hashing::hash(&hash_input_with_oracle);
-    let deployer: T::AccountId = whitelisted_caller();
-    let sig_req_account: T::AccountId = whitelisted_caller();
```
ameba23, line 67: Was this just never used before?  How come we didn't get a compiler warning?

ameba23 (approved):
:1st_place_medal: great to have this in - looks like it was a lot of work!

pallets/parameters/src/benchmarking.rs
JesseAbram: I was asleep but yes this exactly

pallets/programs/src/benchmarking.rs
JesseAbram: good question 95% positive that since it was in a macro cargo lint and fmt were ignored now that it is attributes it is picked up

JesseAbram: https://www.reddit.com/r/rust/comments/1eh7g80/why_does_rustfmt_ignore_macros_when_formatting/

PR #1409, opened 2025-04-25
Fix non local definitions warning

JesseAbram:
Closes #1399


ameba23 (approved):
:+1: works for me.

I haven't actually checked that the warnings dont show up any more, but i expect you have.

PR #1410, opened 2025-04-30
Fix CI check for code changed (suggested by Zach)

ameba23:
In some cases tests were not being run despite code having been changed, due to an issue with the check as to whether code changes. Eg: https://github.com/entropyxyz/entropy-core/actions/runs/14751508488/job/41409844605?pr=1392

@cooldracula suggested this fix - making it so that all commit history is checked out, so `git diff` can find the commits we are referring to.

### DIFF
--- a/.github/workflows/build-and-run-node-test.yaml
+++ b/.github/workflows/build-and-run-node-test.yaml
@@ -15,7 +15,11 @@ jobs:
       - name: Check for relevant changes
         id: filter
         run: |
-          if git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep -E "^(node/|crates/|pallets/|runtime/)"; then
+          ## If this is the first commit on a branch, then the default behaviour of github.event.before is to show the previous ref as
+          ## repeated zeroes, as the previous ref is null
+          if [ "${{ github.event.before }}" = "0000000000000000000000000000000000000000" ]; then
+            echo "changes_detected=true" >> $GITHUB_ENV
+          elif git diff --name-only ${{ github.event.before }} ${{ github.sha }} | grep -E "^(node/|crates/|pallets/|runtime/)"; then
             echo "changes_detected=true" >> $GITHUB_ENV
           else
             echo "changes_detected=false" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


cooldracula (approved):

cooldracula:
With the additional code added to check if this is the first commit on a branch, I think this looks good!

ameba23:
> With the additional code added to check if this is the first commit on a branch, I think this looks good!

That great - on my branch the check evaluates to true when i pushed a trivial code change. But for some reason the tests are still being skipped.  I feel like i must be missing something really stupid.

https://github.com/entropyxyz/entropy-core/actions/runs/14771602673/job/41472444384?pr=1392

cooldracula:
> That great - on my branch the check evaluates to true when i pushed a trivial code change. But for some reason the tests are still being skipped. I feel like i must be missing something really stupid.

The only thing I can think of, at a glance, is that the steps feel a little roundabout.  The code checkout echoes to $GITHUB_ENV which is then echoed to $GITHUB_OUTPUT which is then checked in the next step.  This could cause the value to be different than what we expect.  Could we just echo directly to $GITHUB_OUTPUT in the `code-paths-changed` job?   I'll push a commit to this branch with what I mean.

cooldracula:
Ah, looking at it closer, I think I see the issue with the other job.  My commit above will fix it, indirectly, and so I still want to explain it here.

On [line 10 of the workflow file](https://github.com/entropyxyz/entropy-core/actions/runs/14771602673/workflow?pr=1392#L10), you set the output as:

```yaml
  outputs:
      run_tests: ${{ steps.filter.outputs.changes_detected }}
 ```
 And [the steps](https://github.com/entropyxyz/entropy-core/actions/runs/14771602673/workflow?pr=1392#L11-L28) are then:
 ```yaml
       - name: Check for relevant changes
        id: filter
        run: |
          ## echo values to $GITHUB_ENV
      - name: Set output
        run: # echo env values to $GITHUB_OUTPUT
```
The outputs definition uses the outputs of the step with the ID `filter`, but that step isn't outputting anything.  Instead, the output is changed in the subsequent step, which has no ID.  

So another solution would be to set the `filter` id on the last step, but I don't think that's ideal as ultimately this step is redundant and should be removed. It makes more sense, to me, to have our git diff check echo directly to $GITHUB_OUTPUT and avoid the $GITHUB_ENV middle man.

ameba23:
Thats great - using this correctly triggered the tests to run. Yeah i see now that that code was pretty stupid. But whats strange is that it's been like that for several weeks and no-one including myself has complained about it until now. 

Thanks so much @cooldracula

PR #1411, opened 2025-04-30
Remove deadlines in OCW

### LINKED ISSUE #1323: Remove all OCW timeouts to
Currently there are deadlines on the OCW side, but we have ran into issues there, and because we control the TSS code, the timeouts should be handled within the TSS.

Opening an issue to discuss before fixing

JesseAbram:
Closes #1323 

Will update issue with a point of how it was done (instead of here to keep a longer history of the action)


ameba23 (approved):
My gut feeling is that it might be better to have a really long timeout than no timeout. But i think this fine - if we have timeouts on the entropy-tss side, and the worst case scenario is only that the other OCW requests don't end up going through.

ameba23:
I think this should go in the changelog

PR #1413, opened 2025-05-01
Use ss58 for loggings

### LINKED ISSUE #1271: Use SS58 AccountIDs in logs?
As mentioned in https://github.com/entropyxyz/entropy-core/pull/1228#discussion_r1933135257, we currently
log TSS account IDs (and probably other account IDs) as byte arrays. It would be nicer if
we could log them as SS58 formatted addresses.

The main thing here would be to ensure consistency in our logs, ensuring that we don't
mix up both byte-arrays and SS58 addresses needlessly. Otherwise this will make debugging
issues harder since we'll constantly be converting between them on the side.

JesseAbram:
Closes #1271 

@ameba23 is this enough you think or should I go through the whole code base and do this?

### DIFF
--- a/crates/client/src/user.rs
+++ b/crates/client/src/user.rs
@@ -19,7 +19,7 @@ use crate::{
 };
 use entropy_shared::{user::ValidatorInfo, BlockNumber, HashingAlgorithm};
 use serde::{Deserialize, Serialize};
-use sp_core::{sr25519, Pair};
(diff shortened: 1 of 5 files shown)


ameba23 (approved):
> @ameba23 is this enough you think or should I go through the whole code base and do this?

I don't know how many other places we log account ids - but unless you have checked the whole codebase i would leave the issue open.

ameba23 (approved):
:100:

PR #1415, opened 2025-05-02
FIx test_node_process_stationary_local

JesseAbram:
@ameba23 Found a race condition in this test here https://github.com/entropyxyz/entropy-core/actions/runs/14791179943/job/41528489582

The issue is the ```.staking_extension().proactive_refresh()``` gets cleared at the end of each block. It gets set at block 0 with default so it has a result. Then when it gets cleared it causes an error. This is good behaviour normally, as this only will get fired once by the chain, but bad for this test

I set it to a chain that won't produce blocks so the test doesn't have to race to make this call


ameba23:
Apologies for pushing to your branch - but i wanna to see the tests run.  Since this issue was only appearing sometimes anyway it will be hard to know.  But i expect that it has fixed it.

ameba23 (approved):
:100:

PR #1418, opened 2025-05-05
Add outtie pallet

JesseAbram:
* Creates an Outtie pallet to hold API boxes as per outtie spec 
* Leaves in todos about adding attestations as that will be added after


ameba23 (approved):
:1st_place_medal: Great to have this so fast.

I made an issue for adding the the attestation logic, and also for naming - as we are being inconsistent in places and the longer we leave it the more of a PITA it is gonna be to get right. https://github.com/entropyxyz/entropy-core/issues/1420

Issue #1407, opened 2025-04-24
Improve handling of custom inputs to chainspec when deploying


ameba23:
There have been issues with updating [`generate-chain-spec` playbook](https://github.com/entropyxyz/ansible-collection-entropy/blob/dev/playbooks/generate-chain-spec.yaml) when the chainspec data structure has changed following a substrate upgrade or other breaking changes in a release of entropy-core. 

Currently we build the chainspec from custom inputs using [functions in the entropy node CLI](https://github.com/entropyxyz/entropy-core/blob/d2e5da6e0d972d95a0f824786a244fcbf7cf1b4c/node/cli/src/chain_spec/testnet.rs#L160), release it as JSON, modify that JSON using `jq` in the deployment playbooks, and load the modified JSON when starting the chain node, using the `--chain` command line option.

These custom inputs include TSS details (stash account, public keys, hostname, provisioning certification key), pre-funded account balances, bootnode peer IDs, and (soon) entropy-tss CVM measurement values.

I propose switching from modifying the chainspec JSON using `jq` to **either**:
1. Creating a custom CLI command for `entropy` which builds a JSON chainspec given a smaller JSON object containing only the custom inputs.
2. Giving these inputs as command line arguments when starting the `entropy`, and so never needing to handle the chainspec as JSON.

Both options have the advantage that the chainspec-building logic can be type-checked and tested as part of `entropy-core`, and so need to be maintained on a per-PR rather that per-release basis.

Both options however do not completely solve the problem - we still need to take care to document changes to the structure of the custom inputs and effectively communicate these to the @entropyxyz/system-reliability-engineers team.

PR #1424, opened 2025-05-06
Add custom testnet chainspec inputs

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1407

This allows passing the path to a file containing a JSON object with custom inputs to build a chainspec for a test network when starting the chain node.

This is passed using the `--chain` command line argument with a filename which ends with `chainspec-inputs.json` - for example: `entropy --chain ./path/to/testnet-v0.4.0-chainspec-inputs.json`.  This file should contain a JSON serialized instance of the `TestnetChainspecInputs` struct:

https://github.com/entropyxyz/entropy-core/blob/19f34eaaadb48ad504e264c9ec91581237583be8/node/cli/src/chain_spec/testnet.rs#L57-L79

This struct includes:
- TSS details (public keys etc)
- Accepted TDX measurement values (optional) - if not given will assume non-TDX deployment and accept mock values. 
- Bootnode peer IDs
- Endowed accounts - this is debatable as we do already have some code for reading endowed accounts from a file - but AFAICT this is not what we currently use in the deployment playbooks.

For an example of what the JSON should look like see the test object used in CI to check this works:

https://github.com/entropyxyz/entropy-core/blob/peg/custom-testnet-chainspec-inputs/node/cli/test-chainspec-inputs/example-chainspec-inputs.json

TODO:
- [x] Error handling in the chainspec generation fn
- [x] Endowed accounts
- [ ] Consider whether to retrieve the TSS details from within the function which builds the chainspec - if so i would do it in a followup.  The idea is that you would only need to give the hostnames or IP addresses of TSS servers - and hitting `/info` to get the remaining details would be done internally by `entropy`.
- [ ] Generate JSON schema for the custom inputs struct - i had issues with doing this using the `schemars` derive macro as we have a bunch of foreign sub-types. So probably we need to manually implement the `JsonSchema` trait which will be a fair bit of code.  I would make an issue and do it in a followup. 
- [ ] Document the process of using this - either in the readme of this repo, or the ansible collection repo, or the wiki?


node/cli/src/endowed_accounts.rs
```diff
@@ -16,43 +16,15 @@
...
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct AddressStruct {
     address: String,
     name: String,
 }
 
-/// These are accounts which are populated from an external source, with the intention of them
-/// being funded an ready to use in a `testnet` configuration.
-pub fn endowed_testnet_accounts() -> Vec<AccountId> {
```
ameba23, line 32: This duplicates reading these from a file with other options in as well.  Apparently we also were not using it in the ansible deployment scripts.

cooldracula: We do use an endowed account present on the chianspec as part of our [jumpstart playbook](https://github.com/entropyxyz/ansible-collection-entropy/blob/f35bb7f77d601538b446edf5b9bcc6c66fd83be8/playbooks/jumpstart.yaml#L54), but I may be misunderstanding this change?   I am happy to update that playbook if we were not using available features that make the steps simpler.

node/cli/src/command.rs
```diff
@@ -68,14 +68,13 @@ impl SubstrateCli for Cli {
...
-    // | testnet-local     | Two Nodes, Two threshold servers, Alice and Bob, Testnet Configuration, Docker Compatible |
-    // | testnet           | Four nodes, Two threshold servers, Own Seed, Testnet Configuration |
-    // | tdx-testnet       | Four nodes, Four threshold servers, Alice Bob Chalie and Dave, Development Configuration adapted for TDX testnet |
+    // | --chain                   | Description |
+    // |-------------------------  |----------- |
+    // | dev                       | Four nodes, Four threshold servers, Alice Bob and Charlie and Dave, Development Configuration |
+    // | devnet-local              | Four nodes, four threshold servers, Alice Bob Charlie and Dave, Development Configuration, Docker Compatible |
+    // | integration-tests         | Two nodes, Four threshold servers, Alice and Bob, Development Configuration |
+    // | testnet-local             | Two Nodes, Two threshold servers, Alice and Bob, Testnet Configuration, Docker Compatible |
+    // | testnet-<my-config>.json  | Custom Testnet Configuration with given inputs from a specified JSON file |
```
cooldracula, line 77: Is there a difference between `testnet-<my-config>.json` and `<my-testnet>-chain-inputs.json`?  Does the `--chain` flag basically accept either: 
- one of the four pre-configured setups
- a custom setup provided by a json file
Or is `testnet-<my-config>.json` referring to something else?

ameba23: Ah good that you flagged this up - this need updating.  This is was i had it as originally and then i decided to change it and didn't update the doccomments.

node/cli/src/endowed_accounts.rs
ameba23: This function reads the endowed accounts list from the path `./data/testnet/testnet-accounts.json` if the file exists.  It is run on starting a testnet chain. As far as i understand, in the playbooks you are not creating this file. Instead you are patching the endowed accounts into the chainspec directly: https://github.com/entropyxyz/ansible-collection-entropy/blob/f35bb7f77d601538b446edf5b9bcc6c66fd83be8/playbooks/generate-chain-spec.yaml#L165

HCastano: For a bit of context here, we were using this to pre-fund testnet accounts for Entropy team members. That means that the chainspec that SRE is patching should only be _adding_ new endowed accounts, not replacing these which have already been included by us.

node/cli/src/command.rs
```diff
@@ -68,14 +68,13 @@ impl SubstrateCli for Cli {
...
-    // | testnet-local     | Two Nodes, Two threshold servers, Alice and Bob, Testnet Configuration, Docker Compatible |
-    // | testnet           | Four nodes, Two threshold servers, Own Seed, Testnet Configuration |
-    // | tdx-testnet       | Four nodes, Four threshold servers, Alice Bob Chalie and Dave, Development Configuration adapted for TDX testnet |
+    // | --chain                            | Description |
+    // |----------------------------------  |----------- |
+    // | dev                                | Four nodes, Four threshold servers, Alice Bob and Charlie and Dave, Development Configuration |
+    // | devnet-local                       | Four nodes, four threshold servers, Alice Bob Charlie and Dave, Development Configuration, Docker Compatible |
+    // | integration-tests                  | Two nodes, Four threshold servers, Alice and Bob, Development Configuration |
+    // | testnet-local                      | Two Nodes, Two threshold servers, Alice and Bob, Testnet Configuration, Docker Compatible |
+    // | <my-testnet>chainspec-inputs.json  | Custom Testnet Configuration with given inputs from a specified JSON file |
```
HCastano, line 77:
```suggestion
    // | <my-testnet>-chainspec-inputs.json  | Custom Testnet Configuration with given inputs from a specified JSON file |
```

nit: but would be good to establish a little bit of a convention here

node/cli/src/chain_spec/testnet.rs
```diff
@@ -50,6 +52,40 @@ type TssAccountId = sp_runtime::AccountId32;
 /// The format should be in the form of `scheme://hostname:port`.
 type TssEndpoint = String;
 
+/// Custom input data for building the chainspec for a particular test network
+#[derive(Serialize, Deserialize, Clone, Debug)]
```
HCastano, line 56: Would be really nice to have the JsonSchema available here as you mention in the issue body. This would also help alleviate a lot of Vi's concerns.

Fine to do it in a follow up though.

.github/workflows/check-chainspecs.yaml
```diff
@@ -14,4 +14,4 @@ jobs:
         run: |
           cargo run -p entropy -- build-spec --raw --chain dev > chainspec-dev-raw.json
           cargo run -p entropy -- build-spec --raw --chain integration-tests > chainspec-integration-raw.json
-          cargo run -p entropy -- build-spec --raw --chain testnet > chainspec-testnet-raw.json
+          cargo run -p entropy -- build-spec --raw --chain ./node/cli/test-chainspec-inputs/example-chainspec-inputs.json > chainspec-testnet-raw.json
```
HCastano, line 17: 👍

HCastano (approved):
I like this approach. It tightens up the scope of information that DevOps needs to worry about when dealing with the chainspec.

For your open questions:

> Consider whether to retrieve the TSS details from within the function which builds the chainspec - if so i would do it in a followup. The idea is that you would only need to give the hostnames or IP addresses of TSS servers - and hitting /info to get the remaining details would be done internally by entropy.

My gut says that we shouldn't be "hiding" this from DevOps, but it does sound convenient. Maybe I'll defer to DevOps here, but I'm fine either way.

> Generate JSON schema for the custom inputs struct - i had issues with doing this using the schemars derive macro as we have a bunch of foreign sub-types. So probably we need to manually implement the JsonSchema trait which will be a fair bit of code. I would make an issue and do it in a followup.

I addressed this elsewhere, totally fine to do in a follow-up.

> Document the process of using this - either in the readme of this repo, or the ansible collection repo, or the wiki?

It should be in this repo, and probably linked to by the Ansible step where the patching is taking place.

Issue #1427, opened 2025-05-08
Reading keyshare from disk makes `entropy-tss` vulnerable to rollback attacks from host operator


ameba23:
This possible vulnerability was brought up by @dvdplm in a conversation on slack.

The host operator of is able to read and write the disk volume the `entropy-tss` CVM uses.  We protect the CVM by keeping almost everything in a measured ramdisk.  Except for things which require persistence accross reboots.

Everything we store in persistent storage is encrypted using authenticated encryption, so we have some integrity protection. But @dvdplm points out this is still vulnerable to rollback attacks where the attacker replaces the storage with a previous valid state.

I think for this reason we should avoid reading from storage unless we have no choice.  The only time we have no choice is following a reboot.

Normally, when we read the keyshare from the kvdb, it will read a cached copy from the last write from memory. But this relies on the internal behavior of [`sled`](https://docs.rs/sled) and is not guaranteed to be the case.  For this reason i propose to keep the keyshare in memory as part of application state.  When mutating it, we should also write the latest state to the kvdb, but reading it should always come from memory only.

There may also be some integrity checks we can do when restoring backups to check whether it appears to be the most recent state.

ameba23:
> I am not sure that this issue should be considered resolved by https://github.com/entropyxyz/entropy-core/pull/1429?

Yes i completely agree - gotta be more careful with wording - i wrote:

> This relates to, but does not completely fix https://github.com/entropyxyz/entropy-core/issues/1427

which github took to mean close the issue on merging the PR.

PR #1429, opened 2025-05-09
Only read from the `entropy-tss` key-value store following a reboot 

ameba23:
This relates to, but does not completely fix https://github.com/entropyxyz/entropy-core/issues/1427

In this PR, when reading a keyshare - we always read from the app state cache (memory).  When modifying a keyshare, we write to both the cache and the key-value store (disk).

This turned into quite a big PR because i have made the kv store a private field on app state - to restrict where it is used.  This meant changing the unsafe API to also read from cache rather than the KVDB.

TODO:
- [x] Run reshare e2e test


crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -273,6 +285,44 @@ impl Cache {
             ))
         })
     }
+
+    fn read_network_key_share(&self) -> Result<Option<KeyShareWithAuxInfo>, AppStateError> {
+        let key_share =
+            self.network_key_share.read().map_err(|e| AppStateError::PosionError(e.to_string()))?;
```
JesseAbram, line 291: If there is a panic when doing a read this would lock the TSS until a reset (since there is no clearing of the poisoned mutex). I would consider clearing the mutex since for a read here I would rather have one signing fail then have a whole signer lock and need to be rebooted

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -273,6 +285,44 @@ impl Cache {
...
+    }
+
+    fn write_network_key_share(
+        &self,
+        updated_key_share: Option<KeyShareWithAuxInfo>,
+    ) -> Result<(), AppStateError> {
+        let mut key_share = self
+            .network_key_share
+            .write()
+            .map_err(|e| AppStateError::PosionError(e.to_string()))?;
```
JesseAbram, line 302: similar idea here a write would be during a reshare I would rather have the chain try another reshare in the case something caused a locked mutex then locking the TSS until a manual restart

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -273,6 +285,44 @@ impl Cache {
...
+        *key_share = updated_key_share;
+        Ok(())
+    }
+
+    fn read_next_network_key_share(&self) -> Result<Option<KeyShareWithAuxInfo>, AppStateError> {
+        let key_share = self
```
JesseAbram, line 308: similar concept on these

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -329,6 +400,76 @@ impl AppState {
...
+    /// Update the network key-share - this includes removing it when leaving the signer set
+    /// which is done by setting it to `None`
+    pub async fn update_network_key_share(
+        &self,
+        updated_key_share: Option<KeyShareWithAuxInfo>,
+    ) -> Result<(), AppStateError> {
+        self.cache.write_network_key_share(updated_key_share.clone())?;
+
+        // Write to on-disk store for backup
+        self.kv_store.kv().delete(&hex::encode(NETWORK_PARENT_KEY)).await.unwrap();
```
JesseAbram, line 424: unwraps() in the next two functions

JesseAbram: also should probably check if the key exists first then delete so there is no error if it doesn't exist (on the first write)

crates/threshold-signature-server/src/helpers/app_state.rs
```diff
@@ -329,6 +400,76 @@ impl AppState {
...
+    /// Update the 'next' network key-share - this includes removing it which is done by setting it
+    /// to `None`
+    pub async fn update_next_network_key_share(
+        &self,
+        updated_key_share: Option<KeyShareWithAuxInfo>,
+    ) -> Result<(), AppStateError> {
+        self.cache.write_next_network_key_share(updated_key_share.clone())?;
+
+        // Write to on-disk store for backup
+        self.kv_store.kv().delete(&hex::encode(NEXT_NETWORK_PARENT_KEY)).await.unwrap();
```
JesseAbram, line 444: same here

crates/threshold-signature-server/src/validator/api.rs
```diff
@@ -122,10 +121,7 @@ async fn do_reshare(
         if data.new_signers.contains(&my_stash_address.encode()) {
             None
         } else {
-            let kvdb_result = app_state.kv_store.kv().get(&hex::encode(NETWORK_PARENT_KEY)).await?;
-            let key_share: KeyShareWithAuxInfo =
-                entropy_kvdb::kv_manager::helpers::deserialize(&kvdb_result)
-                    .ok_or_else(|| ValidatorErr::KvDeserialize("Failed to load KeyShare".into()))?;
+            let key_share = app_state.network_key_share()?.unwrap();
```
JesseAbram, line 124: unwrap

crates/threshold-signature-server/src/validator/tests.rs
```diff
@@ -141,7 +142,7 @@ async fn test_reshare_basic() {
     }
     .unwrap();
 
-    // wait for roatate keyshare
+    // wait for rotate keyshare
```
JesseAbram, line 145: pretty good chance I made this type

crates/threshold-signature-server/src/helpers/app_state.rs
ameba23: The kvdb delete handler makes this check internally:

https://github.com/entropyxyz/entropy-core/blob/6843f688135083d43bb5907682bceb34e42fb480/crates/kvdb/src/kv_manager/sled_bindings.rs#L52-L54

crates/threshold-signature-server/src/helpers/app_state.rs
ameba23: Having looked into it - totally agree

crates/threshold-signature-server/src/validator/api.rs
ameba23: I have turned this into an error, making the reshare fail if this is the case.  This feels like the best thing to do as it shows there is a mismatch between the state of the signing committee on-chain and who actually holds keyshares.  But there are probably other possible ways to handle it.

PR #1436, opened 2025-05-16
Downgrade rust version to 1.86 as a temporary measure until we upgrade sp-io

### LINKED ISSUE #1435: Runtime error when starting `entropy` chain node - `runtime requires function imports which are not present on the host`
When starting the chain node i see the following error:

```
$ ./entropy --dev
2025-05-16 13:42:54 Entropy Node
2025-05-16 13:42:54 ✌️  version 0.3.0-unknown
2025-05-16 13:42:54 ❤️  by Entropy Cryptography <engineering@entropy.xyz>, 2022-2025
2025-05-16 13:42:54 📋 Chain specification: Development
2025-05-16 13:42:54 🏷  Node name: attractive-spot-4053
2025-05-16 13:42:54 👤 Role: AUTHORITY
2025-05-16 13:42:54 💾 Database: RocksDb at /tmp/pumkin/substrate0mC26j/chains/dev/db/full
2025-05-16 13:42:55 [0] 💸 generated 4 npos voters, 4 from validators and 0 nominators
2025-05-16 13:42:55 [0] 💸 generated 4 npos targets
Error: Service(Client(VersionInvalid("Other error happened while constructing the runtime: runtime requires function imports which are not present on the host: 'env:_RNvCscSpY9Juk0HT_7___rustc17rust_begin_unwind'")))
2025-05-16 13:42:56 Cannot create a runtime error=Other("runtime requires function imports which are not present on the host: 'env:_RNvCscSpY9Juk0HT_7___rustc17rust_begin_unwind'")
```

This is causing tests to fail on master - see https://github.com/entropyxyz/entropy-core/pull/1434

I'm not sure why this issue has only just popped up as happens on master as well as older commits. The oldest one i have tried is `d3660102c9009fa96e309532a2e48f428f031840` which was march 18th (a week or two before the `v0.4.0-rc.1` release).

This happens in CI as well as locally, and after when building entropy in docker locally.  And after a fresh clone.  The only thing which seems to stop the error is using an existing storage directory from a previously working chain state.

 I am pretty sure this is not code-related but something has changed in the development environment - as tests have been passing earlier this week and i was been able to pull chain metadata twice this week which i am not able to do now.

It looks this is related to the runtime being compiled expecting there to be some unwind on panic functionality - so like it is not being correctly built as no-std.  But why this has popped up from one day to the next baffles me completely.

ameba23:
See https://github.com/entropyxyz/entropy-core/issues/1435

### DIFF
--- a/.github/actions/install-dependencies/action.yaml
+++ b/.github/actions/install-dependencies/action.yaml
@@ -12,8 +12,8 @@ runs:
       shell: bash
     - name: Add Rust components
       run: |
-        rustup install 1.78.0
-        rustup default 1.78.0
+        rustup install 1.86.0
+        rustup default 1.86.0
         rustup target add wasm32-unknown-unknown
         rustup component add rust-src
       shell: bash
--- a/.github/actions/install-dependencies/action.yaml
+++ b/.github/actions/install-dependencies/action.yaml
@@ -12,6 +12,7 @@ runs:
       shell: bash
     - name: Add Rust components
       run: |
+        # FIXME 1435 - 1.86 is only needed until we next update substrate - otherwise we use stable
         rustup install 1.86.0
         rustup default 1.86.0
         rustup target add wasm32-unknown-unknown
--- a/.github/actions/install-dependencies/action.yaml
+++ b/.github/actions/install-dependencies/action.yaml
@@ -12,6 +12,8 @@ runs:
       shell: bash
     - name: Add Rust components
       run: |
+        rustup install 1.78.0
+        rustup default 1.78.0
         rustup target add wasm32-unknown-unknown
         rustup component add rust-src
       shell: bash
--- a/crates/threshold-signature-server/src/lib.rs
+++ b/crates/threshold-signature-server/src/lib.rs
@@ -15,7 +15,7 @@
 
 //! # Server
 //!
-//! The Threshold Server which stores key shares and participates in the signing protocol.
+//! The Threshold Signature Server which stores key shares and participates in the signing protocol.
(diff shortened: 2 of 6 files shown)


ameba23:
It doesn't look like this made any difference - but locally it seemed to. :shrug:

crates/threshold-signature-server/src/lib.rs
```diff
@@ -13,9 +13,9 @@
 // You should have received a copy of the GNU Affero General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-//! # Server
+//! # Treshold Server
```
HCastano, line 16:
```suggestion
//! # Threshold Server
```

.github/actions/install-dependencies/action.yaml
```diff
@@ -12,6 +12,8 @@ runs:
       shell: bash
     - name: Add Rust components
       run: |
+        rustup install 1.86.0
```
HCastano, line 15: Let's add a `NOTE` or `FIXME` comment here pointing to the build issue so we have a reference as to why we're on this particular version

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel="stable"
+channel = "1.86.0"
```
HCastano, line 2: Same thing here

HCastano (approved):
CI needs a few fixes, but from the other build artifact you handed off to SRE it does look like using `1.86.0` fixes the build

PR #1439, opened 2025-05-21
Update substrate to polkadot stable2503-4

JesseAbram:


rust-toolchain.toml
```diff
@@ -1,4 +1,3 @@
 [toolchain]
-# FIXME 1435 - 1.86 is only needed until we next update substrate - otherwise we use stable
-channel="1.86.0"
+channel="stable"
```
ameba23, line 2: :partying_face: 

Great.  Can also remove these lines:

https://github.com/entropyxyz/entropy-core/blob/332796378a758b030c713dd44deadc1dc7b2d76b/.github/actions/install-dependencies/action.yaml#L15-L17

ameba23 (approved):
:1st_place_medal: :1st_place_medal: :1st_place_medal: 
Great - looks like this was really a lot of work.

Im curious what the `DoneSlashHandler` is about and if it has any implications for us.

JesseAbram:
> Im curious what the `DoneSlashHandler` is about and if it has any implications for us.

Really nothing for us but it allows for a callback after a slash, which is cool so if we wanted to do on slash do X we can easily

Issue #1441, opened 2025-05-22
Releasing on master not possible due to chainspec not being updated following PR #1424


ameba23:
This is my fault - following https://github.com/entropyxyz/entropy-core/pull/1424 i forgot to update the release workflow - this line needs updating:

https://github.com/entropyxyz/entropy-core/blob/87df32d041e07389d7516afbff33ad775cf4cf16/.github/workflows/container-build-and-push.yaml#L120

As now a default `testnet` chainspec does not exist - it can only be created from custom inputs.

PR #1442, opened 2025-05-22
Template testnet chainspec artifact

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1441

This adds a 'blank' testnet chainspec which can be built as JSON and then externally modified to produce a working chainspec. 

By 'blank', i mean the fields are set to default values. It cannot actually be used as working testnet configuration but can serve as a template.

This is temporary. I think probably once https://github.com/entropyxyz/ansible-collection-entropy/issues/40 is done we wont need the chainspec artifact anymore.


ameba23:
Test release to see if this works: https://github.com/entropyxyz/entropy-core/releases/tag/test%2Frelease%2Ftest-1424

Issue #1450, opened 2025-05-28
Failing CI tests relating to jumpstart and proactive refresh


ameba23:
Since some time, tests are failing intermittently.  This happens more than half of the time.

I say 'since some time' because it is difficult to find out exactly since when due to a bug with our code change checks which meant that for some time CI tests were being by-passed even when code changes had been made. https://github.com/entropyxyz/entropy-core/pull/1410

My initial hunch was that this relate to an update with synedrion: https://github.com/entropyxyz/entropy-core/pull/1392

But looking at the the commit history - that didn't get merged until after the code check bug was fixed, and this issue was already present: https://github.com/entropyxyz/entropy-core/actions/runs/14856044563/job/41709573488

The tests which often fail are:
- `user::tests::test_jumpstart_network`
- `user::tests::test_validate_jump_start_fail_repeated`
- `signing_client::tests::test_proactive_refresh_validation_fail`

We need to go through commit history re-running tests and figure out when this started happening.

TBH i'm not sure there has ever been a time when there have not been issues with either the jumpstart or reshare tests occasionally failing.

ameba23:
Pasting here the output @JesseAbram posted to slack from running the jumpstart test.

It appears the reshare protocol runs but the aux gen doesn't.  I remember having this issue before but i thought it was resolved, and i can't remember how.

```

  2025-05-27T20:18:48.026719Z  INFO entropy_tss::signing_client::protocol_transport: Got ws connection, with message: SubscribeMessage { session_id: Dkg { block_number: 9 }, public_key: 306bdb49cbbe7104e3621abab3c9d31698b159f48dafe567abb7ea5d872ed329 (5DACCJgQ...), signature: f8d8aa29fdbd80780657c0bb4bb6ba75317e5288bb457297c64a5fea0631fc2b3770fc8cf2dddc61fd24cc736ab1d005bf0fc43302eb00c646e647ff8235eb8f, version: 1 }
    at crates/threshold-signature-server/src/signing_client/protocol_transport.rs:195

[...]

PR #1452, opened 2025-05-28
Run DKG in a separate task to HTTP request handler for jumpstart

ameba23:
To see if it helps with https://github.com/entropyxyz/entropy-core/issues/1450

Unfortunately this does not seem to solve the  problem :cry:


ameba23:
Im closing this since it didn't solve the problem.

Issue #1444, opened 2025-05-26
Registry pallet - derivation path field on user info should not be optional


ameba23:
I came across this in https://github.com/entropyxyz/entropy-core/pull/1429 and it is a relic from the time when we wanted to support both the old (pre-tofino redesign) user accounts as well as new ones which use key derivation.

The `derivation_path` field on `RegisteredInfo` in the registry pallet should not be an `Option`.
 
https://github.com/entropyxyz/entropy-core/blob/87df32d041e07389d7516afbff33ad775cf4cf16/pallets/registry/src/lib.rs#L121

Having it as an `Option` forces us to handle the case that it does not exist in the TSS code, which should be an impossible state to get into.

PR #1455, opened 2025-05-29
Remove option for derivation path

JesseAbram:
#1444


pallets/registry/src/lib.rs
```diff
@@ -127,7 +127,7 @@ pub mod pallet {
         pub programs_data: BoundedVec<ProgramInstance<T>, T::MaxProgramHashes>,
         pub program_modification_account: T::AccountId,
         /// The SCALE encoded BIP-32 `DerivationPath` used to register this account.
-        pub derivation_path: Option<Vec<u8>>,
+        pub derivation_path: Vec<u8>,
```
HCastano, line 130: I think this requires a `spec_version` bump. Do you mind adding that label to the PR?

PR #1459, opened 2025-05-30
Fix staking it_deletes_when_no_bond_left test

JesseAbram:
This test is failing with the new update (weird it wasn't picked up in the CI, anyways instead of dealing with locks we check free balance now


ameba23 (approved):
:star:

PR #1469, opened 2025-06-05
Deterministic alice bootnode for testing

JesseAbram:
Idk why but my nodes were not able to find themselves, not sure how it happened and why not on CI, however now I create alice's node key deterministically so it is hardcoded in our tests and nodes can find each other, this fixes the issue and in general makes this more stable

### DIFF
--- a/crates/testing-utils/src/node_proc.rs
+++ b/crates/testing-utils/src/node_proc.rs
@@ -51,11 +51,19 @@ where
         force_authoring: bool,
         bootnode: Option<String>,
         threshold_url: Option<String>,
+        node_key: Option<String>,
     ) -> TestNodeProcessBuilder
     where
         S: AsRef<OsStr> + Clone,
     {
-        TestNodeProcessBuilder::new(program, chain_type, force_authoring, bootnode, threshold_url)
+        TestNodeProcessBuilder::new(
+            program,
+            chain_type,
+            force_authoring,
+            bootnode,
+            threshold_url,
+            node_key,
+        )
     }
 
     /// Attempt to kill the running substrate process.
@@ -84,6 +92,7 @@ pub struct TestNodeProcessBuilder {
     force_authoring: bool,
     bootnode: Option<String>,
     tss_server_endpoint: Option<String>,
+    node_key: Option<String>,
 }
 
 impl TestNodeProcessBuilder {
@@ -93,6 +102,7 @@ impl TestNodeProcessBuilder {
         force_authoring: bool,
         bootnode: Option<String>,
         tss_server_endpoint: Option<String>,
+        node_key: Option<String>,
     ) -> TestNodeProcessBuilder
     where
         P: AsRef<OsStr>,
@@ -105,6 +115,7 @@ impl TestNodeProcessBuilder {
             force_authoring,
             bootnode,
             tss_server_endpoint,
+            node_key,
         }
     }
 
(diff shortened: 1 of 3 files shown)


crates/testing-utils/src/substrate_context.rs
```diff
@@ -118,13 +126,21 @@ pub async fn test_node_process_testing_state(
...
     let alice_bootnode = Some(
-        "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWM7EoKJfwgzAR1nAVmYRuuFq2f3GpJPLrdfhQaRsKjn38"
+        "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWMrQiZJKkbkZrb7NfkF3u2cu1i5js3tuC3LTYYHfoVbyE"
             .to_string(),
     );
-    let result =
-        test_node(Keyring::Alice, format!("--chain={}", chain_spec_type), force_authoring, None)
-            .await;
+    let alice_node_key =
+        Some("0xd43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d".to_string());
```
ameba23, line 135: How did you generated this, and how did you get the libp2p node ID from this?  Just by starting `entropy --node-key <blah>` and then reading it out from the logs?

crates/testing-utils/src/substrate_context.rs
```diff
@@ -118,13 +126,21 @@ pub async fn test_node_process_testing_state(
     chain_spec_type: ChainSpecType,
     force_authoring: bool,
 ) -> Vec<TestNodeProcess<EntropyConfig>> {
+    // boot node is generated with node key for determinisim
     let alice_bootnode = Some(
-        "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWM7EoKJfwgzAR1nAVmYRuuFq2f3GpJPLrdfhQaRsKjn38"
+        "/ip4/127.0.0.1/tcp/30333/p2p/12D3KooWMrQiZJKkbkZrb7NfkF3u2cu1i5js3tuC3LTYYHfoVbyE"
```
ameba23, line 131: Im a bit confused by how things were working before - we were specifying a bootnode libp2p id, but not the secret key?  Was this being set automatically when we pass `--alice`?

crates/testing-utils/src/node_proc.rs
```diff
@@ -140,6 +151,13 @@ impl TestNodeProcessBuilder {
             cmd.arg(arg);
         }
 
+        if let Some(node_key) = &self.node_key {
+            let arg = format!("--node-key={}", node_key.as_str());
+            cmd.arg(arg);
+        } else {
+            cmd.arg("--unsafe-force-node-key-generation");
```
ameba23, line 158: From looking at the CI output, it looks like this argument is already being given and it barks it we give it more than once.

ameba23 (approved):
Great - generally seems like a good move, and if it helps with testing locally thats good.

Failures in CI need fixing.

crates/testing-utils/src/substrate_context.rs
JesseAbram: well the node key was deterministic i mean in a chaintype Development or Local it should be but for some reason it just stopped so now I set it determinstically here

crates/testing-utils/src/substrate_context.rs
JesseAbram: using subkey any public key it will convert it to the proper key type. For more info I was keeping updates on this issue as it is related as well as polkadot docs on this https://github.com/entropyxyz/devops-infrastructure/issues/121

PR #1473, opened 2025-06-10
Build entropy-tss CVM image in release workflow

ameba23:
When releasing for TDX (a release tag which does not contain the sub-string `non-TDX`) we want to build a CVM image for entropy-tss rather than a docker image.

Here is a test release made on this branch:
https://github.com/entropyxyz/entropy-core/releases/tag/test%2Frelease%2Ftest-tss-cvm-build-06

the file [`core-image-minimal-tdx-gcp.rootfs.wic.tar.gz`](https://github.com/entropyxyz/entropy-core/releases/download/test%2Frelease%2Ftest-tss-cvm-build-06/core-image-minimal-tdx-gcp.rootfs.wic.tar.gz) is the CVM image.

To check the non-production flow still works i also made a `non-TDX` test release on this branch:
- https://github.com/entropyxyz/entropy-core/releases/tag/untagged-9809ba2b32e7012e0eac
- https://github.com/entropyxyz/entropy-core/actions/runs/15635976610

TODO:
- [x] Fix getting branch name
- [x] Update release checklist to document how to use this
- [x] Do a `non-TDX` test release to check that it still works

### DIFF
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -89,7 +89,7 @@ jobs:
             {entropy,entropy-tss}_${{ needs.git-ref-basename.outputs.git_ref_basename }}/*
 
   release-TDX:
-    name: Publish new release
+    name: Publish new TDX release
     if: github.event_name != 'schedule'
     needs:
       - git-ref-basename
@@ -116,4 +116,3 @@ jobs:
             --verify-tag "${{ github.ref_name }}" \
             entropy_${{ needs.git-ref-basename.outputs.git_ref_basename }}/* \
             tss-cvm-image/core-image-minimal-tdx-gcp.rootfs.wic.tar.gz
-
--- a/.github/workflows/build-and-release.yaml
+++ b/.github/workflows/build-and-release.yaml
@@ -61,7 +61,7 @@ jobs:
       if: "!contains(github.ref_name, 'non-TDX')"
       uses: ./.github/workflows/entropy-tss-cvm.yaml
 
-  release:
+  release-non-TDX:
     name: Publish new release
     if: github.event_name != 'schedule'
     needs:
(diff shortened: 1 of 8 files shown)


.github/workflows/build-and-release.yaml
```diff
@@ -82,3 +87,32 @@ jobs:
             $(echo ${{ github.ref_name }} | grep --quiet -E '^test|v[0-9]\.[0-9]\.[0-9]-' && echo '--prerelease') \
             --verify-tag "${{ github.ref_name }}" \
             {entropy,entropy-tss}_${{ needs.git-ref-basename.outputs.git_ref_basename }}/*
+
+  release-TDX:
```
ameba23, line 91: @entropyxyz/system-reliability-engineers here i have duplicated the job above for the case that the CVM image is built.  This appears to work, but it duplicates code, so if you know a neater way to do it, go for it.

.github/workflows/entropy-tss-cvm.yaml
```diff
@@ -0,0 +1,36 @@
...
+          ref: 'peg/allow-setting-src-rev'
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          version: latest
+
+      - name: Build
+        run: |
+          make image-base CVM_SERVICE_SRC_REV=${{ github.sha }} CVM_SERVICE_SRC_BRANCH=peg/release-pipeline-build-cvm
```
ameba23, line 29: This still needs fixing.  We need the branch name, and we cannot use `github.ref_name` because that will give us the tag name.

Probably we need an extra step which uses the git CLI to get the branch name of the current commit (for the entropy-core repo).

I'm not sure how to do this.

.github/workflows/entropy-tss-cvm.yaml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  entropy-tss-cvm-build:
+    runs-on: core-build-runner
+    timeout-minutes: 90
+    steps:
+      - name: Checkout yocto build scripts
+        uses: actions/checkout@v4
+        with:
+          repository: 'entropyxyz/yocto-build'
+          ref: 'peg/allow-setting-src-rev'
```
ameba23, line 20: We can remove this line once https://github.com/entropyxyz/yocto-build/pull/17 is merged

cooldracula:
@ameba23 : I am excited to dig into this, but am realizing that I think we'd benefit from a whiteboard session of sort where we can map out the deployment concretely.  Now that we have working GCP deployments, it is easier to plan and diagram the upgrade to TDX.  From that session we could move backwards to the release workflows.  

In general, the more obvious and explicit the jobs, the better it is for SRE--which can mean that repeated code and single use jobs (i.e. to get a github ref) can be a good thing, even if it doesn't seem the most elegant.  

In the end, the YAML file should just be an articulate summary of our strategy-- which is why I'd love to discuss that in a face-to-face meeting.

I've set an event in our calendar for this, but happy to adjust the time and such so it works for all of us.

cooldracula (approved):
LGTM!

Issue #1438, opened 2025-05-21
Consider storing the full TDX quote as part of 'server info' in the staking extension / outtie pallets


ameba23:
When TSS nodes or API key service nodes join, they must submit a TDX quote.  This is verified by the attestation pallet, and if valid the PCK is extracted stored with the server info.

I think it could be worth storing the whole quote as part of the server info, for two reasons:

1. This would make it possible for external parties (anybody) to verify the quotes / certificate chains and trust that the services they are using are genuine. Since the chain runtime itself is not attested this is the only way to be sure that the quote really is valid. By 'anybody' i am thinking client software and / or other CVM nodes could do their own verification.
2. This would mean we wouldn't need two separate types, `ServerInfo` (containing PCK) and `JoiningServerInfo`:

https://github.com/entropyxyz/entropy-core/blob/731ae4bbd7e1140956d900065d4900ebb72c5c0f/pallets/staking/src/lib.rs#L123-L138

The reason **not** to do this would be if there were concerns about them being too big to store in a `StorageMap`.  I'm guessing this would increase the transaction costs?  Quotes are typically 4 or 5 kb. @JesseAbram

ameba23:
Been discussion this with @dvdplm and we think it would be nice to have for the client (desktop app) to be able to independently verify attestations of a chosen API key service instance before using it.

PR #1480, opened 2025-06-17
Store full TDX quotes on-chain to allow clients to verify them independently

ameba23:
Closes https://github.com/entropyxyz/entropy-core/issues/1438

This is quite a big change, and i guess will make `validate` more expensive, but i really think its worth it.

I propose to do the same thing for the 'forest' pallet in a follow-up PR.


crates/threshold-signature-server/src/attestation/api.rs
```diff
@@ -83,6 +78,9 @@ pub async fn create_quote(
     let mut pck_seeder = StdRng::from_seed(tss_account.0);
     let pck = tdx_quote::SigningKey::random(&mut pck_seeder);
 
+    // In the real thing this is the key used in the quoting enclave
+    let signing_key = tdx_quote::SigningKey::random(&mut pck_seeder);
```
ameba23, line 82: We make this deterministic by using the seeded RNG rather than `OsRng`.  This makes testing simpler - and since these are mock quotes we do not need genuinely random keypairs.

crates/client/src/user.rs
```diff
@@ -174,3 +184,27 @@ pub async fn request_attestation(
 
     Ok(nonce)
 }
+
+/// Check build-time measurement matches a current-supported release of entropy-tss
+/// This differs slightly from the attestation pallet implementation because here we don't have direct
+/// access to the parameters pallet - we need to make a query
+pub async fn check_quote_measurement(
```
ameba23, line 191: This function has not changed, but been moved from entropy-tss to entropy-client as this check can now be done by clients as well.

crates/shared/src/attestation.rs
```diff
@@ -37,12 +37,55 @@ impl QuoteInputData {
...
+        let mut hasher = Blake2s256::new();
+        hasher.update(tss_account_id.encode());
+        hasher.update(x25519_public_key);
+        hasher.update(context.encode());
+        let hashed_input: [u8; 32] = hasher.finalize().into();
+
+        hashed_input == self.0[..32]
+    }
+
+    /// Verify quote input data from TSS `ServerInfo` where exact context is not known
```
ameba23, line 71: This is needed because the context of an on-chain quote can be change - initially it will be `validate` but if a validator changes endpoint or threshold accounts the quote will be updated, so when verifying we check all possible contexts.

JesseAbram (approved):

PR #1488, opened 2025-06-26
Create forest helper functions

JesseAbram:


crates/client/src/forest.rs
```diff
@@ -0,0 +1,84 @@
...
+            .map_err(|_| SubstrateError::NoEvent)?
+            .ok_or(SubstrateError::NoEvent)?;
+        Ok(())
+    };
+    // TODO: maybe add loggings here if fialed
+    backoff::future::retry(backoff.clone(), add_tree).await.map_err(|_| ClientError::TimedOut)?;
+    Ok(())
+}
+
+fn create_test_backoff() -> ExponentialBackoff {
```
ameba23, line 61:
```suggestion
#[cfg(test)]
fn create_test_backoff() -> ExponentialBackoff {
```

ameba23 (approved):
Generally, great :star: - im very happy that this stuff is de-duplicated and available to other 'trees'.

But entropy-client is supposed to be a client library and is a dependency of the network status page and the test-cli.

Im happy with putting this stuff here if its the most logical place in terms of our dependency heirarchy, but i don't want to confuse client developers by having a bunch of server functions exposed by default.  

So can we put the `attestation` and `forest` modules behind a feature flag (eg: `server`) please.

Approving pending this change.

JesseAbram:
@ameba23 makes sense will do for this PR, if you want to create a new crate for this im down, was trying to not like add more and more crates but you know if it makes sense

PR #1490, opened 2025-06-30
In production mode allow chain endpoint to be set via HTTP request

ameba23:
**Note**: This targets `release/v0.4.0` not `master` 

For context see this discussion: https://github.com/entropyxyz/devops-infrastructure/issues/113

When entropy-tss is compiled with the `production` feature flag for TDX, this allows the associated chain-endpoint to be set via an HTTP request like so:

```
curl -X POST http://localhost:3001/v1/set_chain_endpoint \
        -H "Content-Type: text/plain" \
        --data 'ws://chain-endpoint:9944'
```

This request may be made exactly one time, and until this request is received, none of the other HTTP routes will work, including `/version` and `/info`, which means this needs to be set before those endpoint are used to get the details needed for the chainspec.

The way it is implemented is a bit hacky - i want to test if this works, and going forwards i am not sure if we will end up using it, as once we have [persistent storage](https://github.com/entropyxyz/meta-entropy-tss/pull/5) we could maybe store this configuration in the persistent part of the image which is not measured.

Here is a test release built on this branch: https://github.com/entropyxyz/entropy-core/releases/tag/untagged-e87d40f586f0fdf1872b

### DIFF
--- a/crates/threshold-signature-server/src/helpers/launch.rs
+++ b/crates/threshold-signature-server/src/helpers/launch.rs
@@ -17,7 +17,9 @@
 
 use std::{
     fs,
+    net::SocketAddr,
     path::PathBuf,
+    sync::{Arc, Mutex},
     time::{SystemTime, UNIX_EPOCH},
 };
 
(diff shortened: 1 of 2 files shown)


ameba23:
Closing this in favor of setting via cloud init metadata

PR #1499, opened 2025-07-03
Fix production build

JesseAbram:
Seems like there is no check for different feature flags in CI, maybe we should reconsider that for the future, anyways this is a fix


JesseAbram:
> Great ⭐ - i checked that this compiles my end - and totally agree that going forwards we should be checking this in CI

opened an issue #1502

PR #1506, opened 2025-07-09
Generalize logger

JesseAbram:
Moves the logger file in core to client under the server flag, to be used in tree development


node/cli/src/chain_spec/mod.rs
```diff
@@ -231,7 +231,7 @@ pub fn mock_measurement_values() -> Vec<(SupportedCvmServices, MeasurementValues
             vec![BoundedVec::try_from(MEASUREMENT_VALUE_MOCK_QUOTE.to_vec()).unwrap()],
         ),
         (
-            SupportedCvmServices::ApiKeyService,
+            SupportedCvmServices::TreeService,
```
ameba23, line 234: Ok cool so this implies that all supported trees are going to be added to this vec, rather than adding an extra variant to `SupportedCvmServices` for each tree which would be a PITA.

I cant remember where we got to in the discussion about this before i went away, but i think this is a good move.

crates/client/Cargo.toml
```diff
@@ -43,6 +43,15 @@ js-sys ={ version="0.3.74", optional=true }
 tokio  ={ version="1.44", features=["time"] }
 backoff={ version="0.4.0", features=["tokio"] }
 
+# Logging 
+hostname                ={ version="0.4", optional=true }
+tracing-subscriber      ={ version="0.3.19", features=["env-filter", "json"], optional=true }
+tracing-loki            ={ version="0.2", optional=true }
+tower-http              ={ version="0.6.6", features=["trace", "cors"], optional=true }
```
ameba23, line 50: I can't see where we use tower-http in this crate.  Maybe i am missing something - but i tried compiling without it and it seems to work.

ameba23 (approved):
:100: Great to not need to repeat this in each tree

PR #1512, opened 2025-07-28
Fix change_threshold benchmark

JesseAbram:
Threshold would pass in test but fail in running benchmark, this is because instead of control we have in test to set it to null, the quote would be randomly generate, this is mitigated by manually insterting the quote instead of assuming it is null


ameba23 (approved):
:star: nice one.  The benchmark now passes my end.

